PR #7, opened 2018-10-03
Fix library paths for multiarch

aragilar:
Hi, this is a patch from Debian (https://salsa.debian.org/science-team/sundials/blob/master/debian/patches/0002-Fix-library-paths-for-multiarch.patch) which fixes which path the libraries are installed to. Currently the libraries are always installed to <prefix>/lib, which means a 32-bit version and 64-bit version cannot be co-installed (see https://wiki.debian.org/Multiarch/HOWTO and references therein). With this patch, cmake will automatically install the libraries to the correct path (whether it's <prefix>/lib, <prefix>/lib/x86-64 or <prefix>/lib64). I've previously posted this to the mailing list, but there was no reply.


balos1:
Thanks for submitting this pull request, we appreciate it.

However, we are not able to incorporate code directly into Sundials at this time as we do not have a contribution agreement in place. We hope to have this set up soon but until we have an agreement drafted and approved by LLNL we will not be able to include any code directly.

We will address this bug in our next patch release.

PR #10, opened 2019-02-05
Do not expose KLU API to user applications.

carlodefalco:
* include/sunlinsol/sunlinsol_klu.h : only include definitions of exported SUNDIALS symbols.
* include/sunlinsol/sunlinsol_klu_internal.h : (new file) KLU specific definitions needed for implementation of the KLU interface.
* src/sunlinsol_klu/CMakeLists.txt : also install sunlinsol_klu_internal.h in case users need it
* src/sunlinsol_klu/sunlinsol_klu.c : include sunlinsol_klu_internal.h

Signed-off-by: Carlo de Falco <carlo.defalco@polimi.it>

### DIFF
--- a/include/sunlinsol/sunlinsol_klu.h
+++ b/include/sunlinsol/sunlinsol_klu.h
@@ -39,7 +39,6 @@
 #include <sundials/sundials_matrix.h>
 #include <sundials/sundials_nvector.h>
 #include <sunmatrix/sunmatrix_sparse.h>
-#include "klu.h"
 
 #ifdef __cplusplus  /* wrapper to enable C++ usage */
 extern "C" {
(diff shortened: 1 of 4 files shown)


carlodefalco:
The issue solved by this change was already brought up in the mailing list,
and was motivated by making it easier to include the SUNDIALS interface in
the Octave build environment and.

I am creating a pull request  as I think sometimes a few lines of code are easier
to understand than a more verbose explanation.

For reference see this thread on the sundials-users mailing list : http://sundials.2283335.n4.nabble.com/sundials-3-and-4-tp4654629.html

And this is the discussion related to SUNDIALS > 2.x in Octave : 
https://savannah.gnu.org/bugs/index.php?52475

carlodefalco:
Sorry I didn't mean to close the pull request, it was just a mistake, so I am re-opening

balos1:
In future versions of SUNDIALS, it will be necessary for us to expose the KLU API to the user applications. This is because we will be refactoring the SUNDIALS KLU interface to be more of an adapter between a `SUNLinearSolver` and KLU and less of an abstraction of KLU.

For example, a user might have to set the ordering type by setting the `ordering` member in the `klu_common` struct rather than calling the method `SUNKLUSetOrdering`. 

We believe the new approach will provide more flexibility to SUNDIALS users who want to utilize KLU (and other third party libraries). In addition, it will increase SUNDIALS's resilience to changes in third party libraries. 

Therefore, we don't plan on hiding the KLU API as requested. However, I think it is reasonable for us to provide a CMake option to specify how to include KLU @gardner48. We could do something like:

```c
#ifdef KLU_INCLUDE_STANDALONE
#include "klu.h"
#else
#include <suitesparse/klu.h>
#endif
```

`KLU_INCLUDE_STANDALONE` would be controlled via a corresponding CMake option.


Also I will note we are working towards providing a package config file although it is not imminent.

carlodefalco:
Thanks for your comments!

> Therefore, we don't plan on hiding the KLU API as requested. However, I think it is reasonable for us to provide a CMake option to specify how to include KLU @gardner48. We could do something like:
> 
> ```c
> #ifdef KLU_INCLUDE_STANDALONE
> #include "klu.h"
> #else
> #include <suitesparse/klu.h>
> #endif
> ```

Another option that would not even require any CMake modifications

```c
#ifndef _KLU_H
#include <klu.h>
#endif
```

so that user code would be responsible of including klu headers itself and can take care of the  header search path themselves. This would also completely solve the problem for Octave. 
I would say it is reasonable that code that wants to interact with the klu API directly have to include klu.h explicitly. 

> In future versions of SUNDIALS, it will be necessary for us to expose the KLU API to the user applications. This is because we will be refactoring the SUNDIALS KLU interface to be more of an adapter between a `SUNLinearSolver` and KLU and less of an abstraction of KLU.
> 
> For example, a user might have to set the ordering type by setting the `ordering` member in the `klu_common` struct rather than calling the method `SUNKLUSetOrdering`.

Indeed, that is the reason why, following your comment in the mailing list thread, the patch also exports the sunlinsol_klu_internal.h header in the final installation. Applications needing to deal with KLU internals would inlude  sunlinsol_klu_internal.h while those which are just fine with the SUNDIALS abstraction can just include sunlinsol_klu.h.

I don't intend to insist in any way with such a change but I just intended to show that it is minimal and not intrusive at all, while it has been suggested that to work around this in the Octave autotools build system would require a complete rewrite of the suitesparse detection logic.

This rewrite will probably never be done until we see how the "future versions of SUNDIALS" you refer to are out. That means that probably we in Octave and, as a consequence, all distributions packaging Octave will still be stuck with having to support SUNDIALS 2.x for a long time.

If splitting the header still looks too intrusive to you, I would insist with the other option suggested above that is even more minimal.

> `KLU_INCLUDE_STANDALONE` would be controlled via a corresponding CMake option.

Note that even some distributions that install KLU only as part of suitesparse still need 

```c
#include <klu.h> 
```

rather than 

```c
#include <suitesparse/klu.h> 
```

Thanks again for your time reviewing and commenting.

PR #12, opened 2019-02-06
Do not include klu.h if already included by user

carlodefalco:
* include/sunlinsol/sunlinsol_klu.h : check whether klu.h was already included
  by user before including it

Signed-off-by: Carlo de Falco <carlo.defalco@polimi.it>

### DIFF
--- a/include/sunlinsol/sunlinsol_klu.h
+++ b/include/sunlinsol/sunlinsol_klu.h
@@ -39,7 +39,9 @@
 #include <sundials/sundials_matrix.h>
 #include <sundials/sundials_nvector.h>
 #include <sunmatrix/sunmatrix_sparse.h>
-#include "klu.h"
+#ifndef _KLU_H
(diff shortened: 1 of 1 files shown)


carlodefalco:
In the discussion about a previous [pull request](https://github.com/LLNL/sundials/pull/10) I suggested this change here would probably be a simpler solution to the same problem.

Here I am submitting the suggested change in the form of a pull request.

balos1:
We are accepting the changes in this pull request, but since our Github repository is currently just snapshots of our releases I cannot merge it here. I will download the PR as a patch and apply it to our development repository. 

The patch will be included in the SUNDIALS release which will occur in the next couple of weeks.

Thank you for your contribution!

balos1:
Reopening this until the patch makes it into this repository.

balos1:
The 4.1.0 release includes this fix. Closing the PR now.

PR #13, opened 2019-02-15
NLS Solve: check for required callbacks

tbrk:
Avoid a NULL pointer exception on incorrect use of either SUNNonlinearSolver_Newton or SUNNonlinearSolver_FixedPoint from external code.

This allows proper detection of error conditions from managed languages (e.g., OCaml) and appropriate signalling (e.g., raising of exceptions) rather than crashing due to NULL-pointer dereference.


gardner48:
@tbrk You're correct the Newton and FixedPoint modules should include these checks. 

We're accepting the changes in the pull request, but since our GitHub repository is currently just snapshots of release we cannot merge it here. We've propagated this patch to the development repository and it will be included in the next release.

tbrk:
OK. Thank you!

balos1:
This fix has been included in the 5.0.0-dev.0 release. Thanks for the contribution!

PR #19, opened 2019-08-01
CMakeLists.txt: Use CMAKE's "Always full RPATH"

flokli:
When building tests, the test scripts already have RUNPATH set to
`${CMAKE_INSTALL_PREFIX}/lib…`, which might not be populated yet in
sandboxed builds, so the dynamic linker will fail to find them, and the
tests fail.

By using CMAKE's "Always full RPATH" logic described in
https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling#always-full-rpath, we can fix this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,26 @@ else()
   endif()
 endif()
 
+# use, i.e. don't skip the full RPATH for the build tree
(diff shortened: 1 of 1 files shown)


balos1:
Included in SUNDIALS 5.5.0.

PR #25, opened 2019-12-04
Drop dev.2 suffix for release version

xantares:
It makes the version string easier to parse


balos1:
Thanks for reporting this problem. The "dev.2" suffix will be removed in the next release and should have been removed with the 5.0.0 release. However, future "dev" releases will still maintain a suffix like this, so I would recommend parsing it still.

PR #27, opened 2019-12-23
If CMAKE_C_COMPILER is set, use it also to test compilation of KLU

carlodefalco:
If CMAKE_C_COMPILER is set, use it also to test compilation of KLU


gardner48:
Thanks for reporting this issue, the compiler for the KLU test has been fixed in the 5.1.0 release.

Issue #31, opened 2020-02-26
Segmentation fault in CVAfindIndex


aseyboldt:
I am using the adjoint sensitivity analysis functionality of sunodes, and sporadically I get segmentation faults during the backward pass. Unfortunately I could not reproduce this with a small example so far, but the coredump seems to indicate to me, that CVAfindIndex tries to access checkpoints that do not exist relatively near t0 of the forward problem, in a region where the solver is making (ridiculously?) small steps.
![Screenshot from 2020-02-26 11-18-52](https://user-images.githubusercontent.com/1882397/75335535-ee4cdd00-5889-11ea-8828-345fc105eeff.png)
It seems that CVAfindIndex is trying to find a checkpoint for `t = 161.33623519238427` but the largest of the 600 entries in ca_mem->dt_mem has only `t = 161.33623519238293`.

The details of how I'm using sundials are somewhat hidden in a python wrapper and pymc3 (I'm sampling the parameter space with an hamiltonian sampler), but here is a rough outline of what I'm doing:
- Initialize forward and backward solvers
- Repeat (a lot):
    - Call `CVodeReInit` and `CVodeAdjReInit`
    - Run forward solver
    - Call `CVodeReInitB`, `CVodeQuadReInitB` and `CVodeB`solve repeatedly, as the adjoint rhs is not continuous.

The `t` of the segfault is nowhere near the discontinuities of the rhs, the first one of those is at t ~ 12000.

The source for the solver calls is here: https://github.com/aseyboldt/sunode/blob/master/sunode/solver.py#L365

I can also provide the coredump if that is helpful.

aseyboldt:
I think I figured out what the problem here seems to be:
Let's assume there is only one backward problem.
At the beginning of the loop that advances all the backward problems ([here](https://github.com/LLNL/sundials/blob/master/src/cvodes/cvodea.c#L1330)), `ck_mem` is initialized so that `ck_mem->ck_t0 < cvB_mem->cv_mem->cv_tn < ck_mem->ck_t1`.

The solver sets `ck_mem->ck_t0` as stop time ([here](https://github.com/LLNL/sundials/blob/master/src/cvodes/cvodea.c#L1364)) and advances the backward problem. If the solver reached that stop time (so `cvB_mem->cv_tout == ck_mem->ck_t0`), then `cvB_mem->cv_mem->cv_tn` will still be larger than the stop time by a small amount, since it (incorrectly in this case) assumes it can not compute the rhs at the stop time itself.
[...]

PR #32, opened 2020-02-26
Check upper end of buffer in CVAfindIndex

aseyboldt:
Prevent reading out of bounds array in CVafindIndex. This should fix the segfault in #31, but possibly not the whole issue.

### DIFF
--- a/src/cvodes/cvodea.c
+++ b/src/cvodes/cvodea.c
@@ -2264,12 +2264,17 @@ static int CVAfindIndex(CVodeMem cv_mem, realtype t,
 
     *indx = ca_mem->ca_ilast;
     for(;;) {
+      if ( *indx >= ca_mem->ca_nsteps ) break;
       if ( sign*(t - dt_mem[*indx]->t) > ZERO) (*indx)++;
       else                                     break;
     }
 
     ca_mem->ca_ilast = *indx;
-
+    if ( *indx == ca_mem->ca_nsteps - 1 ) {
+      if ( SUNRabs(t - dt_mem[*indx]->t) > FUZZ_FACTOR * cv_mem->cv_uround ) {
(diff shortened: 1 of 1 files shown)


src/cvodes/cvodea.c
```diff
@@ -2264,12 +2264,17 @@ static int CVAfindIndex(CVodeMem cv_mem, realtype t,
...
     for(;;) {
+      if ( *indx == ca_mem->ca_nsteps - 1 ) break;
       if ( sign*(t - dt_mem[*indx]->t) > ZERO) (*indx)++;
       else                                     break;
     }
 
     ca_mem->ca_ilast = *indx;
-
+    if ( *indx == ca_mem->ca_nsteps - 1 ) {
+      if ( SUNRabs(t - dt_mem[*indx]->t) > FUZZ_FACTOR * cv_mem->cv_uround ) {
```
aseyboldt: I copied this from 2255. But shouldn't it distinguish between `t - dt_mem[*indx]->t > 0` and `t - dt_mem[*indx]->t) < 0`?

src/cvodes/cvodea.c
```diff
@@ -2264,12 +2264,17 @@ static int CVAfindIndex(CVodeMem cv_mem, realtype t,
...
     *indx = ca_mem->ca_ilast;
     for(;;) {
+      if ( *indx >= ca_mem->ca_nsteps ) break;
       if ( sign*(t - dt_mem[*indx]->t) > ZERO) (*indx)++;
       else                                     break;
     }
 
     ca_mem->ca_ilast = *indx;
-
+    if ( *indx == ca_mem->ca_nsteps - 1 ) {
```
aseyboldt: This isn't quite right yet. What exactly should happen here at the right end of the array?

PR #39, opened 2020-06-09
examples/cvode/serial/cvRoberts_klu.c: Make the jacobian really sparse

t-b:
The example is supposed to show how the sparse matrices are used. But it
actually allocated and fills in the full matrix with 9 entries. But two
of these are zero, so by omitting these two now we tell the user how filling
it sparsely works.


balos1:
Incorporated into SUNDIALS v6.1.0. Thanks!

PR #40, opened 2020-06-09
config/FindKLU.cmake: Also use debug library suffix

t-b:
It is common to add suffixes to the created libraries in
order to denote things like debug/release configurations.

Search that also for the KLU library.

Encountered with SuiteSparse cmake build from
https://github.com/jlblancoc/suitesparse-metis-for-windows.


balos1:
Included in SUNDIALS v5.5.0.

PR #41, opened 2020-06-09
cvLsDenseDQJac: [Non-functional change] Use correct dimension getter

t-b:
We clearly need to iterate over the columns as the function comment and
later code states. But the rows are queried here. Changing to columns
will not change anything as the matrix is square.


t-b:
@drreynolds Thanks for the review. Good catch, I've fixed it.

balos1:
Fixed in v5.4.0.

Issue #44, opened 2020-06-26
Infinite loop in CVodeF


aseyboldt:
Under certain conditions the forward solver of cvodes enters an infinite loop if called using `CV_NORMAL_STEP`. If the step-size decreases until `t + h = t` numerically, then [CVode](https://github.com/LLNL/sundials/blob/master/src/cvodes/cvodea.c#L504) will print a warning, but still return a 0 return value. The loop [here](https://github.com/LLNL/sundials/blob/master/src/cvodes/cvodea.c#L500) in `CVodeF` will continue forever.

I do not have a reproducing example in C, but here is one that uses the [sunode](https://github.com/aseyboldt/sunode) package to wrap sundials:

We start integrating the ode `dy/dt = 1 / (t - 1)` at `t0 = 1 + 1e-15`:

```python
import sunode
import sunode.symode
import numpy as np

params = {'p': ()}
states = {'a': ()}

def rhs(t, y, p):
    return {
        'a': 1 / (t - 1)
    }

problem = sunode.SympyProblem(
    params=params,
    states=states,
    derivative_params=[('p',)],
    rhs_sympy=rhs,
)

solver = sunode.solver.AdjointSolver(problem)
solver.set_params_dict({'p': 0.})

h = 1e-15
tvals = np.linspace(1 + h, 1 + 1000 * h)
out, *_ = solver.make_output_buffers(tvals)

y0 = np.array(0.)
solver.solve_forward(1 + h, tvals, y0, out)
```

The solver prints error messages repeatedly until the maximum number of warnings in exhausted, then silently continues:
```
[CVODES WARNING]  CVode
  Internal t = 1 and h = 4.96507e-17 are such that t + h = t on the next step. The solver will continue anyway.


[CVODES WARNING]  CVode
  Internal t = 1 and h = 4.96507e-17 are such that t + h = t on the next step. The solver will continue anyway.


[CVODES WARNING]  CVode
  Internal t = 1 and h = 4.96507e-17 are such that t + h = t on the next step. The solver will continue anyway.


[CVODES WARNING]  CVode
  Internal t = 1 and h = 4.96507e-17 are such that t + h = t on the next step. The solver will continue anyway.


[CVODES WARNING]  CVode
  The above warning has been issued mxhnil times and will not be issued again for this problem.
```

CC @astoeriko

balos1:
What version of sundials does sunode interface to?

aseyboldt:
[...]

PR #45, opened 2020-06-26
Check the maximum number of steps in CVodeF

aseyboldt:
Fix for #44.
I'm not sure exactly about what is happening here:
https://github.com/LLNL/sundials/blob/master/src/cvodes/cvodea.c#L576
I think this should still run in the error case?


balos1:
This fix will be incorporated in the next release of sundials.

balos1:
Fixed in v5.4.0.

PR #58, opened 2021-01-04
cmake: don't assume CMAKE_INSTALL_<dir> is relative

doronbehar:
https://github.com/jtojnar/cmake-snips#assuming-cmake_install_dir-is-relative-path


gardner48:
These changes were included in the SUNDIALS v5.7.0 release.

PR #61, opened 2021-03-10
examples/cvode/CXX_serial/cv_Heat.cpp: Fix compilation problem with Visual Studio 2019

ChrisHal:
`stod` and `stoi` were not defined because of missing `include <string>` (seems like gcc does not need this include)


balos1:
Included in v5.8.0.

PR #64, opened 2021-05-21
CMake: Fix exported targets so that build dir can be used directly

ZedThree:
Two small fixes that allow more complete parity between installed versions and using the build directory directly.

The call to `export(EXPORT)` is the main fix -- without this, `add_subdirectory`/`FetchContent` didn't work properly.

The other change is to add aliases without the `_shared/static` suffix -- without this change, using `add_subdirectory`/`FetchContent` would only have e.g. `SUNDIALS::arkode_shared`, while the installed version would have `SUNDIALS::arkode` too.


balos1:
Hi @ZedThree , can you sign this commit as per the [contributing guidelines](https://github.com/LLNL/sundials/blob/master/CONTRIBUTING.md)? Thanks!

ZedThree:
@balos1 Done!

balos1:
@ZedThree Thanks! I will try this out. So long as it passes our tests, it should be incorporated into the next release.

balos1:
Incorporated into SUNDIALS v6.1.0. Thanks!

PR #69, opened 2021-08-04
SUNMatCopy: also copy the matvecsetup op

tbrk:
Signed-off-by: Timothy Bourke <tim@tbrk.org>


balos1:
Thanks @tbrk. This is has already been fixed in our development repository, so it'll be addressed in the next release.

PR #79, opened 2021-09-15
arkRootInit does not initialize the irfnd field which is later used directly in arkRootCheck2

tbrk:


balos1:
Included in v5.8.0.

PR #80, opened 2021-09-22
Only print local red. and XBraid headings in root

tbrk:
This avoids interleaving of headers from multiple processes which makes
it difficult to compare test traces.


balos1:
Included in v5.8.0.

PR #85, opened 2021-10-27
Proposition for *GetUserData

tbrk:
Hello. I wonder whether you would please consider adding GetUserData optional output functions to all solvers (possibly as proposed in the attached commit)?

The rationale is as follows. Callbacks from nonlinear solvers (e.g., sys, lsetup, lsolve, convtest) are passed a `void *mem` argument which is a pointer to the integrator structure (e.g., cvode_mem). When overriding these functions, it would be useful to be able to also get at the associated `void *user_data` pointer. In particular, I would like to exploit such a possibility in the OCaml interface to Sundials.

My proposition returns the user data pointer directly. I guess an alternative would be to return it by reference and use the return value to signal error conditions. I could rework the patch in this way if you prefer.


tbrk:
It may be even more useful to have *GetUserData functions that return a pointer to the user_data field. For interfaces to Sundials from languages with garbage collectors (e.g., OCaml), the pointer could be registered as a "global root" to be followed by the garbage collector and the user_data field could thus hold a value from the higher-level language.

tbrk:
Are you guys open to having `GetUserData` functions? It would help me to avoid a nasty hack in the OCaml interface to Sundials. I can prepare another patch if you like.

drreynolds:
I see no reason against having these, though I'd like to hear from @gardner48 and @balos1 on this before asking @tbrk to create a PR.

gardner48:
I don't see a problem with adding these functions. There's an old branch (`feature/get-user-data`) where I'd implemented them that just needs the documentation. I'll update the docs and open a PR.

tbrk:
Thanks @drreynolds and @gardner48, I would really appreciate it! That will avoid this risky mess: https://github.com/inria-parkas/sundialsml/blob/107f97c9ae5effa8a9e830e35b6a9cc0aecc0960/src/cvode/cvode_ml.c#L903

PR #89, opened 2021-11-04
ark_kpr_mri: mind q's and p's in butcher tables

tbrk:
It seems to me that the second update of `q` should really be an update of `p`.

### DIFF
--- a/examples/arkode/C_serial/ark_kpr_mri.c
+++ b/examples/arkode/C_serial/ark_kpr_mri.c
@@ -264,7 +264,7 @@ int main(int argc, char *argv[])
     B->c[1] = RCONST(0.5);
     B->c[2] = ONE;
     B->q=3;
-    B->q=2;
+    B->p=2;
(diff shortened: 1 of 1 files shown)


drreynolds:
This should be q, not p, and is correct as originally written.  The convention throughout ARKODE is to use “q” for the method order, and “p” for the embedding order.  Within MRIStep there currently do not exist embedded MRI tables, hence all tables must have p=0, but have q that corresponds with the asymptotic order of the MRI method.

tbrk:
I understand. Thanks for responding.

In this case it might be worth deleting the first assignment? I just couldn't understand the point of two successive assignments to the same member:
```
B->q=3;
B->q=2;
```

drreynolds:
Indeed. We’ll be sure to remove the extraneous lines soon.

drreynolds:
These have been fixed in our development repository, and will be released soon.

tbrk:
Thank you!

PR #132, opened 2022-04-26
Added a new ManyVector demonstration code graphic for hybrid CPU-GPU calculation

drreynolds:
Our `sunorg.pptx` file contained our previous ManyVector graphic for the SUNDIALS demonstration code, but that graphic assumed each fluid field was MPI-parallel, and made no distinction between CPU and GPU memory.  I've added a new graphic to that file that better highlights the current configuration of our ManyVector usage in the demonstration code (all existing graphics were left unchanged).


cswoodward (dismissed):

gardner48:
There is also a `sunorg.pptx` in `doc/shared/figs` where all the shared figures are now for the RST docs. Both versions look the same other than the changes in this PR. I think this new version should replace the one in `doc/shared/fig/sunorg.ppt` and `doc/sundials/figures/sunorg.pptx` should be removed.

drreynolds:
There's also a pile of other duplicate/old files in `doc/sundials/figures/`; should we go ahead and remove those as well?

gardner48:
Yes, anything in that's duplicated and not needed to build the example tex docs can be removed.

cswoodward:
Hold off on deleting the old stuff.  In the past we had kept some of these around for historical reasons.  It will be nice to see the layout as it changed over the years.  We should keep the old files and maybe name an old file by the year it was relevant as we go forward.

balos1 (approved):

gardner48 (approved):

PR #146, opened 2022-05-27
Fix example mesh dx and add new option

gardner48:
Fix a bug in computing the mesh spacing and add an option to run fully implicit methods with the DIRK table from the default IMEX-ARK methods.


drreynolds (dismissed):
LGTM

drreynolds (approved):
Still looks good to me.

balos1 (approved):

PR #161, opened 2022-06-21
Add citation DOIs

gardner48:
* Citation formatting fixes
* Add DOIs for most citations


cswoodward:
It looks like the UG year and versions were updated but not the author lists.  Was this intended?

gardner48:
> It looks like the UG year and versions were updated but not the author lists. Was this intended?

Good catch, the author lists for the user and examples guides in the bib file are not consistent with the current versions. I've updated them accordingly and checked that they are correct the in the `CITATIONS.md` and `doc/*/guide/source/Landing.rst` files.

cswoodward (commented):
Looks good.

PR #170, opened 2022-07-01
Bugfix/fortran nvgetarraypointer

### LINKED ISSUE #160: Fortran interface to NVector returns a pointer with the wrong dimension
When using the Fortran 2003 interface to Sundial, retrieved arrays from NVectors are not the correct dimension

https://github.com/LLNL/sundials/blob/e2f29c34f324829302037a1492db480be8828084/src/sundials/fmod/fsundials_nvector_mod.f90#L824

The Fortran-C interface assumes NVector arrays are always length 1. So, for example when I do

```fortran
type(N_Vector)        :: sunvec_y
real(c_double), pointer :: yvec(:)

yvec => FN_VGetArrayPointer(sunvec_y)
```

then `yvec` is a fortran pointer to an array of length 1, instead of the actual length of the NVector. Instead, `yvec` should be given the proper length during the call to `FN_VGetArrayPointer`

balos1:
Opening as a draft to get implementation feedback.

Note some of the changes are from regenerating the Fortran. The focus of this PR is just the `N_VGetArrayPointer` stuff. Would fix #160.


balos1:
We could alternatively do a typedef like
```c
typedef realtype N_Vector_DataType
```
and then the function declarations would be like
```c
N_Vector_DataType* N_VGetArrayPointer(N_Vector v);
```

balos1:
One thought: this could also _maybe_ useful if we ever wanted to do some sort of mixed-precision. `N_Vector_Data` could be defined differently than `realtype`.

gardner48:
* For parallel vectors `N_VGetLength` will be incorrect since it’s the global length. This could be addressed by adding a `N_VGetLocalLength` function. 
* It’s not clear to me how this would help with mixed precision, could you elaborate? 
* Would we also need `SUNMatrix_Data`?

PR #180, opened 2022-07-25
update examples for petsc-next

balos1:


gardner48 (commented):
This looks good. My only question is if we really need the `#if PETSC_VERSION` checks or if we can just use the cases without error checking.

balos1:
> This looks good. My only question is if we really need the `#if PETSC_VERSION` checks or if we can just use the cases without error checking.

The PETSc developer guide seems to indicate the functions perform error checking internally, but it does not say if they always did or not. Satish provided this patch, so I would be inclined to use it as-is.

gardner48 (approved):

PR #188, opened 2022-08-16
Update SUNMemoryHelper_Destroy

gardner48:
Check for unlikely cases when destroying a memory helper

### DIFF
--- a/src/sundials/sundials_memory.c
+++ b/src/sundials/sundials_memory.c
@@ -207,7 +207,7 @@ int SUNMemoryHelper_Destroy(SUNMemoryHelper helper)
 
   if (helper->ops)
   {
-    if (helper->ops->destroy(helper))
+    if (helper->ops->destroy)
     {
       /* user helper defined destroy */
       return helper->ops->destroy(helper);
--- a/src/sundials/sundials_memory.c
+++ b/src/sundials/sundials_memory.c
@@ -205,39 +205,22 @@ int SUNMemoryHelper_Destroy(SUNMemoryHelper helper)
 {
   if (!helper) return 0;
 
-  if (helper->ops)
+  if (helper->ops->destroy)
   {
-    if (helper->ops->destroy)
-    {
-      /* user helper defined destroy */
-      return helper->ops->destroy(helper);
-    }
-    else if (helper->content)
-    {
-      /* helper should have defined destroy */
-      return -1;
-    }
-    else
-    {
-      /* default destroy */
-      free(helper->ops);
-      free(helper);
-      return 0;
-    }
+    /* user helper defined destroy */
+    return helper->ops->destroy(helper);
+  }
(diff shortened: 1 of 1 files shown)


src/sundials/sundials_memory.c
```diff
@@ -203,23 +203,44 @@ int SUNMemoryHelper_CopyAsync(SUNMemoryHelper helper, SUNMemory dst,
 
 int SUNMemoryHelper_Destroy(SUNMemoryHelper helper)
 {
-  if (helper->ops->destroy == NULL)
+  if (!helper) return 0;
+
+  if (helper->ops)
   {
-    if (helper->content != NULL)
+    if (helper->ops->destroy(helper))
```
drreynolds, line 210: Shouldn't this be `if (helper->ops->destroy)` so that you can check the nullity of the pointer before calling the function?

drreynolds (changes requested):
Once we've resolved my previous comment, this looks good to merge.

drreynolds (dismissed):

src/sundials/sundials_memory.c
```diff
@@ -203,23 +203,44 @@ int SUNMemoryHelper_CopyAsync(SUNMemoryHelper helper, SUNMemory dst,
 
 int SUNMemoryHelper_Destroy(SUNMemoryHelper helper)
 {
-  if (helper->ops->destroy == NULL)
+  if (!helper) return 0;
+
+  if (helper->ops)
```
balos1, line 208: Can `helper->ops` really ever be `NULL` and it not be an error (i.e., outside of this function)? If the answer is no, then I think you can change line 206 to
```c
if (!helper || !helper->ops) return 0;
```
and simplify the logic below.

gardner48: This shouldn't happen unless there is an error when constructing the helper and this function is then called to destroy up the partially constructed object. So I'm inclined to leave the extra check if only to cover this corner case.

balos1: We should chat on Tuesday about the error handling updates. Specifically, I think we should remove checks like these and almost uniformly trust that the "class" objects are not corrupt/null. The short version of my reasoning is that it is not an error that will be recoverable from in any case I can think of in SUNDIALS, so it is better to let a segfault happen (which is very easy to debug, while, for example, it is much harder to debug why a function returned NULL). I can elaborate at our meeting.

balos1: I was thinking... for `Destroy` functions we should definitely check if the object is `NULL` or not so we can emulate `delete` rather than `free` (i.e., dont error out if the object is `NULL`). However, I think the check to make sure the ops is not `NULL` is not needed.

balos1 (changes requested):

balos1 (approved):

PR #216, opened 2022-10-20
Feature/findpetsc with pkgconf

balos1:


cmake/tpl/FindPETSC.cmake
```diff
@@ -13,764 +11,109 @@
...
-        "  PETSC_INCLUDES=${PETSC_INCLUDES}\n")
-      message (FATAL_ERROR ${msg})
+# Substitute MPI target if PETSC is built with MPI
+foreach(_next_lib IN LISTS PKG_PETSC_STATIC_LIBRARIES)
+  if(_next_lib MATCHES "mpi")
+    if(DEFINED CMAKE_CXX_COMPILER)
+      list(APPEND _petsc_libs "MPI::MPI_CXX")
+    else()
+      list(APPEND _petsc_libs "MPI::MPI_C")
     endif()
```
gardner48, line 59: Why does which MPI target we use depend on if we have a C++ compiler?

cmake/tpl/FindPETSC.cmake
```diff
@@ -13,764 +11,109 @@
...
+list(APPEND _petsc_libs "${SUNDIALS_MATH_LIBRARY}")
 
-    # check if the include directory was found
-    if (NOT PETSC_INCLUDE_DIR)
-      string (CONCAT msg
-        "Could not determine PETSc include directory from PETSC_INCLUDES:\n"
-        "  PETSC_INCLUDES=${PETSC_INCLUDES}\n")
-      message (FATAL_ERROR ${msg})
+# Substitute MPI target if PETSC is built with MPI
+foreach(_next_lib IN LISTS PKG_PETSC_STATIC_LIBRARIES)
```
gardner48, line 53: Are these libraries always in `PKG_PETSC_STATIC_LIBRARIES`? Could they also be in `PKG_PETSC_LIBRARIES`?

cmake/tpl/FindPETSC.cmake
```diff
@@ -13,764 +11,109 @@
...
-          INTERFACE_LINK_LIBRARIES "${PETSC_LIBRARIES_${suffix}}")
-        set_target_properties (SUNDIALS::PETSC_${suffix} PROPERTIES
-          INTERFACE_COMPILE_OPTIONS ${PETSC_DEFINITIONS})
-        set_target_properties (SUNDIALS::PETSC_${suffix} PROPERTIES
-          IMPORTED_LOCATION ${PETSC_LIBRARY_${suffix}})
-      endif ()
-    endforeach ()
-  endif ()
-endif (PETSC_FOUND)
+mark_as_advanced(PETSC_INCLUDE_DIRS PETSC_LIBRARIES PETSC_VERSION_MAJOR PETSC_VERSION_MINOR PETSC_VERSION_PATCH VERSION_VAR PETSC_VERSION)
```
gardner48, line 119: Is `VERSION_VAR` needed? I didn't see it above.

cmake/tpl/SundialsKokkos.cmake
```diff
@@ -42,7 +42,7 @@ endif()
 # -----------------------------------------------------------------------------
 find_package(Kokkos REQUIRED
   HINTS "${Kokkos_DIR}"
-  NO_DEFALT_PATH)
+  NO_DEFAULT_PATH)
```
gardner48, line 45: Nice catch!

cmake/tpl/SundialsKokkosKernels.cmake
```diff
@@ -42,7 +42,7 @@ endif()
 # -----------------------------------------------------------------------------
 find_package(KokkosKernels REQUIRED
   HINTS "${KokkosKernels_DIR}"
-  NO_DEFALT_PATH)
+  NO_DEFAULT_PATH)
```
gardner48, line 45: Nice catch!

cmake/tpl/SundialsPETSC.cmake
```diff
@@ -49,13 +49,14 @@ endif()
...
-message(STATUS "PETSC_LIBRARIES:   ${PETSC_LIBRARIES_}")
-message(STATUS "PETSC_INCLUDES:    ${PETSC_INCLUDES_}")
-message(STATUS "PETSC_INDEX_SIZE:  ${PETSC_INDEX_SIZE}")
-message(STATUS "PETSC_PRECISION:   ${PETSC_PRECISION}\n")
+message(STATUS "PETSC_DIR:          ${PETSC_DIR}")
+message(STATUS "PETSC_ARCH:         ${PETSC_ARCH}")
+message(STATUS "PETSC_LIBRARIES:    ${PETSC_LIBRARIES}")
+message(STATUS "PETSC_INCLUDE_DIRS: ${PETSC_INCLUDE_DIRS}")
+message(STATUS "PETSC_INDEX_SIZE:   ${PETSC_INDEX_SIZE}")
+message(STATUS "PETSC_PRECISION:    ${PETSC_PRECISION}")
```
gardner48, line 59: Are `PETSC_INDEX_SIZE` and `PETSC_PRECISION` still defined? I think these got removed in the switch to pkgconfig. Can we still get these to check compatibility?

examples/arkode/CXX_serial/CMakeLists.txt
```diff
@@ -68,7 +68,7 @@ foreach(example_tuple ${ARKODE_examples})
 
     set_target_properties(${example_target} PROPERTIES FOLDER "Examples")
 
-    target_compile_features(${example_target} PRIVATE cxx_std_11)
+    target_compile_features(${example_target} PRIVATE cxx_std_14)
```
gardner48, line 71: I think these could actually be removed, I believe `CMAKE_CXX_STANDARD` overrides this

cmake/tpl/FindPETSC.cmake
balos1: This is just how it was done in xsdk-examples. Probably doesnt need to be here. The thought is that if you are using the C++ compiler you should prefer the CXX target, but in SUNDIALS having the C++ compiler doesnt necessarily mean we are using it.

cmake/tpl/FindPETSC.cmake
balos1: No they wont be in `PKG_PETSC_LIBRARIES`, that only will contain PETSc libraries, no dependencies.

balos1: They will be in `PKG_PETSC_STATIC_LIBRARIES` if PETSc was built with MPI and Kokkos respectively.

cmake/tpl/FindPETSC.cmake
balos1: Nope, not needed anymore.

cmake/tpl/SundialsPETSC.cmake
balos1: They are not defined, and actually havent been getting populated in awhile from what I can tell. PETSc doesnt put the information in the files we currently/used to look in. We will have to do something else instead. I left them here for now and they will just continue to be unpopulated.

examples/arkode/CXX_serial/CMakeLists.txt
balos1: They could be I think, but not sure what other effects it would have so didnt do it in this PR (I only updated them as it was causing an error).

gardner48: Sounds good, we can leave them for now and see about removing them after the release.

PR #219, opened 2022-10-22
Feature/kokkos vector benchmark

balos1:


examples/cvode/petsc/CMakeLists.txt
```diff
@@ -91,6 +91,10 @@ if(EXAMPLES_INSTALL)
     set(LIBS "-lsundials_cvode_fused_stubs ${LIBS}")
   endif()
 
+  if(SUNDIALS_BUILD_PACKAGE_FUSED_KERNELS)
+    set(SUNDIALS_EXTRA_LIBS SUNDIALS::cvode_fused_stubs)
+  endif()
+
```
gardner48, line 97: This is covered by `set(LIBS "-lsundials_cvode_fused_stubs ${LIBS}")` above
```suggestion
```

doc/shared/sunmatrix/SUNMatrix_KokkosDense.rst
```diff
@@ -44,7 +44,7 @@ class in the ``sundials::kokkos`` namespace:
 .. code-block:: cpp
 
    template<class ExecutionSpace = Kokkos::DefaultExecutionSpace,
-            class MemorySpace = class ExecSpace::memory_space>
+            class MemorySpace = typename ExecSpace::memory_space>
```
gardner48, line 47: This same edit is needed in `NVector_Kokkos.rst`

gardner48 (changes requested):

drreynolds (dismissed):

benchmarks/nvector/kokkos/test_nvector_performance_kokkos.cpp
```diff
@@ -228,15 +228,15 @@ void sync_device(N_Vector x)
  * Functions required for clearing cache
  * --------------------------------------------------------------------*/
 
-int InitializeClearCache(SizeType cachesize)
+int InitializeClearCache(sunindextype cachesize)
```
balos1, line 231: @gardner48  I don think this should be a `sunindextype` since it has no relation to indexing or sundials core. `size_t` would be more appropriate I think.

benchmarks/nvector/kokkos/test_nvector_performance_kokkos.cpp
```diff
@@ -0,0 +1,267 @@
...
+  sundials::Context sunctx;
+  int flag;            /* return flag     */
+
+  /* CLI args */
+  sunindextype veclen;     /* vector length   */
+  int print_timing;        /* output timings  */
+  int ntests;              /* number of tests */
+  int nvecs;               /* number of tests */
+  int nsums;               /* number of sums  */
+  sunindextype cachesize;  /* cache size      */
```
balos1, line 68: I think cachesize should be size_T rather than sunindextype since it shouldn't depend on what size indexing sundials is built with (it doesn't index a sundials data structure).

benchmarks/nvector/kokkos/test_nvector_performance_kokkos.cpp
```diff
@@ -0,0 +1,267 @@
...
+/* ----------------------------------------------------------------------
+ * Main NVector Testing Routine
+ * --------------------------------------------------------------------*/
+int main(int argc, char* argv[])
+{
+  sundials::Context sunctx;
+  int flag;            /* return flag     */
+
+  /* CLI args */
+  sunindextype veclen;     /* vector length   */
```
balos1, line 63: @gardner48 Should veclen match SizeType since that's what the constructor asks for? I'm not sure what the advantage is to use sunindextype for it and just cast it to SizeType later.

gardner48: This could be `SizeType` but we'd need to swap some occurrences of `veclen` below to `static_cast<sunindextype>(veclen)` to avoid the conversion warning e.g., `flag = Test_N_VLinearSum(X, static_cast<sunindextype>(veclen), ntests);`

benchmarks/nvector/kokkos/test_nvector_performance_kokkos.cpp
gardner48: Agreed `size_t` would be best here.

benchmarks/nvector/kokkos/test_nvector_performance_kokkos.cpp
gardner48: Agreed, `size_t` would be better.

gardner48 (approved):

PR #223, opened 2022-10-27
Bugfix/GitHub issue 74 lapack test fail

balos1:


doc/arkode/guide/source/Introduction.rst
```diff
@@ -126,6 +126,12 @@ Fixed a bug with the Kokkos interfaces that would arise when using clang.
 Fixed a compilation error with the Intel oneAPI 2022.2 Fortran compiler in the
 Fortran 2003 interface test for the serial ``N_Vector``.
 
+Fixed a bug in the SUNLINSOL_LAPACKBAND and SUNLINSOL_LAPACKDENSE modules
+which would cause the tests to fail on some platforms. 
+
+Fixed a bug in the SUNLINSOL_LAPACKBAND and SUNLINSOL_LAPACKDENSE modules
+which would cause the tests to fail on some platforms. 
+
```
gardner48, line 134:
```suggestion
```

gardner48 (changes requested):
Minor doc fix, otherwise LGTM

gardner48 (approved):

examples/sunlinsol/lapackband/test_sunlinsol_lapackband.c
```diff
@@ -135,13 +135,12 @@ int main(int argc, char *argv[])
   /* Print result */
   if (fails) {
     printf("FAIL: SUNLinSol module failed %i tests \n \n", fails);
-    printf("\nA (original) =\n");
-    SUNBandMatrix_Print(B,stdout);
```
drreynolds, line 139: Was printing of the banded matrices causing the test failures, or were these removed for some other reason?

examples/sunlinsol/lapackdense/test_sunlinsol_lapackdense.c
```diff
@@ -132,13 +132,12 @@ int main(int argc, char *argv[])
   /* Print result */
   if (fails) {
     printf("FAIL: SUNLinSol module failed %i tests \n \n", fails);
-    printf("\nA (original) =\n");
-    SUNDenseMatrix_Print(B,stdout);
```
drreynolds, line 136: Was printing of the dense matrices causing the test failures, or were these removed for some other reason?

drreynolds (commented):
Unless printing of the original/factored matrices when encountering a failure causes trouble, I would like for these to be added back in.

examples/sunlinsol/lapackdense/test_sunlinsol_lapackdense.c
balos1: It was not contributing to the failures, but it was making deciphering the log difficult.

examples/sunlinsol/lapackband/test_sunlinsol_lapackband.c
balos1: It was not contributing to the failures, but it was making deciphering the log difficult.

balos1:
@drreynolds I am not sure what the advantage to having them printed is when this is a test of the linear solver not the matrices.

drreynolds:
We would print both the original matrix and the factored matrix.  If the direct linear solver is somehow defective, then that factorization will have errors, and by looking at those we may be able to deduce the problem.  

The probability of an error is rather small, so I do not see any problem with having this extra output when a failure occurs.  Plus, if an error does arise, we would need to manually add these print statements back in to see what was wrong with the factorization.

cswoodward:
It seems that Cody's suggestion of making this an option would be reasonable - it leaves the error output cleaner for gathering info on the initial failure, but then flipping it on gives info on the matrix factorization - which can be useful, as Dan notes.

PR #226, opened 2022-11-08
add profile option to testRunner

balos1:


drreynolds (approved):

cmake/macros/SundialsOption.cmake
```diff
@@ -57,9 +57,10 @@ macro(sundials_option NAME TYPE DOCSTR DEFAULT_VALUE)
 
   if(all_depends_on_dependencies_met)
 
-    # if necessary, create the CACHE variable with its default value
     if(NOT DEFINED ${NAME})
       set(${NAME} "${DEFAULT_VALUE}" CACHE ${TYPE} ${DOCSTR})
+    else()
+      set(${NAME} "${NAME}" CACHE ${TYPE} ${DOCSTR})
```
gardner48, line 63: Is this needed? If `${NAME}` is already defined as a cache variable, this should have no effect.

balos1: Yes, it is needed to avoid uninitialized cache variables (this might only effect newer cmake versions, but it doesn't hurt with older ones as far as I can tell). See https://gitlab.kitware.com/cmake/cmake/-/issues/21615.

balos1: E.g., if you run

```cmake
cmake  -DSUNDIALS_TEST_UNITTESTS=ON -DSUNDIALS_TEST_DEVTESTS=ON \
      -DSUNDIALS_TEST_PROFILE=ON -DSUNDIALS_TEST_NODIFF=ON \
      -B builddir/ -S .
```

then in CMakeCache.txt, you will see

```
//Location of testing answer files
SUNDIALS_TEST_ANSWER_DIR:PATH=

//No help, variable specified on the command line.
SUNDIALS_TEST_DEVTESTS:UNINITIALIZED=ON

//Precision for floating point comparisons (number of digits)
SUNDIALS_TEST_FLOAT_PRECISION:STRING=-1

//Precision for integer comparisons (percent difference)
SUNDIALS_TEST_INTEGER_PRECISION:STRING=-1

//No help, variable specified on the command line.
SUNDIALS_TEST_NODIFF:UNINITIALIZED=ON

//Location to write testing output files
SUNDIALS_TEST_OUTPUT_DIR:PATH=

//No help, variable specified on the command line.
SUNDIALS_TEST_PROFILE:UNINITIALIZED=ON

//No help, variable specified on the command line.
SUNDIALS_TEST_UNITTESTS:UNINITIALIZED=ON
```

This is with cmake@3.24.0

gardner48: Ok, in that case should this be `set(${NAME} "${${NAME}}" CACHE ${TYPE} ${DOCSTR})`?

balos1: That makes more sense, although, it seems to work as-is strangely enough.

PR #235, opened 2022-12-15
Feature/fix fortran array bounds

balos1:
This is a permanent fix for #160. Unfortunately, there does not seem to be a better way to get SWIG to produce the end-result we want, which is to have a function like `FN_VGetArrayPointer` return a Fortran array with the correct dimensions so that bounds checking works. 

Alternative solutions discussed did not pan out in the end for various reasons.


src/sunmatrix/band/fmod/fsunmatrix_band_mod.f90
```diff
@@ -557,4 +553,34 @@ function FSUNMatSpace_Band(a, lenrw, leniw) &
...
+type(SUNMatrix), target, intent(inout) :: a
+integer(C_INT64_T), intent(in) :: j
+type(C_PTR) :: fresult 
+type(C_PTR) :: farg1 
+integer(C_INT64_T) :: farg2 
+
+farg1 = c_loc(a)
+farg2 = j
+fresult = swigc_FSUNBandMatrix_Column(farg1, farg2)
+call c_f_pointer(fresult, swig_result, [FSUNBandMatrix_UpperBandwidth(A)+FSUNBandMatrix_LowerBandwidth(a)+1])
```
gardner48, line 582: I'm confused on the indexing here. The result of `SUNBandMatrix_Column` returns a pointer to the diagonal entry and is indexed (in C) from -mu to ml but this will be indexed from 1 to mu+ml+1.

gardner48, line 582:
```suggestion
call c_f_pointer(fresult, swig_result, [FSUNBandMatrix_UpperBandwidth(a)+FSUNBandMatrix_LowerBandwidth(a)+1])
```

swig/Makefile
```diff
@@ -103,7 +103,9 @@ sunnonlinsol: $(SUNNONLINSOL:%:sunnonlinsol/fsunnonlinsol_%_mod.i)
 	done
 
 patch: .FORCE
-	cd ..; git apply swig/fnvector-getarraypointer.patch
+	cd ..; 
+	
+#git apply swig/fnvector-getarraypointer.patch
```
gardner48, line 108: Remove the patch target?

gardner48 (changes requested):
Made it up to the swig files, I'll take a look at those next.

src/sunmatrix/band/fmod/fsunmatrix_band_mod.f90
balos1: Yeah, I am not sure what would be quite right for Fortran. In C you can index from `-mu` to `ml` and this corresponds to a range of `mu+ml+1` hence why the shape of the array returned is equal to that. However,  as far as I can figure you cannot actually index like that in Fortran without violating the bounds checks, and really since the `SUNBandMatrix_Column` function returns a pointer to the diagonal it only seems to be possible to legally access that diagonal entry... so this should probably return an array of shape [1] or we should remove the function from the Fortran interface.

examples/sunmatrix/band/test_fsunmatrix_band_mod.f90
```diff
@@ -118,21 +119,19 @@ integer(C_INT) function unit_tests() result(fails)
...
     Adata => FSUNBandMatrix_Data(A)
     do jj = 0, N-1
-      offset = smu-mu+1 + jj*(smu+ml+1)
-      colj(-mu:ml) => Adata(offset:mu+ml)
+      offset = smu-mu + jj*(smu+ml+1)
       jstart = merge(-jj, -mu, jj < mu)
       jend = merge(N-1-jj, ml, jj > N-1-ml)
       do ii = jstart, jend
-        colj(ii) = jj - ii
+        Adata(offset+mu+ml+ii) = jj - ii
```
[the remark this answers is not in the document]
balos1: The updating indexing does not seem to produce the same matrix as the C test and results in a floating point exception in the matvec.

examples/sunlinsol/band/test_fsunlinsol_band_mod.f90
```diff
@@ -59,13 +58,12 @@ integer(C_INT) function unit_tests() result(fails)
     ! fill A matrix with uniform random data in [0, 1/N)
     Adata => FSUNBandMatrix_Data(A)
     do j=1, N
```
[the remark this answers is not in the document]
balos1: The updating indexing does not seem to produce the same matrix as the C test (the old indexing does) and results in a floating point exception in the matvec.

examples/sunmatrix/band/test_fsunmatrix_band_mod.f90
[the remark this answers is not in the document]
gardner48: My mistake this should haven been `Adata(offset+ii) = (jj - 1) * ii`. I'm not sure how the old indexing was working. `offset+mu+ml` is not the diagonal entry for column `jj` when `offset = smu-mu + jj*(smu+ml+1)`.

examples/sunlinsol/band/test_fsunlinsol_band_mod.f90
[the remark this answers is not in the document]
gardner48: See my [comment below](https://github.com/LLNL/sundials/pull/235#discussion_r1050180715)

examples/sunmatrix/band/test_fsunmatrix_band_mod.f90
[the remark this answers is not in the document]
balos1: I think all we need to do is keep the indexing Ive written, but change `offset` back to `offset = smu-mu+1 + jj*(smu+ml+1)` and then access it as `Adata(offset+mu+ii)`.  The current case with `offset+mu+ml+ii` only works because `mu` and `ml` are both 1 in the test.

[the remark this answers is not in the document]
balos1: > My mistake this should haven been `Adata(offset+ii) = (jj - 1) - ii`. I'm not sure how the old indexing was working. `offset+mu+ml` is not the diagonal entry for column `jj` when `offset = smu-mu + jj*(smu+ml+1)`.

The old indexing was working because `mu` and `ml` were 1.  I think now with my update the indexing you're doing is equivalent to my fixed version.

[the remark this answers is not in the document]
gardner48: Sounds good, but I'd suggest using `offset = smu + 1 + jj * (smu + ml + 1)` and `Adata(offset + ii)` so the offset is the diagonal entry.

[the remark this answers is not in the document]
balos1: Ill go with your fixed indexing scheme since it makes it `offset+ii` instead of `offset+mu+ii`.

swig/sunmatrix/fsunmatrix_sparse_mod.i
```diff
@@ -26,6 +26,120 @@
...
+integer(C_INT64_T), dimension(:), pointer :: swig_result
+type(SUNMatrix), target, intent(inout) :: a
+type(C_PTR) :: fresult 
+type(C_PTR) :: farg1 
+
+farg1 = c_loc(a)
+fresult = swigc_FSUNSparseMatrix_IndexPointers(farg1)
+call c_f_pointer(fresult, swig_result, [FSUNSparseMatrix_NP(a)+1])
+end function  
+%}
```
gardner48, line 145: Add newline at end of file

test/env/default.sh
```diff
@@ -147,6 +147,7 @@ if [ "$compilername" == "gcc" ]; then
     if [[ "$SUNDIALS_PRECISION" == "double" && "$SUNDIALS_INDEX_SIZE" == "32" ]]; then
         export CFLAGS="${CFLAGS} -Wconversion -Wno-sign-conversion"
         export CXXFLAGS="${CXXFLAGS} -Wconversion -Wno-sign-conversion"
+        export FFLAGS="${FFLAGS} -fbounds-check"
     fi
```
gardner48, line 151: Currently is not getting added when the Fortran interfaces are enabled
```suggestion
    fi
    export FFLAGS="${FFLAGS} -fbounds-check"
```

gardner48 (changes requested):

drreynolds (dismissed):
I went through the entire PR as well, and didn't find anything new to correct.  Looking at my inbox, apparently I missed all the fun a few hours ago.

examples/sunmatrix/band/test_fsunmatrix_band_mod.f90
```diff
@@ -118,39 +119,35 @@ integer(C_INT) function unit_tests() result(fails)
...
+      offset = (jj-1)*(smu+ml+1) + smu + 1 ! offset to diagonal
+      istart = merge(-mu, -(jj-1), jj > mu) ! above diagonal
+      iend = merge(N-jj , ml, jj > N - ml)  ! below diagonal
+      do ii = istart, iend
+        Adata(offset+ii) = (jj-1) - ii
       end do
     end do
 
     xdata(0:N-1) => FN_VGetArrayPointer(x)
     ydata(0:N-1) => FN_VGetArrayPointer(y)
```
gardner48, line 139: Looks like these lines are causing the CI to fail
```suggestion
    xdata => FN_VGetArrayPointer(x)
    ydata => FN_VGetArrayPointer(y)
```

examples/sunmatrix/band/test_fsunmatrix_band_mod.f90
```diff
@@ -118,39 +119,35 @@ integer(C_INT) function unit_tests() result(fails)
...
 
     ! Fill vectors
-    do ii = 0, N-1
-
+    do jj = 0, N-1
```
gardner48, line 142: With the change above the indexing below will need to be updated. Might be easiest to change `xdata(jj)` and `ydata(jj)` to `xdata(jj-1)` and `ydata(jj-1)` rather than change the loop to `1,N`

gardner48 (changes requested):

examples/sunmatrix/band/test_fsunmatrix_band_mod.f90
balos1: Interestingly, they are causing the compiler to crash! 

```
     xdata(0:N-1) => FN_VGetArrayPointer(x)
 1
internal compiler error: Segmentation fault
```

gardner48: That's probably it, the CI machine is using gcc@5.5 and passes without the bound checking flag but with bounds checking enabled if segfaults.

src/sunmatrix/band/fmod/fsunmatrix_band_mod.f90
balos1: I left the function in the interface, but return the shape as `[1]` and then added a warning to the docs.

balos1:
I believe all comments are addressed now @gardner48 .

gardner48 (dismissed):

drreynolds (approved):

test/env/default.sh
```diff
@@ -148,6 +148,9 @@ if [ "$compilername" == "gcc" ]; then
         export CFLAGS="${CFLAGS} -Wconversion -Wno-sign-conversion"
         export CXXFLAGS="${CXXFLAGS} -Wconversion -Wno-sign-conversion"
     fi
+    # TODO(CJB): add this back after we upgrade the GNU compiler stack on the Jenkins box
+    # Currently this causes the compiler to segfault on many of the Fortran example codes.
+    # export FFLAGS="${FFLAGS} -fbounds-check"
```
gardner48, line 153:
```suggestion
    # TODO(CJB): add this unconditionally after we upgrade the GNU compiler stack on the Jenkins box
    # Currently this causes the compiler to segfault on many of the Fortran example codes.
    if [ "$compilerversion" -gt "5.5.0" ]; then
        export FFLAGS="${FFLAGS} -fbounds-check"
    fi
```

gardner48 (approved):

PR #237, opened 2022-12-16
Add support for the SYCL backend with RAJA-v2022.x.y

gardner48:


balos1:
I am going to put this in SUNDIALS Next since it is not a bugfix and we are past the code freeze date. If we end up having time to get it into 6.5.0, then I suppose that is a bonus 😃 .

drreynolds (approved):

balos1 (approved):

drreynolds (approved):

Issue #253, opened 2023-02-20
`SUNMatScaleAdd_Sparse` has time complexity O(M*N)


phannebohm:
As the title says, adding two sparse matrices scales badly. I believe the problem are loops like the sparsity pattern overlap check here:
https://github.com/LLNL/sundials/blob/7c318244182061ea482b449bdb5a14d4d3c71e57/src/sunmatrix/sparse/sunmatrix_sparse.c#L827-L832

Just filling the work array `w` with zeros takes M*N operations. This happens in a few places inside this function with `w` and `x`. If I replace the for loops with calls to `memset` I save a significant amount of computation time (maybe 75% on my machine, don't trust that number though). However it still scales badly. I would expect addition to scale like O(NNZ) instead.

On a side note, for my particular use case (OpenModelica/OpenModelica#10178) I can take a modified version of `SUNMatScaleAddI_Sparse`, I call it `SUNMatScaleIAdd_Sparse` where the result is $A + c I$. This scales linearly as expected. Perhaps this function could be useful for other people.

balos1:
Thanks for noting this. Indeed this is because the current implementation support a very general case. Perhaps we can add an option (controlled with a Set function) to use less general, but faster implementations of `ScaleAdd` and `ScaleAddI`.

phannebohm:
I think I have a general solution that avoids these scaling issues, at least for `ScaleAddI`. I can open a PR for it and see if it can also be ported to `ScaleAdd`.

balos1:
Closed by #257.

PR #257, opened 2023-03-02
Fix time complexity issue for sparse matrix

### LINKED ISSUE #256: Sparse matrix elements not scaled completely
While working on a fix for #253, I found this part of `ScaleAddI` which is supposed to scale matrix A by c and add the identity:

https://github.com/LLNL/sundials/blob/7c318244182061ea482b449bdb5a14d4d3c71e57/src/sunmatrix/sparse/sunmatrix_sparse.c#L630-L637

It seems to me that only the left (top) part of A is scaled by c (if A is not square). Is this the desired behavior?

phannebohm:
Fixes part of #253 by avoiding nested loops that take `O(M*N)` operations.

Also fixes #256.


phannebohm:
I found the installation procedure for sundials and regression tests at https://sundials.readthedocs.io/en/latest/Install_link.html#testing-the-build-and-installation.

Running it locally revealed some failing tests. I will investigate.

src/sunmatrix/sparse/sunmatrix_sparse.c
```diff
@@ -656,30 +661,44 @@ int SUNMatScaleAddI_Sparse(realtype c, SUNMatrix A)
...
+        Ax[nz] = x[w[i]];
+      }
+
+      /* clear out temporary values for this column (row) */
+      for (i=0; i<nw; i++) {
+        x[w[i]] = RCONST(0.0);
+      }
+      if (j < M) {
+        x[j] = RCONST(0.0);
       }
```
drreynolds, line 702: Is this block even necessary?  It looks like the above code only ever uses values from `x` that it has filled in, so pre-existing values are just ignored.

src/sunmatrix/sparse/sunmatrix_sparse.c
```diff
@@ -724,30 +745,39 @@ int SUNMatScaleAddI_Sparse(realtype c, SUNMatrix A)
...
-        }
+      for (p=Ap[j]; p<Ap[j+1] && Ai[p]<j; p++) {
+        Ci[nz] = Ai[p];
+        Cx[nz++] = x[Ai[p]];
+      }
+      if (Ai[p] != j || Ap[j] == Ap[j+1]) {
+        Ci[nz] = j;
+        Cx[nz++] = x[j];
+      }
+      for (; p<Ap[j+1]; p++) {
```
drreynolds, line 770: Again, I recommend small comments be added to clarify structure, e.g.:
```suggestion
      /*  fill entries before diagonal */
      for (p=Ap[j]; p<Ap[j+1] && Ai[p]<j; p++) {
        Ci[nz] = Ai[p];
        Cx[nz++] = x[Ai[p]];
      }
      /*  fill diagonal if applicable */
      if (Ai[p] != j || Ap[j] == Ap[j+1]) {
        Ci[nz] = j;
        Cx[nz++] = x[j];
      }
      /*  fill entries after diagonal */
      for (; p<Ap[j+1]; p++) {
```

src/sunmatrix/sparse/sunmatrix_sparse.c
```diff
@@ -724,30 +745,39 @@ int SUNMatScaleAddI_Sparse(realtype c, SUNMatrix A)
...
+        Cx[nz++] = x[Ai[p]];
+      }
+
+      /* clear out temporary values for this column (row) */
+      for (p=Ap[j]; p<Ap[j+1]; p++) {
+        x[Ai[p]] = RCONST(0.0);
+      }
+      if (j < M) {
+        x[j] = RCONST(0.0);
       }
```
drreynolds, line 781: Again, is this code block necessary?

drreynolds (changes requested):
I apologize that it has taken me so long to get to this.  

I really like the basic change to operate on only the nonzeros in a given column/row, instead of creating and using work arrays that traverse all entries (even zero-valued ones).  I do have some requests, recommendations, and questions.

phannebohm:
Thanks for the feedback and suggestions. I'm a little busy rn, but I'll pick this up soon.

src/sunmatrix/sparse/sunmatrix_sparse.c
```diff
@@ -598,6 +598,9 @@ int SUNMatScaleAddI_Sparse(realtype c, SUNMatrix A)
   if (SM_DATA_S(A))       Ax = SM_DATA_S(A);
   else  return (SUNMAT_MEM_FAIL);
 
+  /* create work array for storing row (column) indices */
```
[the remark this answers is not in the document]
phannebohm: That's a good point. I'll revert my "enhancement" for case 1 so that `w` is only needed (and allocated) in case 2.

src/sunmatrix/sparse/sunmatrix_sparse.c
phannebohm: That's right

drreynolds (dismissed):

src/sunmatrix/sparse/sunmatrix_sparse.c
```diff
@@ -656,30 +655,38 @@ int SUNMatScaleAddI_Sparse(realtype c, SUNMatrix A)
...
-          Ai[--nz] = i;
-          Ax[nz] = x[i];
-        }
+      /* fill entries past diagonal */
+      for (i=nw-1; i>=0 && w[i]>j; i--) {
+        Ai[--nz] = w[i];
+        Ax[nz] = x[w[i]];
+      }
+      /* fill diagonal if applicable */
+      if (w[i] != j) {
```
balos1, line 682: When there are no nonzeros in the column (row), then `nw` would be `0` yielding `i = -1` which will result in out-of-bounds indexing of `w` here.  This might be related to why the tests are failing.

balos1 (changes requested):

src/sunmatrix/sparse/sunmatrix_sparse.c
phannebohm: Indeed. Thank you!

drreynolds:
It looks like only one test fails on the MacOS auto-tests (`examples/sunmatrix/sparse/test_sunmatrix_sparse 450 450 1 0`).  Interestingly, this test *passes* in Linux, but also fails on my mac.

This is the only test that uses a square matrix in CSR format (the non-square tests, and the square CSC tests, all pass).  The specific sub-test that fails is `Test_SUNMatScaleAddI2`, where the original matrix is missing nonzero values on the diagonal, and doesn't already include sufficient storage to add those in.  The specific failure occurs when checking the following (I'll use MATLAB notation below, for the input matrix `A`, vector `x`, and vector `y=A*x`):
```
B = I-A;
z = B*x;
w = x-y;
w == z;
```

Re-running this test manually with *smaller* matrices, I can get the same error on an 8x8 case:
```
A = [ 0 0 0 0 0 0 0 0
         0 0 0 0 0 0.8278173002543939 0 0
         0.8459815601101059 0 0 0 0 0 0 0
         0 0 0 0 0 0.5297473936945887 0 0
         0 0 0 0.3510152480336908 0 0 0 0
         0 0 0 0.7362245143094214 0 0 0 0
         0 0 0 0 0 0.6884553011918698 0 0.05008398371286876
         0 0 0 0 0 0 0 0]
B = [ 1 0 0 0 0 0 0 0
         0 1 0 0 0 -0.8278173002543939 0 0
         -0.8459815601101059 0 1 0 0 0 0 0
         0 0 0 1 0 -0.5297473936945887 0 0
         0 0 0 -0.3510152480336908 1 0 0 0
         0 0 0 -0.7362245143094214 0 0 0 0
         0 0 0 0 0 -0.6884553011918698 1 -0.05008398371286876
         0 0 0 0 0 0 0 1]
```
It looks like the only error is that in the 6th row, the "1" was never placed on the diagonal.

phannebohm:
Thanks @drreynolds for reviewing the failing test! I was trying to run [klee](https://github.com/klee/klee) on this to see which cases I'm missing, but I guess I'm not smart enough. I'll keep trying though.

In the meantime maybe your 8x8 example can reveal the bug. Unfortunately I don't have access to a mac, so I don't know if I can reproduce the missing "1". Qualitatively row 6 should be no different than row 3, right? That's weird.

phannebohm:
I think the problem was the following structure in rows 2 and 3:
```
A = [ * * *
      x 0 0
      0 x *]
```
(0 being zero, x being non-zero, * being anything)

If the diagonal entry was missing and followed by zeros and also the next row started with non-zeros just one column left of the diagonal entry, i.e. just below the missing diagonal entry above it, then my code missed that and thought the below entry belonged to the above row. (Hope that is comprehensible)

This should be fixed with the latest commit.

balos1 (approved):

balos1:
I did a bit of performance testing and verified this does indeed now scale as O(NNZ) as expected.

drreynolds (approved):
I agree with @balos1, this looks great.  My only concern is why that previous commit passed in Linux but failed in MacOS, though that is a matter for a separate conversation.

PR #259, opened 2023-03-10
Feature/clang 15 corona pipeline

yu-nix:
Adds CI support for Corona for testing Sundials with HIP (and TPLs).

Two jobs
- Just HIP +MPI
- HIP + Magma + Raja

Both jobs are for Clang 15.0.0 (rocmcc 5.4.1) and target gfx906.

Notes: 
- MPI is disabled for the time being (I will add it back once I change MPIEXEC_EXECUTABLE to Flux).
- ^ this has been addressed with a new variant in Spack and CMake option to specify run commands for MPI jobs.


.gitlab/build_and_test.sh
```diff
@@ -46,8 +47,20 @@ hostname=${hostname%%[0-9]*}
 BUILD_JOBS=${BUILD_JOBS:-"1"}
 
 # load newer python to try the clingo concretizer
-echo "module load python/3.8.2"
-module load python/3.8.2
+# Corona does not have python 3.8.2
+if [[ -n "${AMDGPU_TARGET}" ]]; then
```
balos1, line 51: Can we check to see if the machine is corona by checking the hostname instead?

.gitlab/build_and_test.sh
```diff
@@ -46,8 +47,20 @@ hostname=${hostname%%[0-9]*}
...
+    echo "module load python/3.9.12"
+    module load python/3.9.12
+else
+    echo "module load python/3.8.2"
+    module load python/3.8.2
+fi
+
+# Rocm version specific to Corona.
+if [[ -n "${AMDGPU_TARGET}" ]]; then
```
balos1, line 60: Can we check to see if the machine is corona by checking the hostname instead?

.gitlab/corona-jobs.yml
```diff
@@ -0,0 +1,37 @@
...
+# ------------------------------------------------------------------------------
+
+# Builds with Hip
+corona_clang_hip:
+  parallel:
+    matrix:
+      - COMPILER_SPEC: clang@15.0.0
+        AMDGPU_TARGET: [gfx906]
+  variables:
+    SPEC: "%${COMPILER_SPEC} cstd=99 cxxstd=14 precision=double amdgpu_target=${AMDGPU_TARGET} +rocm~mpi"
```
balos1, line 24: Why is mpi disabled?

.gitlab/corona-jobs.yml
```diff
@@ -0,0 +1,37 @@
...
+# ------------------------------------------------------------------------------
+# HIP + TPLs
+# ------------------------------------------------------------------------------
+corona_clang_hip_tpls:
+  parallel:
+    matrix:
+      - COMPILER_SPEC: clang@15.0.0
+        AMDGPU_TARGET: [gfx906]
+  variables:
+    SPEC: "%${COMPILER_SPEC} cstd=99 cxxstd=14 precision=double ~int64 amdgpu_target=${AMDGPU_TARGET} +rocm~mpi+magma+raja ^magma+rocm amdgpu_target=${AMDGPU_TARGET} ^raja+rocm~openmp~examples~exercises amdgpu_target=${AMDGPU_TARGET}"
```
balos1, line 36: Why is mpi disabled?

.gitlab/lassen-jobs.yml
```diff
@@ -69,7 +69,7 @@ lassen_xl_cuda_tpls:
       - COMPILER_SPEC: xl@16.1.1.10
         CUDA_SPEC: [cuda@10.1.243]
   variables:
-    SPEC: "%${COMPILER_SPEC} cstd=99 cxxstd=14 precision=double ~int64 +mpi+openmp+cuda~raja+magma+superlu-dist cuda_arch=70 ^superlu-dist+cuda cuda_arch=70 ^magma+cuda cuda_arch=70 ^${CUDA_SPEC}"
+    SPEC: "%${COMPILER_SPEC} cstd=99 cxxstd=14 precision=double ~int64 +mpi+openmp+cuda~raja+magma~superlu-dist cuda_arch=70 ^superlu-dist+cuda cuda_arch=70 ^magma+cuda cuda_arch=70 ^${CUDA_SPEC}"
```
balos1, line 72: Note to self: This was a change I made in the maintenance branch. I need the superlu team to figure out what is going on. It would be good to have this enabled before the PR is merged.

balos1 (changes requested):

.gitlab/corona-jobs.yml
yu-nix: It uses srun to run the tests, which Corona no longer supports. I'm currently trying to swap it over to flux.

gardner48: I thought Corona had aliases setup so slurm commands could still be used but if that's not working, setting adding `-DMPIEXEC_EXECUTABLE='flux mini run'` to the configuration should hopefully be all that's needed

yu-nix: Unfortunately, `DMPIEXEC_EXECUTABLE` seems to be for specifying where the scheduler is, so if I do `flux mini run` it gets confused when it tries to look for it.

cmake/SundialsBuildOptionsPre.cmake
```diff
@@ -292,3 +292,7 @@ sundials_option(SUNDIALS_TEST_DEVTESTS BOOL
 # Include unit tests in regression tests
 sundials_option(SUNDIALS_TEST_UNITTESTS BOOL
   "Include unit tests in make test" OFF ADVANCED)
+
+sundials_option(ENABLE_CUSTOM_MPI_RUN_COMMAND BOOL "Enable passing in a custom command prefix to run when running MPI jobs" OFF ADVANCED)
```
balos1, line 296: I think we can remove this option (see comment on the prefix variable).

balos1, line 296: I think we can remove this option (see comment on the prefix variable).

cmake/SundialsBuildOptionsPre.cmake
```diff
@@ -292,3 +292,7 @@ sundials_option(SUNDIALS_TEST_DEVTESTS BOOL
...
+
+sundials_option(CUSTOM_MPI_RUN_COMMAND_PREFIX STRING "Executable prefix to append before each MPI job command." "flux mini run" ADVANCED)
```
balos1, line 298:
```suggestion
sundials_option(SUNDIALS_TEST_MPIRUN_COMMAND STRING "Job scheduler or mpirun command used to launch SUNDIALS MPI tests" "${MPIEXEC_EXECUTABLE}" ADVANCED)
```

I think we can always use this variable in place of `MPIEXEC_EXECUTABLE` and just have it default to `MPIEXEC_EXECUTABLE`.  Then we do not need the `ENABLE_CUSTOM_MPI_RUN_COMMAND` option. 

On Corona, if you were to run cmake manually then you would do:
```
cmake -DSUNDIALS_TEST_MPIRUN_COMMAND="flux mini alloc"
```

To make this work in the CI, what you will need to do is add a spack variant that allows to set this value to the sundials spack package in `.gitlab/spack_packages/sundials/package.py`. We can discuss how to do this further offline if need be.

balos1 (changes requested):

cmake/SundialsBuildOptionsPre.cmake
yu-nix: Changed to only SUNDIALS_TEST_MPIRUN_COMMAND.

balos1 (dismissed):

balos1 (approved):

.gitlab-ci.yml
```diff
@@ -50,6 +50,8 @@
 # ON_QUARTZ:
 # Should the Quartz pipeline run? Set to  "ON" or "OFF" to enable/disable.
 #
+# SPACK_PREFIX: prefix used for shared spack installation, usually this a spack version number that matches the version set in the uberenv_config.json file. Spack installs go in /usr/workspace/sundials/spack_installs/${SPACK_PREFIX}/$(hostname).
+#
```
gardner48, line 54: Wrap long comment

.gitlab/build_and_test.sh
```diff
@@ -46,8 +47,20 @@ hostname=${hostname%%[0-9]*}
...
+else
+    echo "module load python/3.8.2"
+    module load python/3.8.2
+fi
+
+# Rocm version specific to Corona.
+if [[ "${hostname}" == "corona" ]]; then
+    echo "module load rocm/5.4.1"
+    module load rocm/5.4.1
+fi
```
gardner48, line 63: Combine with the block above.

gardner48: As a follow-on item, could use a spack environment with uberenv to avoid having machine specific modules loads?

.gitlab/corona-templates.yml
```diff
@@ -0,0 +1,35 @@
...
+    paths:
+      - spack-*.txt
+      - build_*/*
+    when: always
+  rules:
+    # Don't run if...
+    - if: '$CI_COMMIT_BRANCH =~ /_lnone/ || $ON_CORONA == "OFF" || $BENCHMARK == "ON"'
+      when: never
+    # Default is to run if previous stage succeeded
+    - when: on_success
```
gardner48, line 35: Add newline at end of file

.gitlab/corona-templates.yml
```diff
@@ -0,0 +1,35 @@
...
+# All rights reserved.
+#
+# See the top-level LICENSE and NOTICE files for details.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+# SUNDIALS Copyright End
+# ------------------------------------------------------------------------------
+
+# ------------------------------------------------------------------------------
+# Tags and rules to run tests on Lassen
```
gardner48, line 14:
```suggestion
# Tags and rules to run tests on Corona
```

.gitlab/corona-templates.yml
```diff
@@ -0,0 +1,35 @@
...
+  stage: c_build_and_test
+  needs: []
+  artifacts:
+    paths:
+      - spack-*.txt
+      - build_*/*
+    when: always
+  rules:
+    # Don't run if...
+    - if: '$CI_COMMIT_BRANCH =~ /_lnone/ || $ON_CORONA == "OFF" || $BENCHMARK == "ON"'
```
gardner48, line 32:
```suggestion
    - if: '$CI_COMMIT_BRANCH =~ /_cnone/ || $ON_CORONA == "OFF" || $BENCHMARK == "ON"'
```

.gitlab/lassen-jobs.yml
```diff
@@ -69,7 +69,7 @@ lassen_xl_cuda_tpls:
       - COMPILER_SPEC: xl@16.1.1.10
         CUDA_SPEC: [cuda@10.1.243]
   variables:
-    SPEC: "%${COMPILER_SPEC} cstd=99 cxxstd=14 precision=double ~int64 +mpi+openmp+cuda~raja+magma+superlu-dist cuda_arch=70 ^superlu-dist+cuda cuda_arch=70 ^magma+cuda cuda_arch=70 ^${CUDA_SPEC}"
+    SPEC: "%${COMPILER_SPEC} cstd=99 cxxstd=14 precision=double ~int64 +mpi+openmp+cuda~raja+magma+superlu-dist cuda_arch=70 superlu-dist cuda_arch=70 ^magma+cuda cuda_arch=70 ^${CUDA_SPEC}"
```
gardner48, line 72:
```suggestion
    SPEC: "%${COMPILER_SPEC} cstd=99 cxxstd=14 precision=double ~int64 +mpi+openmp+cuda~raja+magma+superlu-dist cuda_arch=70 ^superlu-dist+cuda cuda_arch=70 ^magma+cuda cuda_arch=70 ^${CUDA_SPEC}"
```

.gitlab/lassen-jobs.yml
```diff
@@ -78,7 +78,7 @@ lassen_gcc_cuda_tpls:
       - COMPILER_SPEC: gcc@8.3.1
         CUDA_SPEC: [cuda@11.5.0]
   variables:
-    SPEC: "%${COMPILER_SPEC} cstd=99 cxxstd=14 precision=double ~int64 +mpi+openmp+cuda+raja+magma+superlu-dist cuda_arch=70 ^superlu-dist+cuda cuda_arch=70 ^magma+cuda cuda_arch=70 ^raja+cuda~openmp~examples~exercises cuda_arch=70 ^${CUDA_SPEC}"
+    SPEC: "%${COMPILER_SPEC} cstd=99 cxxstd=14 precision=double ~int64 +mpi+openmp+cuda+raja+magma+superlu-dist+petsc+hypre+ginkgo cuda_arch=70 ^ginkgo+cuda cuda_arch=70 ^hypre ^petsc+cuda cuda_arch=70 ^superlu-dist+cuda cuda_arch=70 ^magma+cuda cuda_arch=70 ^raja+cuda~openmp~examples~exercises cuda_arch=70 ^${CUDA_SPEC}"
```
gardner48, line 81: Maybe this should be `^petsc` rather than `^petsc+cuda cuda_arch=70` (I'm not sure cuda capabilities in PETSc are actually getting used)

.gitlab/quartz-jobs.yml
```diff
@@ -55,11 +55,11 @@
 quartz_clang_tpls:
   parallel:
     matrix:
-      - COMPILER_SPEC: clang@12.0.0
+      - COMPILER_SPEC: clang@12.0.1
         INDEX_SPEC: [~int64]
         PRECISION_SPEC: [double]
   variables:
-    SPEC: "%${COMPILER_SPEC} cstd=99 cxxstd=14 ${INDEX_SPEC} precision=${PRECISION_SPEC} +mpi +openmp +hypre +superlu-dist +lapack +klu"
+    SPEC: "%${COMPILER_SPEC} cstd=99 cxxstd=14 ${INDEX_SPEC} precision=${PRECISION_SPEC} +mpi +openmp +hypre +superlu-dist +lapack +klu +petsc ^suite-sparse@5.13.0 ^openblas"
```
gardner48, line 62: Why is a specific version of `suite-sparse` needed? Is `^openblas` necessary?

.gitlab/quartz-jobs.yml
```diff
@@ -69,7 +69,7 @@ quartz_gcc_tpls:
         INDEX_SPEC: [~int64]
         PRECISION_SPEC: [double]
   variables:
-    SPEC: "%${COMPILER_SPEC} cstd=99 cxxstd=14 ${INDEX_SPEC} precision=${PRECISION_SPEC} +mpi +openmp +hypre +superlu-dist +lapack +klu"
+    SPEC: "%${COMPILER_SPEC} cstd=99 cxxstd=14 ${INDEX_SPEC} precision=${PRECISION_SPEC} +mpi +openmp +hypre +superlu-dist +lapack +klu +petsc ^suite-sparse@5.13.0"
```
gardner48, line 72: Why is a specific version of `suite-sparse` needed?

.gitlab/quartz-jobs.yml
```diff
@@ -79,5 +79,5 @@ quartz_intel_tpls:
         INDEX_SPEC: [~int64]
         PRECISION_SPEC: [double]
   variables:
-    SPEC: "%${COMPILER_SPEC} cstd=99 cxxstd=14 ${INDEX_SPEC} precision=${PRECISION_SPEC} +mpi +openmp +hypre ~superlu-dist +lapack +klu"
+    SPEC: "%${COMPILER_SPEC} cstd=99 cxxstd=14 ${INDEX_SPEC} precision=${PRECISION_SPEC} +mpi +openmp +hypre ~superlu-dist +lapack +klu ^suite-sparse@5.13.0"
```
gardner48, line 82: Why is a specific version of `suite-sparse` needed?

.gitlab/radiuss-spack-configs
gardner48, line 1: Is this to use the current `main` or `corona-ci` branch in our fork? If `corona-ci`, should we merge that to `main`?

.gitlab/spack_packages/netlib-lapack/package.py
gardner48, line 1: Do we need a `netlib-lapack` package (I think the tests are using OpenBLAS)? If not, can we remove the `netlib-lapack` directory?

.gitlab/spack_packages/sundials/package.py
```diff
@@ -27,7 +27,8 @@ class Sundials(CachedCMakePackage, CudaPackage, ROCmPackage):
     # Versions
     # ==========================================================================
     version("develop", branch="develop")
-    version("6.4.0", branch="develop")
+    version("6.4.1", sha256="7bf10a8d2920591af3fba2db92548e91ad60eb7241ab23350a9b1bc51e05e8d0")
```
gardner48, line 30: Add `6.5.0` and `6.5.1`

.gitlab/spack_packages/sundials/package.py
```diff
@@ -169,6 +178,9 @@ class Sundials(CachedCMakePackage, CudaPackage, ROCmPackage):
         "profiling", default=False, when="@6.0.0:", description="Build with profiling capabilities"
     )
 
+    # Scheduler
+    variant("scheduler", default="slurm", description="Specify which scheduler the system runs on.", values=("flux", "lsf", "slurm"))
```
gardner48, line 182: Looks like only the `flux` case is handled below. Should we be using `jsrun` and `srun` for LSF and SLURM, respectively? Should this have a "mpiexec" or "none" option to use whatever CMake finds for the MPI executable?

.gitlab/spack_packages/sundials/package.py
```diff
@@ -666,10 +698,9 @@ def initconfig_hardware_entries(self):
                     cmake_cache_path("HIP_PATH", spec["hip"].prefix),
                     cmake_cache_path("HIP_CLANG_INCLUDE_PATH", spec["llvm-amdgpu"].prefix.include),
                     cmake_cache_path("ROCM_PATH", spec["llvm-amdgpu"].prefix),
-                    cmake_cache_string("AMDGPU_TARGETS", spec.variants["amdgpu_target"].value)
+                    cmake_cache_string("AMDGPU_TARGETS", spec.variants["amdgpu_target"].value[0])
```
gardner48, line 701: Is `[0]` necessary or will this work with multiple `amdgpu_target` values?

.gitlab/spack_packages/sundials/package.py
```diff
@@ -788,23 +844,30 @@ def initconfig_package_entries(self):
...
             if "camp" in spec:
                 entries.append(cmake_cache_path("camp_DIR", spec["camp"].prefix.lib.cmake + '/camp'))
+            if "+rocm" in spec:
+                entries.append(cmake_cache_string("SUNDIALS_RAJA_BACKENDS", "HIP"))
 
         # Building with SuperLU_DIST
         if "+superlu-dist" in spec:
-            if spec.satisfies("@6.4.0:"):
+            #if spec.satisfies("@6.4.0:"):
+            if False:
```
gardner48, line 853: Can this be reverted to `if spec.satisfies("@6.4.0:"):`

cmake/macros/SundialsAddTest.cmake
```diff
@@ -135,8 +135,13 @@ macro(SUNDIALS_ADD_TEST NAME EXECUTABLE)
...
 
       # check if this test is run with MPI and set the MPI run command
-      if((SUNDIALS_ADD_TEST_MPI_NPROCS) AND (MPIEXEC_EXECUTABLE))
-        set(RUN_COMMAND "${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${SUNDIALS_ADD_TEST_MPI_NPROCS} ${MPIEXEC_PREFLAGS}")
+      if ((SUNDIALS_ADD_TEST_MPI_NPROCS) AND ((MPIEXEC_EXECUTABLE) OR (SUNDIALS_TEST_MPIRUN_COMMAND)))
+        if (SUNDIALS_TEST_MPIRUN_COMMAND STREQUAL "")
+          set(RUN_COMMAND "${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${SUNDIALS_ADD_TEST_MPI_NPROCS} ${MPIEXEC_PREFLAGS}")
+        elseif (MPIEXEC_EXECUTABLE)
+          set(RUN_COMMAND "${SUNDIALS_TEST_MPIRUN_COMMAND} ${MPIEXEC_NUMPROC_FLAG} ${SUNDIALS_ADD_TEST_MPI_NPROCS} ${MPIEXEC_PREFLAGS}")
+        endif()
```
gardner48, line 143:
```suggestion
        if(SUNDIALS_TEST_MPIRUN_COMMAND)
          set(RUN_COMMAND "${SUNDIALS_TEST_MPIRUN_COMMAND} ${MPIEXEC_NUMPROC_FLAG} ${SUNDIALS_ADD_TEST_MPI_NPROCS} ${MPIEXEC_PREFLAGS}")
        elseif(MPIEXEC_EXECUTABLE)
          set(RUN_COMMAND "${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${SUNDIALS_ADD_TEST_MPI_NPROCS} ${MPIEXEC_PREFLAGS}")
        endif()
```

cmake/macros/SundialsAddTest.cmake
```diff
@@ -174,11 +179,16 @@ macro(SUNDIALS_ADD_TEST NAME EXECUTABLE)
       endif()
 
       # check if this test is run with MPI and add the test run command
-      if((SUNDIALS_ADD_TEST_MPI_NPROCS) AND (MPIEXEC_EXECUTABLE))
+      if ((SUNDIALS_ADD_TEST_MPI_NPROCS) AND ((MPIEXEC_EXECUTABLE) OR (SUNDIALS_TEST_MPIRUN_COMMAND)))
```
gardner48, line 182:
```suggestion
      if((SUNDIALS_ADD_TEST_MPI_NPROCS) AND ((MPIEXEC_EXECUTABLE) OR (SUNDIALS_TEST_MPIRUN_COMMAND)))
```

cmake/macros/SundialsAddTest.cmake
```diff
@@ -174,11 +179,16 @@ macro(SUNDIALS_ADD_TEST NAME EXECUTABLE)
...
         if(MPIEXEC_PREFLAGS)
           string(REPLACE " " ";" PREFLAGS "${MPIEXEC_PREFLAGS}")
         endif()
-        add_test(NAME ${NAME} COMMAND ${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${SUNDIALS_ADD_TEST_MPI_NPROCS} ${PREFLAGS} $<TARGET_FILE:${EXECUTABLE}> ${TEST_ARGS})
+        if (SUNDIALS_TEST_MPIRUN_COMMAND STREQUAL "")
+          add_test(NAME ${NAME} COMMAND ${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${SUNDIALS_ADD_TEST_MPI_NPROCS} ${PREFLAGS} $<TARGET_FILE:${EXECUTABLE}> ${TEST_ARGS})
+        else()
+          string(REPLACE " " ";" MPI_EXEC_ARGS "${SUNDIALS_TEST_MPIRUN_COMMAND}")
+          add_test(NAME ${NAME} COMMAND ${MPI_EXEC_ARGS} ${MPIEXEC_NUMPROC_FLAG} ${SUNDIALS_ADD_TEST_MPI_NPROCS} ${PREFLAGS} $<TARGET_FILE:${EXECUTABLE}> ${TEST_ARGS})
+        endif()
```
gardner48, line 191:
```suggestion
        if (SUNDIALS_TEST_MPIRUN_COMMAND)
          add_test(NAME ${NAME} COMMAND ${SUNDIALS_TEST_MPIRUN_COMMAND} ${MPIEXEC_NUMPROC_FLAG} ${SUNDIALS_ADD_TEST_MPI_NPROCS} ${PREFLAGS} $<TARGET_FILE:${EXECUTABLE}> ${TEST_ARGS})        
        else()
          add_test(NAME ${NAME} COMMAND ${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${SUNDIALS_ADD_TEST_MPI_NPROCS} ${PREFLAGS} $<TARGET_FILE:${EXECUTABLE}> ${TEST_ARGS})
        endif()
```

cmake/SundialsBuildOptionsPre.cmake
```diff
@@ -294,3 +294,5 @@ sundials_option(SUNDIALS_TEST_DEVTESTS BOOL
 # Include unit tests in regression tests
 sundials_option(SUNDIALS_TEST_UNITTESTS BOOL
   "Include unit tests in make test" OFF ADVANCED)
+
+sundials_option(SUNDIALS_TEST_MPIRUN_COMMAND STRING "Job scheduler or mpirun command used to launch SUNDIALS MPI tests." "" ADVANCED)
```
gardner48, line 298: Add new line at end of file

gardner48 (changes requested):

.gitlab/lassen-jobs.yml
gardner48: I think this is causing the most recent failure in the lassen tests

.gitlab/lassen-jobs.yml
balos1: Even though we dont use CUDA in PETSc, we should keep `petsc+cuda` since this is how xsdk will build it.

.gitlab/quartz-jobs.yml
balos1: Both are indeed necessary due to bugs in spack.

.gitlab/quartz-jobs.yml
balos1: There is a bug in spack@0.19.1 that causes a very old suite-sparse to be picked up unless we specify a version.

.gitlab/quartz-jobs.yml
balos1: There is a bug in spack@0.19.1 that causes a very old suite-sparse to be picked up unless we specify a version.

.gitlab/spack_packages/netlib-lapack/package.py
balos1: Yes, it is needed. Some of the packages use `netlib-lapack`, and we need a fix that is not in spack 0.19.1.

.gitlab/spack_packages/sundials/package.py
balos1: No, we need to always use the other way for now due to the issue with the RPATH in superlu-dist. I think leaving this block in here with `if False:` (or commented out) is fine, but we should add a comment on why we are doing so.

.gitlab/build_and_test.sh
balos1: Probably, but not worth doing this until we update to the latest uberenv which has significant changes including the use of environments.

.gitlab/lassen-jobs.yml
balos1: Also ginkgo is being added in a separate PR woth kokkos.

.gitlab/spack_packages/sundials/package.py
yu-nix: I'll check if it's necessary, though I recall earlier there were issues with parsing the target incorrectly resulting in an empty entry

yu-nix: There is a fix to this, which is to do `",".join(spec.variants["amdgpu_target"].value`

.gitlab/radiuss-spack-configs
yu-nix: The corona-ci branch contains additional rocm packages, so the branch will need to be merged in.

gardner48: Ok, we should probably do a PR to update `main` in that repo before before merging this PR so the submodule can point to most recent commit on `main`. I've had some issues with submodules pointing to commits on branches that eventually get deleted when they're merged.

.gitlab/spack_packages/netlib-lapack/package.py
gardner48: Interesting, do those packages depend specifically on `netlib-lapack` rather than the virtual `lapack` or is spack picking a different LAPACK provider for some reason?

.gitlab/spack_packages/sundials/package.py
gardner48: Should these be comma separated or semi-colon separated?

.gitlab/radiuss-spack-configs
balos1: @gardner48 https://github.com/sundials-codes/radiuss-spack-configs/pull/1

.gitlab/spack_packages/sundials/package.py
balos1: I believe cmake will want semi-colon separated.

yu-nix: The ROCm documentation had them as commas, but it might have been in a different context

.gitlab/radiuss-spack-configs
gardner48: Just merged the PR, so you can update to the head of `main` for the submodule now

.gitlab/spack_packages/sundials/package.py
yu-nix: I can add both to the variant, and define jsrun and srun for the two schedulers, though I think the system currently defaults to mpiexec and srun for the two currently, but it'd be good to be explicit about that.

cmake/macros/SundialsAddTest.cmake
yu-nix: For flux, the command to run is `flux run` which if passed in without parsing it out leads to the system searching for `flux run` as a program rather than executing run with flux.

gardner48: Ok, must be a problem with CMake's behavior with space vs semi-colon separate strings/lists (this is often an issue). I'd still suggest updating to use logic with `if(SUNDIALS_TEST_MPIRUN_COMMAND)` rather than `if(SUNDIALS_TEST_MPIRUN_COMMAND STREQUAL "")`

yu-nix: The alternative is to pass in commands as `flux;run` rather than `flux run`

gardner48: I think passing `flux run` and dealing with it internally like you've done is the more natural way to go.

balos1 (approved):

.gitlab/corona-jobs.yml
```diff
@@ -30,7 +30,7 @@ corona_clang_hip:
 corona_clang_hip_tpls:
   parallel:
     matrix:
-      - COMPILER_SPEC: clang@15.0.0
+      - COMPILER_SPEC: rocmcc@5.4.1
```
gardner48, line 33: Why the change from `clang` to `rocmcc`?

gardner48 (approved):

.gitlab/corona-jobs.yml
yu-nix: The radiuss-spack-configs renamed the clang compilers to their rocmcc version.

PR #267, opened 2023-03-28
Default to linear interpolant with first order method in ARKODE

gardner48:
Currently ARKODE defaults to a constant interpolant when using a first order method. This can lead to unexpected differences in outputs when verifying interfaces to ARKODE against manually implemented first order methods. This PR updates the default interpolation degree with first order methods only so a linear interpolant is used and the expected values are returned at the ends of a time interval.


balos1 (approved):
LGTM

drreynolds (approved):

PR #281, opened 2023-06-07
Feature/tidy and format

balos1:
Apply clang-tidy for braces and clang-format to everything.


drreynolds:
I'm noticing that this PR excludes both C++ and Fortran files.  Are "tidying" processes planned for those as well?

drreynolds:
I'm also finding that the automated tidying process has some difficulty with block comments, where we indent to provide some structure/formatting.  I'm catching a bunch of those while reviewing, but it's likely that I'll miss some.

balos1:
> I'm also finding that the automated tidying process has some difficulty with block comments, where we indent to provide some structure/formatting. I'm catching a bunch of those while reviewing, but it's likely that I'll miss some.

Rather than point out all of them, can you point me to a specific one to see if I can tweak the format file to address it.

drreynolds:
@balos1: for example, in `cvode_impl.h` a too-long comment line was tidied into this:
```
 * SMALL_NEF_DEFAULT   if small_nef <= nef <= MXNEF1, then eta =  SUNMIN(eta,
 * eta_max_ef) ONEPSM (1+epsilon)  used in testing if the step size is below its
 * bound
```
when it should have instead been this:
```
 * SMALL_NEF_DEFAULT   if small_nef <= nef <= MXNEF1,  
 *                     then eta =  SUNMIN(eta, eta_max_ef) 
 * ONEPSM (1+epsilon)  used in testing if the step size is below its
 *                     bound
```

balos1:
@drreynolds Thanks. I will update the format file to leave block comments alone and re-apply.

PR #289, opened 2023-06-19
Feature/oo adaptivity

drreynolds:
New object-oriented infrastructure for time step adaptivity controllers, and time step heuristic constraints.  These are currently used in ARKODE.  In the not-distant future, we can implement additional step/order controllers within this existing infrastructure to support CVODE(S) and IDA(S) (and maybe even ARKODE).


drreynolds:
GitHub just sent me a message saying that this PR failed two tests: `ark_kpr_Mt_2_8_0_-10` and `ark_kpr_Mt_2_4_0_-10_0`.

The output from GitHub shows that the answers from this PR are uniformly more accurate than in the answer file, so it seems like I should update the `.out` files for those two examples in the repository.  Given that these tests did not fail on my machine, does anyone know how to get the `.out` files from the CI?

balos1:
Looks like it just the single-precision case that is failing. You can get those output files by downloading the "build files" under artifacts on the page https://github.com/LLNL/sundials/actions/runs/5316007238?pr=289. The generated out files are in `Testing/output`. 

If you need to update any out files, you would do so in https://github.com/sundials-codes/answers.

drreynolds:
Thanks for noticing the single/double, @balos1.  I had tried with double and saw identical results on my workstation, making it challenging to identify the precise cause of the accuracy improvement.  I'll investigate whether any differences show up on my computers when using single.

However, in any case I would not be at all surprised if this PR induces minor floating-point accumulation differences from before.  While I did my best to surgically separate error controllers from heuristic constraints, some operations must be performed in a different order, resulting in tiny differences that can eventually grow.  Based on the rather thorough testing that I've done so far, I do not believe that there are any remaining mathematical problems, but like every time that we've had to adjust the order of otherwise-identical operations, I'd expect accumulation effects to cause minor differences.

Steven-Roberts:
Moving the fix in #321 to the discussion here. Here are some issues ordered from highest to lowest importance I think we should address in this PR:

- The current factor scaling a step size `err^(-1/k)`, where `k=p or q`, is slightly off from the conventional value
    - I propose that the error controllers have `p` and `q` as properties, and we remove `pq`. If we consider the I controller, the standard step size factor is `err^(-1/(min(p,q) + 1))`, which I think we can use for all methods (ERK, DIRK, ARK). This expression is valid for edge cases where the embedding is an order higher (like Runge–Kutta–Fehlberg methods) or [two order lower](https://epubs.siam.org/doi/pdf/10.1137/S1064827596302230). The current setup does not handle the later and may cause confusion in that the `p` property of the controller can be the primary or embedded order. The bias and safety factors can be relaxed once this change is made. The concern of order reduction  raised by @drreynolds could be addressed by a user setting `p` and `q` values different than the classical order.
 - @gardner48 suggested we consider no deadzone by default. [Söderlind claims deadzones hurt performance](https://link.springer.com/content/pdf/10.1023/A:1021160023092.pdf)
 - The convention in the Runge-Kutta literature is $p$ is the order of the primary method and $\widehat{p}$ is the order of the embedding. If there is interest in making this refactor, now sees like the time.

include/arkode/arkode_arkstep.h
```diff
@@ -159,6 +161,7 @@ SUNDIALS_EXPORT int ARKStepRootInit(void *arkode_mem, int nrtfn,
 
 /* Optional input functions -- must be called AFTER ARKStepCreate */
 SUNDIALS_EXPORT int ARKStepSetDefaults(void* arkode_mem);
+SUNDIALS_DEPRECATED_EXPORT_MSG("use individual controls instead")
 SUNDIALS_EXPORT int ARKStepSetOptimalParams(void *arkode_mem);
```
gardner48, line 165: Here and elsewhere `SUNDIALS_DEPRECATED_EXPORT_MSG` should replace `SUNDIALS_EXPORT`

gardner48: Does this function need to be deprecated with the change to controller objects or is being deprecated on its own?

include/suncontrol/suncontrol_i.h
```diff
@@ -0,0 +1,72 @@
...
+
+/* ------------------
+ * Exported Functions
+ * ------------------ */
+
+SUNDIALS_EXPORT
+SUNControl SUNControlI(SUNContext sunctx);
+SUNDIALS_EXPORT
+int SUNControlI_SetParams(SUNControl C, sunbooleantype pq,
+                          realtype k1);
```
gardner48, line 48: Here and for the other controllers, I think `SetParams` should following the same naming convention as the other methods i.e., `SUNControl_SetParams_<implementation name>`

include/suncontrol/suncontrol_i.h
```diff
@@ -0,0 +1,72 @@
...
+};
+
+typedef struct _SUNControlContent_I *SUNControlContent_I;
+
+/* ------------------
+ * Exported Functions
+ * ------------------ */
+
+SUNDIALS_EXPORT
+SUNControl SUNControlI(SUNContext sunctx);
```
gardner48, line 45: It's not mentioned in the dev guide but we should probably follow the same naming convention for constructors as well i.e., `SUNControl_<implementation name>`, here and for the other controllers.

include/sundials/sundials_control.h
```diff
@@ -0,0 +1,248 @@
...
+ *    NONE    - empty controller (does nothing)
+ *    H       - controls a single-rate step size
+ *    HQ      - controls a single-rate step size and method order
+ *    MRI_H   - controls two multirate step sizes
+ *    MRI_TOL - controls slow and fast relative tolerances
+ * ----------------------------------------------------------------- */
+
+typedef enum
+{
+  SUNDIALS_CONTROL_NONE,
```
gardner48, line 44: Would the `NONE` controller be the same as a fixed step size and order controller? In that case do we need a `NONE` type?

include/sundials/sundials_control.h
```diff
@@ -0,0 +1,248 @@
...
+   'content' field, and a pointer to a structure of
+   operations corresponding to that implementation. */
+struct _generic_SUNControl
+{
+  void* content;
+  SUNControl_Ops ops;
+  SUNContext sunctx;
+#ifdef __cplusplus
+  _generic_SUNControl() = default;
+#endif
```
gardner48, line 115: I think this can be removed (see #329)

doc/shared/suncontrol/SUNControl_Description.rst
```diff
@@ -0,0 +1,480 @@
...
+   :param C: the :c:type:`SUNControl` object.
+   :return: :c:type:`SUNControl_ID` type identifier.
+
+   Usage:
+
+   .. code-block:: c
+
+      SUNControl_ID id = SUNControlGetID(C);
+
+.. c:function:: void SUNControlDestroy(SUNControl C)
```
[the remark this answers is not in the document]
balos1: We should return a flag. I think we need to move to always returning a flag from our functions for consistency.

doc/shared/suncontrol/SUNControl_Description.rst
```diff
@@ -0,0 +1,480 @@
...
+
+   See the top-level LICENSE and NOTICE files for details.
+
+   SPDX-License-Identifier: BSD-3-Clause
+   SUNDIALS Copyright End
+   ----------------------------------------------------------------
+
+.. _SUNControl.Description:
+
+The SUNControl API
```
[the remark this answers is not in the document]
balos1: How about `SUNAdaptController` and then the heuristics class could be `SUNAdaptHeuristics`.

doc/shared/suncontrol/SUNControl_Description.rst
```diff
@@ -0,0 +1,480 @@
...
+SUNControl Operations
+---------------------
+
+The base SUNControl class defines and implements all SUNControl functions. Most
+of these routines are merely wrappers for the operations defined by a particular
+SUNControl implementation, which are accessed through the *ops* field of the
+``SUNControl`` structure. However, the base SUNControl class also provides the
+convenience routine
+
+.. c:function:: SUNControl SUNControlNewEmpty(SUNContext sunctx)
```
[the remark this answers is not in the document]
balos1: Indeed we should stick to the new naming convention for new classes.

doc/shared/suncontrol/SUNControl_Description.rst
```diff
@@ -0,0 +1,480 @@
...
+should note its type, so that integrators will understand the types of
+adaptivity that the controller is designed to perform. These are encoded in the
+following set of SUNControl types:
+
+.. _SUNControl.Description.controllerIDs:
+.. table:: Identifiers associated with SUNControl implementations
+   :align: center
+
+   ========================  =====================================================
+   SUNControl ID             Intended form of adaptivity
```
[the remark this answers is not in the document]
balos1: > Some more naming convention questions, should this be `SUNControl_Type` or `SUNControl_Type` and `SUNDIALS_CONTROL_NONE` or `SUN_CONTROL_NONE`?

The naming convention we defined implies it could be either, however I prefer `SUN_ADAPTCONTROLLER_NONE` and `SUNAdaptController_Type` (another comment chain below explains why `SUNAdaptController`.

src/arkode/arkode.c
```diff
@@ -123,15 +126,27 @@ ARKodeMem arkCreate(SUNContext sunctx)
   /* No user_data pointer yet */
   ark_mem->user_data = NULL;
 
-  /* Allocate step adaptivity structure and note storage */
-  ark_mem->hadapt_mem = arkAdaptInit();
-  if (ark_mem->hadapt_mem == NULL) {
+  /* Allocate default step controller (PID) and note storage */
+  ark_mem->hcontroller = SUNControlPID(sunctx);
```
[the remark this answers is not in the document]
gardner48: Should the stepper's own the controllers to allow for different defaults? How will this work in the MRI case, does the MRI stepper override the ARKODE default controller?

src/arkode/arkode_arkstep_io.c
```diff
@@ -1618,17 +1659,19 @@ int ARKStepGetTimestepperStats(void *arkode_mem, long int *expsteps,
...
   *fe_evals      = step_mem->nfe;
   *fi_evals      = step_mem->nfi;
   *nlinsetups    = step_mem->nsetups;
   *netfails      = ark_mem->netf;
 
+  /* get expsteps and accsteps from heuristics object */
+  retval = SUNHeuristicsGetNumExpSteps(ark_mem->hconstraints, expsteps);
+  if (retval != ARK_SUCCESS) { return(ARK_HEURISTICS_ERR); }
+  retval = SUNHeuristicsGetNumAccSteps(ark_mem->hconstraints, accsteps);
+  if (retval != ARK_SUCCESS) { return(ARK_HEURISTICS_ERR); }
```
gardner48, line 1673: Are these counters retained across are reset/resize calls and set to zero on re-initialization? If the controller is replaced during a run are these counts lost?

src/arkode/arkode_erkstep_io.c
```diff
@@ -234,6 +207,244 @@ int ERKStepGetNumRelaxSolveIters(void* arkode_mem, long int* iters)
   return arkRelaxGetNumRelaxSolveIters(arkode_mem, iters);
 }
 
+/*===============================================================
+  DEPRECATED ERKStep optional input/output functions
```
gardner48, line 211: I did not see similar wrappers in ARKStep for deprecated functions

src/arkode/arkode_erkstep_io.c
```diff
@@ -265,19 +477,55 @@ int ERKStepSetDefaults(void* arkode_mem)
     return(retval);
   }
 
+  /* Remove current SUNHeuristics object, and replace with "Default" */
```
gardner48, line 480: I didn't see similar updates to the ARKStep version of this function (or is this covered by the ARKODE defaults?)

src/arkode/arkode_erkstep_io.c
```diff
@@ -546,15 +794,62 @@ int ERKStepGetTimestepperStats(void *arkode_mem, long int *expsteps,
...
+  /* set integrator outputs */
   *attempts = ark_mem->nst_attempts;
   *fevals   = step_mem->nfe;
   *netfails = ark_mem->netf;
 
+  /* get expsteps and accsteps from heuristics object */
+  retval = SUNHeuristicsGetNumExpSteps(ark_mem->hconstraints, expsteps);
+  if (retval != ARK_SUCCESS) { return(ARK_HEURISTICS_ERR); }
+  retval = SUNHeuristicsGetNumAccSteps(ark_mem->hconstraints, accsteps);
+  if (retval != ARK_SUCCESS) { return(ARK_HEURISTICS_ERR); }
```
gardner48, line 806: Same questions about retaining/resetting counters

src/arkode/arkode_impl.h
```diff
@@ -253,6 +253,25 @@ int arkInterpEvaluate(void* arkode_mem, ARKInterp interp,
...
+};
+
+typedef struct ARKUserStabilityDataMem *ARKUserStabilityData;
+
+void* ARKUserStability(void* arkode_mem, ARKExpStabFn EStab,
+                       void* estab_data);
+
+void ARKUserStability_Free(void* sdata);
+
+int ARKControlExpStab(realtype *hstab, void* user_data);
```
gardner48, line 273: See earlier comment on if explicit stability control should be it's an `SUNControl` implementation

src/arkode/arkode_io.c
```diff
@@ -588,6 +611,35 @@ int arkSetFixedStep(void *arkode_mem, realtype hfixed)
...
+    ark_mem->liw -= leniw;
+    ark_mem->lrw -= lenrw;
+  }
+  SUNControlDestroy(ark_mem->hcontroller);
+  ark_mem->hcontroller = NULL;
+
+  /* If re-enabling time adaptivity, create default PID controller
+     and attach object to ARKODE */
+  SUNControl C = NULL;
+  if (hfixed == 0)
```
gardner48, line 626:
```suggestion
  if (hfixed == ZERO)
```

src/arkode/arkode_io.c
```diff
@@ -588,6 +611,35 @@ int arkSetFixedStep(void *arkode_mem, realtype hfixed)
...
   ark_mem = (ARKodeMem) arkode_mem;
 
+  /* Remove current SUNControl object */
+  retval = SUNControlSpace(ark_mem->hcontroller, &lenrw, &leniw);
+  if (retval == SUNCONTROL_SUCCESS) {
+    ark_mem->liw -= leniw;
+    ark_mem->lrw -= lenrw;
+  }
+  SUNControlDestroy(ark_mem->hcontroller);
+  ark_mem->hcontroller = NULL;
```
gardner48, line 621: Is it necessary to destroy the controller here or should this only happen if `hfixed == 0` and a controller does not already exist?

src/arkode/arkode_io.c
```diff
@@ -819,389 +877,220 @@ int arkSetMaxNumConstrFails(void *arkode_mem, int maxfails)
...
-  arkSetAdaptivityMethod:
+  arkSetAdaptivityMethod:  ***DEPRECATED***
 
   Specifies the built-in time step adaptivity algorithm (and
   optionally, its associated parameters) to use.  All parameters
   will be checked for validity when used by the solver.
+
+  Users should transition to constructing non-default SUNControl
+  objects directly, and providing those directly to the integrator
+  via the time-stepping module *SetController routines.
```
gardner48, line 888: See earlier comment about deprecating the current signature of the method but switching to a new version later that will allow users to select from the provided controllers without having to construct one themselves.

src/arkode/arkode_io.c
```diff
@@ -819,389 +877,220 @@ int arkSetMaxNumConstrFails(void *arkode_mem, int maxfails)
...
+  SUNControl C = NULL;
+  switch (imethod) {
+  case (ARK_ADAPT_PID):
+    C = SUNControlPID(ark_mem->sunctx);
+    if (C == NULL) {
+      arkProcessError(ark_mem, ARK_MEM_FAIL, "ARKODE", "arkSetAdaptivityMethod",
+                      "SUNControlPID allocation failure");
+      return(ARK_MEM_FAIL);
+    }
+    retval = SUNControlPID_SetParams(C, pq, k1, k2, k3);
```
gardner48, line 935: Here and for the other methods below, should this only be called if `adapt_params != NULL`?

src/arkode/arkode_io.c
```diff
@@ -1210,19 +1099,39 @@ int arkSetMaxCFailGrowth(void *arkode_mem, realtype etacf)
...
+    return(ARK_MEM_FAIL);
+  }
+
+  /* Attach ARKODE function and wrapper structure to heuristics object */
+  retval = SUNHeuristicsSetExpStabFn(ark_mem->hconstraints,
+                                     ARKControlExpStab, wrapper_data);
+  if (retval != SUNHEURISTICS_SUCCESS) {
+    arkProcessError(ark_mem, ARK_HEURISTICS_ERR, "ARKODE",
+                    "arkSetStabilityFn", "SUNHeuristicsSetExpStabFn failure");
+    return(ARK_HEURISTICS_ERR);
```
gardner48, line 1134: Should this be in an `else` to `if (EStab == NULL)` above?

src/arkode/arkode_user_controller.h
```diff
@@ -0,0 +1,78 @@
...
+
+#ifdef __cplusplus  /* wrapper to enable C++ usage */
+extern "C" {
+#endif
+
+/* -------------------------------------------
+ * ARKUserControl implementation of SUNControl
+ * ------------------------------------------- */
+
+struct _ARKUserControlContent {
```
gardner48, line 34: See earlier comment on reserved identifiers

src/arkode/arkode_user_controller.c
```diff
@@ -0,0 +1,183 @@
...
+ * ----------------------------------------------------------------- */
+
+SUNControl_ID SUNControlGetID_ARKUserControl(SUNControl C)
+{ return SUNDIALS_CONTROL_H; }
+
+int SUNControlEstimateStep_ARKUserControl(SUNControl C, realtype h,
+                                          realtype dsm, realtype* hnew)
+{
+  /* call user-provided function to compute new step */
+  int retval = SC_HADAPT(C)(SC_ARKMEM(C)->ycur, SC_ARKMEM(C)->tn, h, SC_HP(C),
```
gardner48, line 112: Does this need 
```
  ttmp = (dsm <= ONE) ? ark_mem->tn + ark_mem->h : ark_mem->tn;
  nsttmp = (dsm <= ONE) ? ark_mem->nst+1 : ark_mem->nst;
```
to match what would have been passed before?

gardner48 (changes requested):
Finished pass over arkode source

examples/arkode/CXX_parallel/ark_diffusion_reaction_p.cpp
```diff
@@ -76,6 +76,12 @@
 #include "sunlinsol/sunlinsol_pcg.h"
 #include "sunlinsol/sunlinsol_spgmr.h"
 #include "sunnonlinsol/sunnonlinsol_fixedpoint.h"
+#include "suncontrol/suncontrol_pid.h"
+#include "suncontrol/suncontrol_pi.h"
+#include "suncontrol/suncontrol_i.h"
+#include "suncontrol/suncontrol_expgus.h"
+#include "suncontrol/suncontrol_impgus.h"
+#include "suncontrol/suncontrol_imexgus.h"
```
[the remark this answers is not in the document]
balos1: I agree with this.

examples/arkode/CXX_parallel/ark_diffusion_reaction_p.cpp
```diff
@@ -1016,9 +1022,17 @@ static int SetupARK(SUNContext ctx, UserData* udata, N_Vector u,
...
-                                      SUNFALSE, NULL);
-    if (check_flag(&flag, "ARKStepSetAdaptivityMethod", 1)) return 1;
+    SUNControl Ctrl = NULL;
+    switch (udata->controller) {
+    case (ARK_ADAPT_PID):      Ctrl = SUNControlPID(ctx);     break;
+    case (ARK_ADAPT_PI):       Ctrl = SUNControlPI(ctx);      break;
+    case (ARK_ADAPT_I):        Ctrl = SUNControlI(ctx);       break;
+    case (ARK_ADAPT_EXP_GUS):  Ctrl = SUNControlExpGus(ctx);  break;
+    case (ARK_ADAPT_IMP_GUS):  Ctrl = SUNControlImpGus(ctx);  break;
+    case (ARK_ADAPT_IMEX_GUS): Ctrl = SUNControlImExGus(ctx); break;
```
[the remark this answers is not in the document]
balos1: > It might be worth deprecating the current `ARKStepSetAdaptivityMethod` signature but changing it later so users can select the native controllers without having to construct an object (unless they want to alter the default parameters).

I second this suggestion.

doc/shared/sunheuristics/SUNHeuristics_Description.rst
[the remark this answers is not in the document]
balos1: I believe the benefit of having the heuristics class is that it makes it possible to easily turn off or modify the heuristics further then than just adjusting the parameter values (although the I am less clear on how useful the latter is).  For example, we can implement a "identity" heuristics module which does nothing but produce the exact time step size received from the controller.  

One use case for turning off the heuristics is SPRKStep. To implement a controller that is compatible with the symplectic method, I need to implement a `SUNControl` that produces a time step size that is precisely what the the time stepper will use. If we do not have a way to turn off the heuristics, then they can cause the time step size produced by the controller to be modified then given to the stepper, which breaks the symplectic-compatible controller.

src/arkode/arkode_impl.h
```diff
@@ -355,8 +374,6 @@ struct ARKodeMemRec
   /* Time step data */
   realtype hin;                /* initial step size                        */
   realtype h;                  /* current step size                        */
-  realtype hmin;               /* |h| >= hmin                              */
```
balos1, line 358: `hmin` is used in arkode_relaxation.c, so this being removed is now causing a compilation error.

doc/shared/suncontrol/SUNControl_Description.rst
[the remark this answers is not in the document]
drreynolds: I'm all for consistency, but not for blanket rules that do not make sense.  In C, we cannot catch any errors that arise from calling `free`, and thus all such routines would return `SUCCESS` arbitrarily (potentially giving the false impression if indeed `free` malfunctioned).

doc/shared/suncontrol/SUNControl_Description.rst
```diff
@@ -0,0 +1,480 @@
...
+   :return: error code indicating success failure
+            (see :numref:`SUNControl.Description.errorCodes`).
+
+   Usage:
+
+   .. code-block:: c
+
+      retval = SUNControlEstimateStepAndOrder(C, hcur, qcur, dsm, &hnew, &qnew);
+
+.. c:function:: int SUNControlEstimateMRISteps(SUNControl C, realtype H, realtype h, realtype DSM, realtype dsm, realtype* Hnew, realtype *hnew)
```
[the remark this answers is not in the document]
drreynolds: Potentially, however I find it unlikely that with highly nested MRI calculations, anyone would want "adaptivity" that forces fixed values for `h[0]`, `h[1]`, ... `h[maxl]` throughout the entire "slow" MRI step.

Instead, I envision that for nested MRI it should be preferable to use the H-Tol controllers, since these may naturally be nested without any changes, and would allow varying the "fastest" step size throughout a given "slowest" step.  

If you would like to investigate MRI adaptivity with multiple fixed step sizes values throughout all nested levels within a given slow step, then I can update this to be more flexible.  However, I do not intend to pursue that direction.

doc/shared/suncontrol/SUNControl_Description.rst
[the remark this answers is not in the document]
balos1: It is possible that some implementation of `SUNControl` may do more than calling `free`. Also, sometimes we do things in the `Destroy` routine that isnt just calling `free`.

doc/shared/suncontrol/SUNControl_Description.rst
```diff
@@ -0,0 +1,480 @@
...
+   :return: error code indicating success failure
+            (see :numref:`SUNControl.Description.errorCodes`).
+
+   Usage:
+
+   .. code-block:: c
+
+      retval = SUNControlEstimateMRISteps(C, Hcur, hcur, DSM, &Hnew, &hnew);
+
+.. c:function:: int SUNControlEstimateStepTol(SUNControl C, realtype H, realtype tolfac, realtype DSM, realtype *Hnew, realtype* tolfacnew)
```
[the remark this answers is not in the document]
drreynolds: Ah, this type is already "nestable" in that a given MRIStep instance (at any level of the nested hierarchy) need only receive an accuracy tolerance as *input* from its next-slower level to know what is expected from it.  That MRIStep instance can then adapt its own `H` value, and subsequently request changes in the tolerance for the next-finer level.

doc/shared/suncontrol/SUNControl_Description.rst
```diff
@@ -0,0 +1,480 @@
...
+   :return: error code indicating success failure
+            (see :numref:`SUNControl.Description.errorCodes`).
+
+   Usage:
+
+   .. code-block:: c
+
+      retval = SUNControlSetErrorBias(C, 1.2);
+
+.. c:function:: int SUNControlUpdate(SUNControl C, realtype h, realtype dsm)
```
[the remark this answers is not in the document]
drreynolds: Great question, and I bet that it would be needed for something like CVODE or IDA.  That said, I would prefer those to use a different update function so as not to over-complicate things (e.g., a variable-order MRI controller would then also need P and p added to the list).  I recommend that we expand this API when/if we create H-p controllers, and since all implementations use `NewEmpty` then those updates wouldn't require major version number changes.

doc/shared/suncontrol/SUNControl_Description.rst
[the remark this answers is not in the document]
drreynolds: Given that "controller" has a specific connotation in this field regarding control-theoretic error-based adaptivity, I feel like extending the name to `SUNAdaptController` only increases typing burden.  Given that some of these function names are already incredibly long, I am reticent to make it longer unnecessarily.

That said, there are many forms of heuristics throughout SUNDIALS, so the base name `SUNHeuristics` may not be sufficiently specific (although it's already long enough).  Since it focuses on step size heuristics, then what do you think about `SUNTimestepHeuristics`?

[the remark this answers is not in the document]
balos1: I like either `SUNTImestepHeuristics` or `SUNAdaptHeuristics`.  I dislike `SUNControl` but would be OK with `SUNController`. However, I most prefer `SUNAdaptController` or `SUNTimestepController` but I tend to prefer descriptive names over saving a few keystrokes.  Also, not every user of SUNDIALS is in tune with the time integration field and they might not associate "controller" with time-step adaptivity automatically like we do.

src/arkode/arkode.c
```diff
@@ -2972,14 +3006,17 @@ int arkCheckConstraints(ARKodeMem ark_mem, int *constrfails, int *nflag)
...
-  /* Return with error if |h| == hmin */
-  if (SUNRabs(ark_mem->h) <= ark_mem->hmin*ONEPSM) return(ARK_CONSTR_FAIL);
-
-  /* Reduce h by computing eta = h'/h */
+  /* Recommend step size adjustment to satisfy constraints */
   N_VLinearSum(ONE, ark_mem->yn, -ONE, ark_mem->ycur, tmp);
   N_VProd(mm, tmp, tmp);
-  ark_mem->eta = RCONST(0.9)*N_VMinQuotient(ark_mem->yn, tmp);
-  ark_mem->eta = SUNMAX(ark_mem->eta, TENTH);
+  hnew = ark_mem->h * SUNMAX(RCONST(0.9)*N_VMinQuotient(ark_mem->yn, tmp), TENTH);
```
Steven-Roberts, line 3012: Could this heuristic be handled by the SUNHeuristic object?

include/sunheuristics/sunheuristics_default.h
```diff
@@ -0,0 +1,120 @@
...
+  realtype     etamax;     /* eta <= etamax                              */
+  realtype     etamx1;     /* max step size change on first step         */
+  realtype     etamxf;     /* h reduction factor on multiple error fails */
+  realtype     etamin;     /* eta >= etamin on error test fail           */
+  int          small_nef;  /* bound to determine 'multiple' above        */
+  realtype     etacf;      /* h reduction factor on nonlinear conv fail  */
+  SUNExpStabFn expstab;    /* step stability function                    */
+  void*        estab_data; /* user pointer passed to expstab             */
+  realtype     cfl;        /* cfl safety factor                          */
+  realtype     safety;     /* step safety factor                         */
```
Steven-Roberts, line 43: The heuristics object handles the safety factor, but the controller handles a bias. It is a bit strange they are in different classes when they are so similar.

src/sunheuristics/default/sunheuristics_default.c
```diff
@@ -0,0 +1,488 @@
...
+#define SH_UBOUND(H)      ( SH_CONTENT(H)->ubound )
+#define SH_NST_ACC(H)     ( SH_CONTENT(H)->nst_acc )
+#define SH_NST_EXP(H)     ( SH_CONTENT(H)->nst_exp )
+
+/* ------------------
+ * Default parameters
+ * ------------------ */
+
+#define DEFAULT_CFLFAC     RCONST(0.5)
+#define DEFAULT_SAFETY     RCONST(0.96)
```
Steven-Roberts, line 52: With the fix to the controller exponent, I'd suggest setting this to 0.9 which is a pretty conventional value. Then the bias should be set to 1 along with that.

src/sunheuristics/default/sunheuristics_default.c
```diff
@@ -0,0 +1,488 @@
...
+#define DEFAULT_GROWTH     RCONST(20.0)
+#define DEFAULT_HFIXED_LB  RCONST(1.0)
+#define DEFAULT_HFIXED_UB  RCONST(1.5)
```
Steven-Roberts, line 55: I suggest setting this to 1 to disable deadzones by default.

src/suncontrol/i/suncontrol_i.c
```diff
@@ -0,0 +1,192 @@
...
+#define SC_I_BIAS(C)        ( SC_I_CONTENT(C)->bias )
+#define SC_I_P(C)           ( SC_I_CONTENT(C)->p )
+#define SC_I_PQ(C)          ( SC_I_CONTENT(C)->pq )
+
+/* ------------------
+ * Default parameters
+ * ------------------ */
+
+#define DEFAULT_K1     RCONST(1.0)
+#define DEFAULT_BIAS   RCONST(1.5)
```
Steven-Roberts, line 38: I suggest making this 1 along with a default safety factor of 0.9 in the heuristics.

src/suncontrol/i/suncontrol_i.c
```diff
@@ -0,0 +1,192 @@
...
+ * implementation of controller operations
+ * ----------------------------------------------------------------- */
+
+SUNControl_Type SUNControlGetType_I(SUNControl C) { return SUNDIALS_CONTROL_H; }
+
+int SUNControlEstimateStep_I(SUNControl C, realtype h,
+                               realtype dsm, realtype* hnew)
+{
+  /* set usable time-step adaptivity parameters */
+  const realtype k1 = -SC_I_K1(C) / SC_I_P(C);
```
Steven-Roberts, line 117: If we edit the controller struct to include p and q, the denominator should be `min(p,q)+1`

Steven-Roberts (commented):
For controller-related comments, I only considered the I controller, but they apply to all controllers.

doc/shared/suncontrol/SUNControl_Description.rst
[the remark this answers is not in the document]
drreynolds: I'll go with `SUNAdaptController` and `SUNTimestepHeuristics`

doc/shared/suncontrol/SUNControl_ExpGus.rst
```diff
@@ -0,0 +1,88 @@
...
+* ``ep`` - storage for the previous error estimate, :math:`\varepsilon_{n-1}`.
+
+* ``p`` - asymptotic order to use in error control.
+
+* ``pq`` - flag indicating whether ``p`` corresponds to the order of accuracy
+  for the time integration method (``SUNTRUE``) or the embedding (``SUNFALSE``).
+
+
+The header file to be included when using this module is
+``suncontrol/suncontrol_expgus.h``.
```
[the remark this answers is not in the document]
drreynolds: But that isn't how we "package" other modules throughout SUNDIALS.  That said, I'm fine with including these inside `arkode_arkstep.h` and `arkode_erkstep.h` (i.e., packaging single-rate controllers with single-rate steppers).

doc/shared/suncontrol/SUNControl_Description.rst
```diff
@@ -0,0 +1,480 @@
...
+   ========================  =====================================================
+   SUNDIALS_CONTROL_NONE     Empty object that performs no control.
+   SUNDIALS_CONTROL_H        Controls a single-rate step size.
+   SUNDIALS_CONTROL_HQ       Controls a single-rate step size and method order.
+   SUNDIALS_CONTROL_MRI_H    Controls two multirate step sizes.
+   SUNDIALS_CONTROL_MRI_TOL  Controls slow multirate step size and fast tolerance.
+   ========================  =====================================================
+
+
+.. _SUNControl.Description.operations:
```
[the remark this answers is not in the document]
drreynolds: I think it would be cleaner to:
* if a file and/or section is new, just add it to the very start of that file/section.
* if a function is being added to an existing file, then only add it to that function.

doc/shared/sunheuristics/SUNHeuristics_Description.rst
```diff
@@ -0,0 +1,647 @@
...
+   ========================  =====================================================
+   SUNDIALS_HEURISTICS_STD   Performs "standard" heuristic stepsize constraints.
+   SUNDIALS_HEURISTICS_NULL  Performs no heuristic constraints.
+   ========================  =====================================================
+
+
+Additionally, if the heuristics object can be used to limit stepsizes based on
+explicit stability, then we define the following user-supplied function type:
+
+.. c:type:: int (*SUNExpStabFn)(realtype *hstab, void *user_data)
```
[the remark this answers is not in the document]
drreynolds: Stability is orthogonal to accuracy, and so I do not consider it to be part of the error controller, but instead an heuristic constraint on the allowable time step.

Furthermore, as currently organized, it is perfectly feasible to combine both this stability restriction to the heuristics, while also using temporal error control.  Then, a problem can transition between accuracy-limited and stability-limited steps seamlessly.  We cannot, however, currently combine two error controllers within the same calculation, so if the explicit stability were a separate `SUNAdaptController`, it could NOT be combined with a true temporal error controller.

doc/shared/sunheuristics/SUNHeuristics_Description.rst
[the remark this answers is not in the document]
drreynolds: I'll defer to Cody on this, since I originally had no intention of creating this "heuristics" class, and only added it based on his request.  However, at this point it would be incredibly messy to smash all of these heuristics into each adaptivity controller.  If anything, I would just put it back in the main ARKODE memory structure (where it was before this PR).

doc/superbuild/source/suncontrol/index.rst
```diff
@@ -0,0 +1,20 @@
...
+   and Southern Methodist University.
+   All rights reserved.
+
+   See the top-level LICENSE and NOTICE files for details.
+
+   SPDX-License-Identifier: BSD-3-Clause
+   SUNDIALS Copyright End
+   ----------------------------------------------------------------
+
+.. include:: ../../../shared/suncontrol/SUNControl_Description.rst
```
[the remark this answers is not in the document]
drreynolds: I don't quite follow -- there is indeed a `SUNAdaptController_links.rst` file in this folder (mirroring the structure of all other subdirectories inside `superbuild/source`).  Perhaps if you just make the change that you're recommending it would be faster all the way around?

doc/superbuild/source/sunheuristics/index.rst
```diff
@@ -0,0 +1,20 @@
...
+   and Southern Methodist University.
+   All rights reserved.
+
+   See the top-level LICENSE and NOTICE files for details.
+
+   SPDX-License-Identifier: BSD-3-Clause
+   SUNDIALS Copyright End
+   ----------------------------------------------------------------
+
+.. include:: ../../../shared/sunheuristics/SUNHeuristics_Description.rst
```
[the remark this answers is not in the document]
drreynolds: Again, I don't follow, since this structure replicates all other `superbuild/source` folders).

examples/arkode/CXX_parallel/ark_diffusion_reaction_p.cpp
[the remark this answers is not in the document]
drreynolds: Then let's add this "changing it later..." to our longer-term to-do list, since we certainly cannot change it yet.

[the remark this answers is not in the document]
drreynolds: Also, ARKODE destroys whatever controller and heuristics objects it has in-hand when it is destroyed, thus there is no need to manually destroy it in the examples (if you do, you get a double-free).  If you feel strongly that we should carry around a bool to say whether ARKODE "owns" those objects, and then only frees them if it owned them originally, let me know.

include/arkode/arkode_arkstep.h
drreynolds: It is being deprecated for three reasons: 
1. The new controller and heuristics objects are the natural places to modify these values.
2. It is now clear that the "defaults" my optimization method honed in on years ago were biased toward very stiff applications with direct algebraic solvers, and thus the name implies some meaningless guarantee.
3. I've never heard of a single user calling this function.

examples/arkode/CXX_parallel/ark_diffusion_reaction_p.cpp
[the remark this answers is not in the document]
drreynolds: I had converted all uses of `ARKStepSetAdaptivityMethod` to the new approach, but I've just reverted the change in `ark_heat1D_adapt.c` to serve as a unit test.  That said, these kind of unit tests should not reside in `examples` since we should strive for those to show "best practices" instead of deprecated functions.

include/sundials/sundials_control.h
drreynolds: The `NONE` type is used when ARKODE is run with a `NULL`-valued `SUNAdaptController` object.  This is what we `arkSetFixedStep` now does (removes the controller, and adjusts the min/max step sizes in the `SUNTimestepHeuristics` object).

include/sundials/sundials_control.h
drreynolds: These were all inserted to emulate what is done in `SUNMemory`.  I'll remove them from this PR, but perhaps the others should be removed as well.

src/arkode/arkode.c
```diff
@@ -1958,21 +1968,39 @@ int arkInitialSetup(ARKodeMem ark_mem, realtype tout)
...
     } else {
-      /* Use standard growth factor (e.g., for reset) */
-      ark_mem->hadapt_mem->etamax = ark_mem->hadapt_mem->growth;
+      /* Notify time step heuristic controller object of previous successful steps */
+      retval = SUNHeuristicsUpdate(ark_mem->hconstraints);
+      if (retval != 0) {
+        arkProcessError(ark_mem, ARK_HEURISTICS_ERR, "ARKODE", "arkInitialSetup",
+                        "Failure updating heuristics object");
+        return(ARK_HEURISTICS_ERR);
+      }
```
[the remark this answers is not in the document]
drreynolds: This is handled by a call to `SUNTimestepHeuristics_Update`, which only occurs after a successful time step.  On construction, it uses `etamx1`, but after this call it changes to `etamax`.

src/arkode/arkode.c
```diff
@@ -3041,41 +3079,21 @@ int arkCheckTemporalError(ARKodeMem ark_mem, int *nflagPtr, int *nefPtr, realtyp
...
+  /* If step reduction successful, store result and return to reattempt
+     the step.  Otherwise, return ARK_ERR_FAILURE or ARK_HEURISTICS_ERR,
+     depending on the circumstances */
+  if (retval == SUNHEURISTICS_SUCCESS) {
+    ark_mem->eta = hnew / ark_mem->h;
+    return(TRY_AGAIN);
+  } else if (retval == SUNHEURISTICS_CANNOT_DECREASE) {
+    return(ARK_ERR_FAILURE);
+  } else {
+    return(ARK_HEURISTICS_ERR);
```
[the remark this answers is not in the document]
drreynolds: Again, I'll defer to Cody on the need for a new heuristics class.

src/arkode/arkode_arkstep_io.c
drreynolds: These are wiped out whenever the heuristics object is `Reset` (this is called automatically upon heuristics creation).
`arkInit` will only call `SUNTimestepHeuristics_Reset` when it is called with `init_type == FIRST_INIT`.

src/arkode/arkode_erkstep_io.c
drreynolds: Those have been added back in -- I either deleted them accidentally or it occurred when merging changes from upstream.

src/arkode/arkode_erkstep_io.c
drreynolds: This is because in previous releases, the ARKODE "defaults" matched those in ARKStep, whereas ERKStep had a slightly different set of default parameters.  Thus in this PR, only ERKStep needs to clear out the controller/heuristics objects that were created/initialized by ARKODE, and then set these up directly.

src/arkode/arkode_io.c
drreynolds: This call shouldn't hurt anything, since (a) if the controller did not exist then `Destroy` will immediately return with success, and (b) if some controller did already exist and a user called this with `hfixed == 0` then this should indicate a return to the ARKODE "default" PID controller.  

I suppose that this could cause unexpected results in the [very odd] corner case where a user sets up a custom controller, attaches it to ARKODE, and then calls `SetFixedStep` with `hfixed==0`.  Do you actually think that we need to support this kind of usage?

src/arkode/arkode.c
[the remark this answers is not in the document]
drreynolds: I've updated the PR to include ownership flags.

Regarding MRI -- yes, upon initialization MRIStep should immediately destroy the default controller and replace it with its own, but that branch is still in progress.

src/arkode/arkode_user_controller.c
drreynolds: Yes (the `ttmp` part, at least).  Nice catch!

src/suncontrol/i/suncontrol_i.c
drreynolds: I'm good with most of these suggestions, but I have one follow-on question.  Suppose a user knows that their problem will exhibit severe order reduction but they want to use a higher-order DIRK method anyways, and in so doing they want to modify the controller to use a smaller exponent to account for the reduced method/embedding order.  

I see no way to do this (or equivalently, to allow current ARKODE users to get the same exponent they have now), based on these suggestions.  Even if a user manually called your proposed `SUNAdaptController_SetOrder` routine to set their value, it would be overridden when `arkStep_Init` calls that same routine.  

What about if the controller had a corresponding "GetOrder" routine, under the restriction that controllers with non-NULL `SetOrder` routine should initialize their order to 0.  Then if a user never called `SetOrder` directly, ARKStep could call `GetOrder` to see if this was nonzero before itself callin `SetOrder`?

Steven-Roberts: In that case, the user could create a tableau with `ARKodeButcherTable_Create` and set p and q to values that account for order reduction. Unfortunately, I don't see a straightforward way to load something from our catalog and modify the order since we don't have getters or setters for table properties. This is a much simpler solution but more verbose and manual.

Your solution also seems reasonable, but could have strange side effects if the same controller is used for two integrations with methods of different order.

drreynolds: Historically, whenever we have changed any defaults, we always strive for a "simple" way for users to recover what they had before (case in point, consider Carol's current collaborators who get significantly fewer steps with the current off-by-one exponents).  

While I like your idea of manipulating the Butcher table structure, I don't see this as "simple" for users.  I'll give some more thought to this question today (I teach and have office hours all afternoon), and come up with something tonight/tomorrow.

drreynolds: I think that the simplest solution is to add an `AdjustOrder` routine to our provided controllers, where the integer argument is then applied to whatever order it is subsequently handed by the integrator.  The integer order adjustment can then always be undone by a second call with a `0` input.

I'll work on implementing Steven's recommendations, plus the `AdjustOrder` routine, tomorrow.

Steven-Roberts: The `AdjustOrder` solution sounds good

doc/shared/suncontrol/SUNControl_Description.rst
```diff
@@ -0,0 +1,480 @@
...
+
+   .. code-block:: c
+
+      retval = SUNControlSpace(C, &lenrw, &leniw);
+
+
+
+.. _SUNControl.Description.errorCodes:
+
+SUNControl Error Codes
```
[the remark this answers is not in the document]
drreynolds: I'll note that the `SUNAdaptController` and `SUNTimestepHeuristics` modules live outside of ARKODE, and are instead at the higher SUNDIALS-wide level (where we do not yet have shared error codes).  Are you recommending that instead of having error codes for these new modules, they instead use 0, >0, <0 to indicate success, recoverable failure, and nonrecoverable failure?

src/arkode/arkode.c
drreynolds: I don't think so, since this calculation inherently relies on N_Vectors that encode the degree of constraint violation.  Perhaps the hard-coded value of `0.9` could be parameterized, but I think that the rest of this calculation should reside here inside `arkCheckConstraints`.

@gardner48, do you have any thoughts on this?

doc/shared/suncontrol/SUNControl_Description.rst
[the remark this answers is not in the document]
balos1: These should have their own error codes since they live outside of ARKODE.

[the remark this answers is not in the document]
gardner48: I meant that `SUNAdaptController` and `SUNTimestepHeuristics` could return `>0`, `0`, or `<0` and ARKODE would internally map that to its own values similar to what we do for callbacks.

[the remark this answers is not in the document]
balos1: I think error codes are certainly better than just using `>0`, `0` or `<0`, although we probably have some other places where we should have error codes but are just doing `>0`, `0` or `<0`.

include/sunheuristics/sunheuristics_default.h
drreynolds: I agree that this separation seems artificial at first.  I initially had both inside the adaptivity controller, but moved this safety factor here to sit alongside the corresponding CFL stability safety factor.  The main idea was that the controller specifies the step it recommends to meet the error threshold, but any post-processing of that recommendation (e.g., via `safety`) constitutes an heuristic change.

Furthermore, although both `bias` and `safety` play similar roles, `bias` multiplies the local temporal error estimate directly, whereas `safety` sits outside the `1/(p+1)` exponent.  Thus, I consider the `bias` as encoding a user-provided estimate of how a given problem will relate local to global error (or for that matter, how much more tightly the user wants local error to be controlled in the hopes of achieving a given global error).  

Hence the separation of `bias` and `safety` in this PR.

drreynolds:
I *believe* that I've addressed all PR comments (modulo some answers I provided to questions that may require further discussion, and modulo changing any "default" behavior as Steven requested).  The automated tests are currently failing because of differences in runtime statistics (some improved, others not).  While I work through those test result differences, I would appreciate it if @gardner48, @balos1, and @Steven-Roberts could look back through this PR to see if their previous requests/questions have been satisfied, and to see if my new changes pass their approval.

doc/arkode/guide/source/Usage/ERKStep_c_interface/User_callable.rst
```diff
@@ -1159,33 +1171,72 @@ the code, is provided in :numref:`ARKODE.Mathematics.Adaptivity`.
...
+   +-----------------------------------------------------------+------------------------------------------+-----------+
+
+
+
+.. c:function:: int ERKStepSetAdaptController(void* arkode_mem, SUNAdaptController C)
+
+   Sets a user-supplied time-step controller object.
+
+   **Arguments:**
+      * *arkode_mem* -- pointer to the ERKStep memory block.
```
balos1, line 1214: Sphinx has a `param` and `returns` directive that makes structuring arguments and return values much simpler and also render nicer. 

Here is an example
```
.. c:function:: void* SPRKStepCreate(ARKRhsFn f1, ARKRhsFn f2, realtype t0,\
                                     N_Vector y0, SUNContext sunctx)

   This function allocates and initializes memory for a problem to
   be solved using the SPRKStep time-stepping module in ARKODE.

   :param f1: the name of the C function (of type :c:func:`ARKRhsFn()`) defining :math:`f_1(t,q) = \frac{\partial V(t,q)}{\partial q}`
   :param f2: the name of the C function (of type :c:func:`ARKRhsFn()`) defining :math:`f_2(t,p) = \frac{\partial T(t,p)}{\partial p}`
   :param t0: the initial value of :math:`t`
   :param y0: the initial condition vector :math:`y(t_0)`
   :param sunctx: the :c:type:`SUNContext` object (see :numref:`SUNDIALS.SUNContext`)

   :returns: If successful, a pointer to initialized problem memory of type
             ``void*``, to be passed to all user-facing SPRKStep routines listed
             below.  If unsuccessful, a ``NULL`` pointer will be returned, and
             an error message will be printed to ``stderr``.
```

I am not suggesting that you change the documentation to this style now, just noting it for future reference. Eventually, I would like to change all the documentation to use these directives.

doc/shared/sunadaptcontroller/SUNAdaptController_Description.rst
```diff
@@ -0,0 +1,502 @@
...
+should note its type, so that integrators will understand the types of
+adaptivity that the controller is designed to perform. These are encoded in the
+following set of SUNAdaptController types:
+
+.. c:enum:: SUNAdaptController_Type
+
+   The enumerated type :c:type:`SUNAdaptController_Type` defines the enumeration
+   constants for SUNDIALS error controller types
+
+.. c:enumerator:: SUN_ADAPTCONTROLLER_NONE
```
balos1, line 96: Is this type necessary? Can't we implement a `SUN_ADAPTCONTROLLER_H` that does not change the value of `h`, i.e., an identity controller.

doc/arkode/guide/source/Usage/ERKStep_c_interface/User_callable.rst
balos1: I see you used these directives elsewhere, so you probably were just matching the existing style of ERKStep. Disregard my comment.

include/sunadaptcontroller/sunadaptcontroller_expgus.h
balos1, line 1: Did you happen to apply clang-format to all of the new source files? If not, we should going forward, but I will update these later.

include/sundials/sundials_timestepheuristics.h
```diff
@@ -0,0 +1,286 @@
...
+/* -----------------------------------------------------------------
+ * SUNTimestepHeuristics types (currently, only "H" is implemented; others
+ * are planned):
+ *    STD     - standard heuristic stepsize constraints
+ *    NULL    - placeholder object that applies no constraints
+ * ----------------------------------------------------------------- */
+
+typedef enum
+{
+  SUN_TIMESTEPHEURISTICS_STD,
```
balos1, line 41: Do we prefer to call them standard, or default? I think this should match how we talk about them in the docs.

balos1: Looks like the implementation is called "default", so we should change this enum name to "default".

doc/arkode/guide/source/Usage/ERKStep_c_interface/User_callable.rst
drreynolds: Yep, the new documentation files as part of this PR use this new structure, but for existing files I respected the style of the surrounding documentation.  I agree that a future PR could standardize this across SUNDIALS.

src/sundials/sundials_adaptcontroller.c
```diff
@@ -0,0 +1,261 @@
...
+}
+
+
+/* -----------------------------------------------------------------
+ * Required functions in the 'ops' structure for non-NULL controller
+ * ----------------------------------------------------------------- */
+
+SUNAdaptController_Type SUNAdaptController_GetType(SUNAdaptController C)
+{
+  if (C == NULL) { return SUN_ADAPTCONTROLLER_NONE; }
```
balos1, line 94: I think we need to be very careful about purposefully making `NULL` to be a valid object, which is what we seem to be doing here.  This could lead to some hard to debug issues I imagine. I think the better approach would be to implement a controller that does not change the step size (as I recommended in a different comment).

src/sundials/sundials_timestepheuristics.c
```diff
@@ -0,0 +1,376 @@
...
+}
+
+
+/* -----------------------------------------------------------------
+ * Required functions in the 'ops' structure
+ * ----------------------------------------------------------------- */
+
+SUNTimestepHeuristics_ID SUNTimestepHeuristics_GetID(SUNTimestepHeuristics H)
+{
+  if (H == NULL) { return SUN_TIMESTEPHEURISTICS_NULL; }
```
balos1, line 88: Same comment as on the controller.  I dont think we want to make `NULL` a valid "object."

balos1 (changes requested):
This is looking really good. I have one request which is that we do not use `NULL` as a valid option for the controller and heuristic objects. Instead, I suggest we implement versions that just do not do anything and call them "noop" or "identity" variants.

doc/shared/suncontrol/SUNControl_Description.rst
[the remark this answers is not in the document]
gardner48: I think it's good for our implementations to use a consistent set of error codes but I'm not sure it's necessary for a user-supplied controller have to use these codes. The integratos are not worried about the specific error value just that it was `>0`, `0`, or `<0` e.g., for any `<0` return ARKODE could return `ARK_CONTROLLER_ERROR` and the user could query the controller for it's last error if necessary (like `GetLastFlag` with the linear solvers).

doc/shared/sunheuristics/SUNHeuristics_Description.rst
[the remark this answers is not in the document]
gardner48: I guess I'm thinking about controller objects more generally as any method for adjusting the step size/order instead of only an error-based controller. So a CFL controller would just be another "type" of controller (e.g., `SUN_Controller_CFL`) that gets attached with a corresponding set function (e.g., `ARKStepSetCFLController`).

This could similarly be applied to the heuristic controller to break it up into smaller controllers with a more specific purpose (something like `SUN_Controller_H_Bounds`, `SUN_Controller_H_Growth`, etc.) but this would require several implementations and I'm not sure that's necessary in order to turn off the heuristics that are problematic for enabling adaptivity in SPRK step.

doc/shared/sunheuristics/SUNHeuristics_Description.rst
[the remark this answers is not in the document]
gardner48: I think moving [these lines](https://github.com/LLNL/sundials/blob/a4553b5021b7089b036f1edcfdd9457a7764d5a0/src/arkode/arkode.c#L890C1-L897C1) inside `arkCheckTemporalError` (which happened when adding the heuristic class) addresses what's needed for adaptivity in SPRKStep without having to add the heuristic class. 

With the above change all the heuristic limits that could apply to SPRKStep are now inside `arkCheckTemporalError` (the only ones outside are related to solver fails, constraints, relaxation, and if the first step size is estimated internally). So if SPRKStep sets the step size change factor (`eta`) and `arkCheckTemporalError` is not called (by adding a flag or check if an error-based controller is not present), then none of the heuristic limits are applied (as far as I can tell). 

The idea of having the option to supply alternative approaches to the current heuristics would be interesting. Specifically, I'm wondering if we can some how use information from a failed solve to get a better reduction factor. However, I think that could be done (in the future) with a new controller type, see [this comment above](https://github.com/LLNL/sundials/pull/289#discussion_r1324019842).

doc/superbuild/source/suncontrol/index.rst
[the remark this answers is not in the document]
gardner48: I've moved this line inside `SUNAdaptController_links.rst`. Other subdirectories have a brief introduction to the chapter which appears before the first section by adding the `include` before the `toctree`. In this case the material that would normally be the first section was appearing as the intro to the chapter.

doc/superbuild/source/sunheuristics/index.rst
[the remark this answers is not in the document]
gardner48: Moved this line inside `SUNTimestepHeuristics_links.rst` ([see comment above](https://github.com/LLNL/sundials/pull/289#discussion_r1324047334))

src/arkode/arkode.c
gardner48: I think we should parameterize the `0.9`. We could create a controller type (see [this comment](https://github.com/LLNL/sundials/pull/289#discussion_r1359967838)) for determining the new step size with constraints, possibly doing something different from the current linear interpolant, but that could be done in a later PR.

doc/shared/suncontrol/SUNControl_Description.rst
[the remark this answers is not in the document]
balos1: We have defined error codes for all of our core classes though, so I think it is most consistent to have these error codes defined.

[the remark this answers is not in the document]
gardner48: I don't disagree that we should have defined error codes for these classes. I think we should clarify that since the integrators do not rely on the specific return value user-defined implementations of these classes are not required to use these error codes.

include/sunadaptcontroller/sunadaptcontroller_expgus.h
drreynolds: I did not run clang-format.

drreynolds:
Closing this PR, and replacing with #358

PR #294, opened 2023-06-27
Fix SUNDIALS_SUPERLUMT_THREAD_TYPE definition

jschueller:
Assimulo looks for this define
unfortunately the syntax was wrong


jschueller:
done

PR #314, opened 2023-08-02
Fix memory leak in SUNMemoryHelpers

gardner48:


gardner48:
@balos1 Looks like there is a permissions issue in GitLab. When I rerun a test, it then fails because it can't find the Kokkos install directory
```
CMake Error at cmake/tpl/SundialsKokkos.cmake:43 (find_package):
  Could not find a package configuration file provided by "Kokkos" with any
  of the following names:
    KokkosConfig.cmake
    kokkos-config.cmake
  Add the installation prefix of "Kokkos" to CMAKE_PREFIX_PATH or set
  "Kokkos_DIR" to a directory containing one of the above files.  If "Kokkos"
  provides a separate development package or SDK, be sure it has been
  installed.
```

balos1 (approved):

Issue #312, opened 2023-07-31
Attempting to use an MPI routine (internal_Wtime) before initializing or after finalizing MPICH


sagitter:
Hi all.

This errors break many `Sundials-6.6.0`'s tests in `Fedora 39 i686`with `MPICH-4.1.2` + `GCC-13.1.1`:

```
83: Test command: /builddir/build/BUILD/sundials-6.6.0/buildmpich_dir/build/examples/sunnonlinsol/fixedpoint/test_sunnonlinsol_fixedpoint "2" "0.5"
/builddir/build/BUILD/cmake-3.27.1/Source/CTest/cmCTestRunTest.cxx:695 83: Working Directory: /builddir/build/BUILD/sundials-6.6.0/buildmpich_dir/build/examples/sunnonlinsol/fixedpoint
/builddir/build/BUILD/cmake-3.27.1/Source/cmCTest.cxx:309    Current_Time: Jul 30 20:16 UTC
/builddir/build/BUILD/cmake-3.27.1/Source/CTest/cmCTestRunTest.cxx:806 83: Test timeout computed to be: 1500
/builddir/build/BUILD/cmake-3.27.1/Source/CTest/cmCTestRunTest.cxx:39 83: Attempting to use an MPI routine (internal_Wtime) before initializing or after finalizing MPICH
/builddir/build/BUILD/cmake-3.27.1/Source/CTest/cmCTestRunTest.cxx:39 83: Solve the nonlinear system:
/builddir/build/BUILD/cmake-3.27.1/Source/CTest/cmCTestRunTest.cxx:39 83:     3x - cos((y-1)z) - 1/2 = 0
/builddir/build/BUILD/cmake-3.27.1/Source/CTest/cmCTestRunTest.cxx:39 83:     x^2 - 81(y-0.9)^2 + sin(z) + 1.06 = 0
/builddir/build/BUILD/cmake-3.27.1/Source/CTest/cmCTestRunTest.cxx:39 83:     exp(-x(y-1)) + 20z + (10 pi - 3)/3 = 0
/builddir/build/BUILD/cmake-3.27.1/Source/CTest/cmCTestRunTest.cxx:39 83: Analytic solution:
/builddir/build/BUILD/cmake-3.27.1/Source/CTest/cmCTestRunTest.cxx:39 83:     x = 0.5
/builddir/build/BUILD/cmake-3.27.1/Source/CTest/cmCTestRunTest.cxx:39 83:     y = 1
[...]

Issue #139, opened 2022-05-11
unit_tests which use internal functions will not link successfully on Windows


balos1:
I have not confirmed, but I believe this is because internal functions are not marked for export.

```
[100%] Linking C executable ark_test_arkstepsetforcing.exe
cd /d C:\Users\cjbal\Documents\Workspace\sundials\sundials\buildir-mingw\test\unit_tests\arkode\C_serial && "C:\Program Files\CMake\bin\cmake.exe" -E cmake_link_script CMakeFiles\ark_test_arkstepsetforcing.dir\link.txt --verbose=1
"C:\Program Files\CMake\bin\cmake.exe" -E remove -f CMakeFiles\ark_test_arkstepsetforcing.dir/objects.a
C:\MinGW\bin\ar.exe cr CMakeFiles\ark_test_arkstepsetforcing.dir/objects.a @CMakeFiles\ark_test_arkstepsetforcing.dir\objects1.rsp
C:\MinGW\bin\gcc.exe    -Wl,--whole-archive CMakeFiles\ark_test_arkstepsetforcing.dir/objects.a -Wl,--no-whole-archive  -o ark_test_arkstepsetforcing.exe -Wl,--out-implib,libark_test_arkstepsetforcing.dll.a -Wl,--major-image-version,0,--minor-image-version,0 @CMakeFiles\ark_test_arkstepsetforcing.dir\linklibs.rsp
c:/mingw/bin/../lib/gcc/mingw32/9.2.0/../../../../mingw32/bin/ld.exe: CMakeFiles\ark_test_arkstepsetforcing.dir/objects.a(ark_test_arkstepsetforcing.c.obj):ark_test_arkstepsetforcing.c:(.text+0xdfe): undefined reference to `arkStep_SetInnerForcing'
c:/mingw/bin/../lib/gcc/mingw32/9.2.0/../../../../mingw32/bin/ld.exe: CMakeFiles\ark_test_arkstepsetforcing.dir/objects.a(ark_test_arkstepsetforcing.c.obj):ark_test_arkstepsetforcing.c:(.text+0x1267): undefined reference to `arkStep_SetInnerForcing'
[...]

PR #317, opened 2023-08-10
Bugfix/mpi wtime profiler

balos1:
Todo:

- [x] Check profiling unit test output on windows


balos1:
Fixes #312.

src/sundials/sundials_profiler.c
```diff
@@ -496,22 +514,44 @@ int sunCompareTimes(const void* l, const void* r)
...
+    if (!ticks_per_sec.QuadPart)
+    {
+      return -1;
+    }
+  }
+
+  QueryPerformanceCounter(&ticks);
+
+  ts->tv_sec  = (long)(ticks.QuadPart / ticks_per_sec.QuadPart);
+  ts->tv_nsec = (long)(((ticks.QuadPart % ticks_per_sec.QuadPart) * 1000000000) /
```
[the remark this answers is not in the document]
balos1: The Windows documentation says "1 microsecond or better" but I have not found a way to get the exact resolution.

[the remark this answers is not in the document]
balos1: I think this is still right, since `ticks.QuadPart % ticks_per_sec.QuadPart` would give a remainder in seconds, multiplying it by 1,000,000,000 then yields the remainder in nanoseconds.

balos1:
I need to add a unit test which does a sleep or something to verify the timer is giving the value we would expect.

gardner48:
> I need to add a unit test which does a sleep or something to verify the timer is giving the value we would expect.

Take a look at the `profiler-csv-output` branch I've added a test like that there

drreynolds (dismissed):

src/sundials/sundials_profiler.c
```diff
@@ -496,22 +548,43 @@ int sunCompareTimes(const void* l, const void* r)
...
+  /* QueryPerformanceCounter is ticks in microseconds */
+
+  ts->tv_sec  = (long)(ticks.QuadPart / ticks_per_sec.QuadPart);
+  ts->tv_nsec = (long)(((ticks.QuadPart % ticks_per_sec.QuadPart) * 1000000) /
+                       ticks_per_sec.QuadPart);
+
+  return 0;
+#else
+  return clock_gettime(CLOCK_MONOTONIC, ts);
+#endif
```
gardner48, line 589: Switch the order to simplify the `#if` to `#if defined(SUNDIALS_HAVE_POSIX_TIMERS)`

test/unit_tests/profiling/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
+# ------------------------------------------------------------------------------
+# Programmer(s): David J. Gardner @ LLNL
+# ------------------------------------------------------------------------------
+# SUNDIALS Copyright Start
+# Copyright (c) 2002-2022, Lawrence Livermore National Security
```
gardner48, line 5:
```suggestion
# Copyright (c) 2002-2023, Lawrence Livermore National Security
```

test/unit_tests/profiling/test_profiling.cpp
```diff
@@ -0,0 +1,243 @@
+/* -----------------------------------------------------------------------------
+ * Programmer(s): David J. Gardner @ LLNL
+ * -----------------------------------------------------------------------------
+ * SUNDIALS Copyright Start
+ * Copyright (c) 2002-2022, Lawrence Livermore National Security
```
gardner48, line 5:
```suggestion
 * Copyright (c) 2002-2023, Lawrence Livermore National Security
```

test/unit_tests/profiling/test_profiling.cpp
```diff
@@ -0,0 +1,243 @@
...
+#include <chrono>
+#include <cstdio>
+#include <iostream>
+#include <ostream>
+#include <string>
+#include <thread>
+
+#include "sundials/sundials_math.h"
+#include "sundials/sundials_profiler.h"
+#include "sundials/sundials_types.h"
```
gardner48, line 24:
```suggestion
```

test/unit_tests/profiling/test_profiling.cpp
```diff
@@ -0,0 +1,243 @@
...
+    return 1;
+  }
+
+  flag = SUNProfiler_Print(prof, stdout);
+  if (flag)
+  {
+    std::cerr << ">>> FAILURE: "
+              << "SUNProfiler_Print returned " << flag << "\n";
+    return 1;
+  }
```
gardner48, line 58: Remove duplicate blocks until we get around to adding a `SetFormat` option for output
```suggestion
```

cmake/SundialsSetupCompilers.cmake
```diff
@@ -223,13 +223,6 @@ if(SUNDIALS_POSIX_TIMERS AND POSIX_TIMERS_NEED_POSIX_C_SOURCE)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -D_POSIX_C_SOURCE=${SUNDIALS_POSIX_C_SOURCE}")
 endif()
 
-# Check if profiling is being built with no timers.
-if(SUNDIALS_BUILD_WITH_PROFILING AND
-   (NOT ENABLE_CALIPER) AND
-   (NOT ENABLE_MPI) AND
-   (NOT SUNDIALS_POSIX_TIMERS))
-  message(SEND_ERROR "The SUNDIALS native profiler requires POSIX timers or MPI_Wtime, but neither were found.")
-endif()
```
gardner48, line 232: Do we always build the internal profiler, even when Caliper is enabled? If so, we should double check that the documentation notes POSIX or windows timers are required when profiling is enabled and remove potential the MPI requirement if that was noted.

gardner48 (changes requested):

cmake/SundialsSetupCompilers.cmake
balos1: Yes, we include the profiler even when caliper is enabled because otherwise user code would need a lot of ifdefs. I'll add a note to the docs.

src/sundials/sundials_profiler.c
balos1: I am not sure I follow. We would still have to check if it's WIN32, so it would only rearrange the logic rather than simplify.

src/sundials/sundials_profiler.c
```diff
@@ -15,36 +15,51 @@
...
-#define SUNDIALS_ROOT_TIMER ((const char*) "From profiler epoch")
+#if (defined(WIN32) || defined(_WIN32)) && !defined(SUNDIALS_HAVE_POSIX_TIMERS)
+typedef struct _sunTimespec
+{
+  long int tv_sec;
+  long int tv_nsec;
+} sunTimespec;
+#else
+typedef struct timespec sunTimespec;
+#endif
```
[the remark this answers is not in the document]
balos1: This makes the assumption that not having posix implies its windows. While that may be true in most cases, it's not strictly true. So I think we should leave this as is with the check for WIN32.

src/sundials/sundials_profiler.c
```diff
@@ -53,96 +68,78 @@ static int sunCompareTimes(const void* l, const void* r);
...
+  long ns_difference = 0;
+
+  sunclock_gettime_monotonic(entry->toc);
+
+  s_difference  = entry->toc->tv_sec - entry->tic->tv_sec;
+  ns_difference = entry->toc->tv_nsec - entry->tic->tv_nsec;
+  if (ns_difference < 0.0)
+  {
+    s_difference--;
+    ns_difference = 1000000000 + entry->toc->tv_nsec - entry->tic->tv_nsec;
```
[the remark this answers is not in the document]
balos1: This change would make it a floating point operation though, which was not the intention.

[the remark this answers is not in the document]
balos1: I see the result is a double, so maybe it doesn't matter.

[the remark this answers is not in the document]
gardner48: > This change would make it a floating point operation though, which was not the intention.

Good point, this should stay as is.

src/sundials/sundials_profiler.c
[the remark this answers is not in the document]
gardner48: I think these are the same, i.e., use the `typedef` if POSIX timers are present else define `sunTimespec`. Are you suggesting this should be something like the `#if` at the top of the file i.e., 
```
#if defined(SUNDIALS_HAVE_POSIX_TIMERS)
// use typedef
#elif defined(WIN32) || defined(_WIN32)
// define sunTimespec
#else
#error No supported timers found
#endif
```

src/sundials/sundials_profiler.c
```diff
@@ -254,57 +250,96 @@ int SUNProfiler_Begin(SUNProfiler p, const char* name)
...
+int SUNProfiler_GetTimerResolution(SUNProfiler p, double* resolution)
+{
+#if defined(SUNDIALS_HAVE_POSIX_TIMERS)
+  sunTimespec spec;
+  clock_getres(CLOCK_MONOTONIC, &spec);
+  *resolution = 1e-9 * ((double)spec.tv_nsec);
+
+  return (0);
+#elif (defined(WIN32) || defined(_WIN32)) && \
+  !defined(SUNDIALS_HAVE_POSIX_TIMERS)
```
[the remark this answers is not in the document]
gardner48: Checking for `SUNDIALS_HAVE_POSIX_TIMERS` is already covered by the initial `#if`

src/sundials/sundials_profiler.c
gardner48: It would simplify the `#if` from
```
#if (defined(WIN32) || defined(_WIN32)) && !defined(SUNDIALS_HAVE_POSIX_TIMERS)
...
#else
...
#endif
```
to
```
#if defined(SUNDIALS_HAVE_POSIX_TIMERS)
...
#else
...
#endif
```

gardner48:
The CI is failing with 
```
/var/lib/jenkins/workspace/SUNDIALS_GitHub_Test_PR-317/test/tarballs/sundials-6.6.1/include/sundials/sundials_profiler.h:122:3: error: C++ style comments are not allowed in ISO C90 [-Werror]
 } // namespace sundials
```
Maybe there's a misplaced `#ifdef __cplusplus`. For some of the other classes we've started putting the convenience wrappers for C++ into a `.hpp` file. We should probably do the same with here as well.

src/sundials/sundials_profiler.c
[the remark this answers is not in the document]
balos1: Yes, but its probably redundant with the check at the top so your suggestion should be fine.

doc/shared/sundials/Profiling.rst
```diff
@@ -48,6 +48,11 @@ If Caliper is enabled, then users should refer to the `Caliper documentation <ht
 for information on getting profiler output. In most cases, this involves
 setting the ``CALI_CONFIG`` environment variable.
 
+
+.. note:: 
+
+   The SUNDIALS profilers requires POSIX timers or the Windows ``profileapi.h`` timers.
```
gardner48, line 54:
```suggestion
   The SUNDIALS profiler requires POSIX timers or the Windows ``profileapi.h`` timers.
```

include/sundials/sundials_profiler.hpp
gardner48, line 1: Ci is failing because this file needs to be added to the install list

balos1:
I have addressed #139 in this too so that the unit tests can build on Windows.

test/unit_tests/arkode/CXX_serial/CMakeLists.txt
```diff
@@ -50,10 +50,17 @@ foreach(test_tuple ${unit_tests})
       ${CMAKE_SOURCE_DIR}/include
       ${CMAKE_SOURCE_DIR}/src)
 
-    # Libraries to link against
+    # We explicitly choose which object libraries to link to
+    # and link in the arkode objects so that we have access to
+    # private functions w/o changing their visibility in the installed libraries.
```
gardner48, line 55: Nice! Are similar changes need to other unit tests to resolve #139? I think the ARKODE tests might be the only ones that access internal functions currently.

balos1: It should only be needed for the tests that use internal functions.

balos1: It seems my fix may not work on all systems or all versions of CMake that we support though, so I may need to revert this and turn off the unit tests on Windows.

gardner48 (approved):

PR #321, opened 2023-08-19
Bugfix/arkode adaptivity

Steven-Roberts:
Fixes an off by one error in exponents used by ARKODE's error controllers. There are a couple things to consider before merging:

- Will the parameters set in `ARKStepSetOptimalParams` still be optimal?
- Are the current bias and safety factors too conservative now? For example, for a second order method, the suggested step size effectively get scaled by the relatively small value `safety_factor*(bias)^(-1/2) = 0.78`.


Steven-Roberts:
Closing to make changes in #289

PR #343, opened 2023-09-20
Fix soversion ?

jschueller:
some libs are missing soversion, and this causes clients libs to link against full version instead:
xref https://github.com/conda-forge/assimulo-feedstock/issues/88
this doesnt look intentional because matching soversion macros are actually defined

cc @gardner48 @jmrohwer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # SUNDIALS Changelog
 
+Fixed missing soversions.
+
 ## Changes to SUNDIALS in release 6.6.1
 
 Updated the Tpetra NVector interface to support Trilinos 14.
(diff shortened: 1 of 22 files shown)


balos1:
Yep, this seems to be a typo in the CMake. Thanks for opening the PR, its greatly appreciated!  Could you add a note to the CHANGELOG and sign the commit as per https://github.com/LLNL/sundials/blob/main/CONTRIBUTING.md?

jschueller:
done

balos1 (approved):

balos1 (approved):

jschueller:
@balos1 would it be possible to backport this patch into the next point release if any ?
Sadly it was not part of 6.6.2 version

balos1:
@jschueller We are planning to do so, yes. It was missed in 6.6.2 unfortunately.

PR #345, opened 2023-09-23
Feature/default order 4 ark

Steven-Roberts:
Changes the default ERK and DIRK integrators to improve robustness of method and embedding


drreynolds (changes requested):
The improvements from switching out these default integrators ranges from dramatic to marginal (most problems improve dramatically).  

My one question is: how much testing with various RK table options did you do before settling on these specific default tables?

Also, you'll need to edit the `CHANGELOG.md` file, as well as the "recent changes" section in ARKODE's `Introduction.rst` file.  In those, I recommend adding a sentence to tell users how to switch back to the previous default.

Steven-Roberts:
> how much testing with various RK table options did you do before settling on these specific default tables?

I specifically chose these method due to there optimized properties and so that the number of explicit stages or implicit solves would be the same as the old defaults. Thus, the per step cost will likely not change significantly, but the accuracy or embedding quality will be better. This constraint doesn't leave many options, and I only tested the methods in this PR with the old defaults. If you have candidate methods you think me be better defaults, I can run some comparisons.

drreynolds:
> > how much testing with various RK table options did you do before settling on these specific default tables?
> 
> I specifically chose these method due to there optimized properties and so that the number of explicit stages or implicit solves would be the same as the old defaults. Thus, the per step cost will likely not change significantly, but the accuracy or embedding quality will be better. This constraint doesn't leave many options, and I only tested the methods in this PR with the old defaults. If you have candidate methods you think might be better defaults, I can run some comparisons.

This is fine (at least for now).  I'm working on updates to my regression-testing repository for ARKODE, that would hopefully allow us to do more thorough testing of different methods.  That said, given the improvement here, I think that we can approve this PR, and if needed we can update the default method again if we find something better before the next release.

balos1:
This PR will need a corresponding update of https://github.com/sundials-codes/answers to update the output files used in the GitHub actions CI.

drreynolds (dismissed):

PR #346, opened 2023-09-29
bugfix-Support-ARK2-3-1-2

drreynolds:
Although we added a new low-order additive RK Butcher table pair, `ARKStepSetTableNum` did not recognize this as a valid ARK pair.  This fixes that issue.


gardner48 (changes requested):
Looks good, I think all that's needed is to update the change log and recent changes in the ARKODE docs to note the fix.

gardner48 (approved):

PR #349, opened 2023-10-03
Bugfix: stop time and output time reached in same step

### LINKED ISSUE #339: IDAStopTest2 bug
bfec386d9 moved check for tn past tstop before check for tn past tout in IDAStopTest2:
 https://github.com/LLNL/sundials/commit/bfec386d94218dd27e85ceb8798a9021a83690ef?diff=split#diff-a624e4aaf10771efab2a0c451839e9c4f24613e0f5ee296e82cf5f5bc4fcecebL2101

in a test i have a simple linear system dxdt = 1.0, and root function g(x) = 1.0 - x, doing fixed-step integration with dt = 0.1, and tstop = 1.5, at t = 0.7, instead of IDAStep computing the solution t = 0.8, like it does in version `6.6.0`, it now, in `6.6.1`, instead does an early return and reports stop!

gardner48:
Fixes #339: A regression was introduced by the stop time bug fix in v6.6.1 causing integrators to return at the stop time rather than the requested output time if the stop time was reached in the same step in which the output time was passed.

Also fixes a missing check in ARKODE to interpolate (or not) the solution at the stop time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ CVODE, CVODES, IDA, and IDAS would return at the stop time rather than the
 requested output time if the stop time was reached in the same step in which the
 output time was passed.
 
+Fixed a bug in ARKODE where `ARKStepSetInterpolateStopTime` would return an
+interpolated solution at the stop time in some cases when interpolation was
+disabled.
+
 Fixed a bug in `ARKStepSetTableNum` wherein it did not recognize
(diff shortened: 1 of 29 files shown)


drreynolds (approved):

drreynolds (approved):

drreynolds (approved):

balos1 (approved):

balos1:
great to have a unit test for this now!

PR #351, opened 2023-10-18
Add Ginkgo dpcpp into example

yhmtsai:
This PR adds the ginkgo dpcpp support in the examples: sunmatrix, sunlinsol, and cvode/cv_heat2D

I currently only know `queue->wait_and_throw()` to synchronize which requires queue unlike `cudaDeviceSynchronize` or `hipDeviceSynchronize`
In sunlinsol and cv_heat2D, some function signatures are also used in other files such that I can not pass an additional parameter.
I use global variable to store ginkgo executor and then get the queue when SYCL needs sync or submit the kernel in those functions.


examples/sunlinsol/ginkgo/test_sunlinsol_ginkgo.cpp
```diff
@@ -143,7 +148,45 @@ void fill_matrix(gko::matrix::Csr<sunrealtype, sunindextype>* matrix)
   unsigned num_blocks        = (mat_rows + threads_per_block - 1) / threads_per_block;
 
   fill_kernel<<<num_blocks, threads_per_block>>>(mat_rows, mat_cols, row_ptrs, col_idxs, mat_data);
-  HIP_OR_CUDA(hipDeviceSynchronize();, cudaDeviceSynchronize(););
+  HIP_OR_CUDA_OR_SYCL(hipDeviceSynchronize(), cudaDeviceSynchronize(), matrix->get_executor()->synchronize());
```
balos1, line 151: This line would only compile when HIP or CUDA is used, so I think you can revert this line back.

examples/sunlinsol/ginkgo/test_sunlinsol_ginkgo.cpp
```diff
@@ -171,12 +214,37 @@ void fill_matrix(gko::matrix::Dense<sunrealtype>* matrix)
...
   sunrealtype* mat_data = matrix->get_values();
 
-#if defined(USE_CUDA)
+#if defined(USE_CUDA) || defined(USE_HIP)
   unsigned threads_per_block = 256;
   unsigned num_blocks        = (mat_rows + threads_per_block - 1) / threads_per_block;
 
   fill_kernel<<<num_blocks, threads_per_block>>>(mat_rows, mat_cols, mat_data);
-  HIP_OR_CUDA(hipDeviceSynchronize();, cudaDeviceSynchronize(););
+  HIP_OR_CUDA(hipDeviceSynchronize(), cudaDeviceSynchronize(), matrix->get_executor()->synchronize());
```
balos1, line 222: Since this line would only be compiled for HIP or CUDA, I think you can revert it back.

examples/sunlinsol/ginkgo/test_sunlinsol_ginkgo.cpp
balos1, line 1: Apply clang-format if it was not done already (I can't quite tell).

examples/sunmatrix/ginkgo/test_sunmatrix_ginkgo.cpp
balos1, line 1: Apply clang-format to this file if it was not alread.

examples/cvode/ginkgo/cv_heat2D_ginkgo.cpp
```diff
@@ -334,8 +353,45 @@ int f(sunrealtype t, N_Vector u, N_Vector f, void* user_data)
 
   f_kernel<<<num_blocks, threads_per_block>>>(nx, ny, dx, dy, cx, cy, cc, bx, by, sin_t_cos_t, cos_sqr_t, uarray, farray);
 
-  HIP_OR_CUDA(hipDeviceSynchronize();, cudaDeviceSynchronize(););
+  HIP_OR_CUDA_OR_SYCL(hipDeviceSynchronize(), cudaDeviceSynchronize(), global_exec->synchronize());
```
balos1, line 356: I think you can revert this line since it would only compile for HIP or CUDA.

examples/cvode/ginkgo/cv_heat2D_ginkgo.cpp
```diff
@@ -334,8 +353,45 @@ int f(sunrealtype t, N_Vector u, N_Vector f, void* user_data)
...
+
+#elif defined(USE_DPCPP)
+  // Access device data arrays
+  sunrealtype* uarray = N_VGetDeviceArrayPointer(u);
+  if (check_ptr(uarray, "N_VGetDeviceArrayPointer")) return -1;
 
+  sunrealtype* farray = N_VGetDeviceArrayPointer(f);
+  if (check_ptr(farray, "N_VGetDeviceArrayPointer")) return -1;
+
+  std::dynamic_pointer_cast<const gko::DpcppExecutor>(global_exec)->get_queue()->submit([&](sycl::handler& cgh) {
```
balos1, line 366: Instead of using the global_exec, we could store a reference/pointer to `gko_exec` in the `UserData` class. The `UserData` object is already passed into this callback as `void* user_data`.

balos1 (changes requested):
Overall, this looks good. I will try it out.

examples/sunlinsol/ginkgo/test_sunlinsol_ginkgo.cpp
yhmtsai: I revert them back, but I need to provide `HIP_OR_CUDA` and `HIP_OR_CUDA_OR_SYCL`.
which one do you prefer?

examples/sunlinsol/ginkgo/test_sunlinsol_ginkgo.cpp
yhmtsai: Becuase clang-format does not give consistence between version for those features which is not controlled by options, I wonder whether the current formatting result is expected or not.
I currently use clang-format-17.

balos1: This looks fine, thanks.

examples/sunlinsol/ginkgo/test_sunlinsol_ginkgo.cpp
balos1: I think we should actually go with
```
 HIP_OR_CUDA_OR_SYCL(hipDeviceSynchronize(), cudaDeviceSynchronize(), );
```
Then we only need one of the macros, but it is also clear that nothing happens in the SYCL case in this block of code.

balos1:
@yhmtsai Please sign your next commit as per https://github.com/LLNL/sundials/blob/main/CONTRIBUTING.md.

balos1 (dismissed):

balos1 (dismissed):

balos1 (approved):

balos1 (approved):

drreynolds (approved):

PR #353, opened 2023-10-20
Fortran interface for lapack dense solver

balos1:


drreynolds (approved):

gardner48 (changes requested):
I think the tarscripts need to be updated, otherwise this looks good to me.

gardner48 (approved):

PR #360, opened 2023-10-27
Feature/remove deprecated code

balos1:


examples/kinsol/serial/kinAnalytic_fp.c
```diff
@@ -231,11 +231,11 @@ int main(int argc, char *argv[])
   infofp = fopen("kinsol.log", "w");
   if (check_retval((void *)infofp, "fopen", 0)) return(1);
 
-  retval = KINSetInfoFile(kmem, infofp);
-  if (check_retval(&retval, "KINSetInfoFile", 1)) return(1);
+  // retval = KINSetInfoFile(kmem, infofp);
+  // if (check_retval(&retval, "KINSetInfoFile", 1)) return(1);
```
drreynolds, line 235: Can we just delete these lines instead of commenting them out?
```suggestion
```

examples/kinsol/serial/kinAnalytic_fp.c
```diff
@@ -231,11 +231,11 @@ int main(int argc, char *argv[])
...
 
-  retval = KINSetPrintLevel(kmem, 1);
-  if (check_retval(&retval, "KINSetPrintLevel", 1)) return(1);
+  // retval = KINSetPrintLevel(kmem, 1);
+  // if (check_retval(&retval, "KINSetPrintLevel", 1)) return(1);
```
drreynolds, line 238: Can we just delete these lines instead of commenting them out?
```suggestion
```

examples/kinsol/serial/kinLaplace_picard_kry.c
```diff
@@ -144,11 +144,11 @@ int main()
 
   infofp = fopen("KINSOL.log", "w");
 
-  retval = KINSetInfoFile(kmem, infofp);
-  if (check_retval(&retval, "KINSetInfoFile", 1)) return(1);
+  // retval = KINSetInfoFile(kmem, infofp);
+  // if (check_retval(&retval, "KINSetInfoFile", 1)) return(1);
```
drreynolds, line 148: Can we just delete these lines instead of commenting them out?
```suggestion
```

examples/kinsol/serial/kinLaplace_picard_kry.c
```diff
@@ -144,11 +144,11 @@ int main()
...
 
-  retval = KINSetPrintLevel(kmem, 3);
-  if (check_retval(&retval, "KINSetPrintLevel", 1)) return(1);
+  // retval = KINSetPrintLevel(kmem, 3);
+  // if (check_retval(&retval, "KINSetPrintLevel", 1)) return(1);
```
drreynolds, line 151: Can we just delete these lines instead of commenting them out?
```suggestion
```

drreynolds (approved):
Oooh, it feels nice deleting this old stuff!

examples/kinsol/serial/kinLaplace_picard_kry.c
balos1: Yep, I meant to do so

balos1:
Looks like all of the tests are passing now except for the extended precision.

examples/arkode/CXX_xbraid/ark_heat2D_hypre_pfmg_xbraid.cpp
```diff
@@ -742,12 +702,10 @@ int main(int argc, char* argv[])
   // Clean up and return
   // --------------------
 
-  if ((udata->diagnostics || udata->lsinfo) && udata->myid_c == 0) fclose(diagfp);
-
   ARKStepFree(&arkode_mem);              // Free integrator memory
   SUNLinSolFree(LS);                     // Free linear solver
   N_VDestroy(u);                         // Free vectors
-  FreeUserData(udata);                   // Free user data
+  FreeUserData(udata);                   // Free user dat
```
drreynolds, line 708:
```suggestion
  FreeUserData(udata);                   // Free user data
```

drreynolds (dismissed):

drreynolds (approved):

balos1:
I think this is ready to be merged now

include/nvector/nvector_openmpdev.h
```diff
@@ -195,22 +195,6 @@ SUNDIALS_EXPORT int N_VEnableWrmsNormMaskVectorArray_OpenMPDEV(N_Vector v, boole
...
- */
-
-/* use N_VCloneVectorArray */
-SUNDIALS_DEPRECATED_EXPORT N_Vector *N_VCloneVectorArray_OpenMPDEV(int count, N_Vector w);
-
-/* use N_VCloneVectorArrayEmpty */
-SUNDIALS_DEPRECATED_EXPORT N_Vector *N_VCloneVectorArrayEmpty_OpenMPDEV(int count, N_Vector w);
-
-/* use N_VDestroyVectorArray */
-SUNDIALS_DEPRECATED_EXPORT void N_VDestroyVectorArray_OpenMPDEV(N_Vector *vs, int count);
```
gardner48, line 211: Remove implementations in source

include/nvector/nvector_parallel.h
```diff
@@ -225,21 +225,6 @@ SUNDIALS_EXPORT int N_VEnableLinearCombinationVectorArray_Parallel(N_Vector v, b
...
- */
-
-/* use N_VCloneVectorArray */
-SUNDIALS_DEPRECATED_EXPORT N_Vector* N_VCloneVectorArray_Parallel(int count, N_Vector w);
-
-/* use N_VCloneVectorArrayEmpty */
-SUNDIALS_DEPRECATED_EXPORT N_Vector* N_VCloneVectorArrayEmpty_Parallel(int count, N_Vector w);
-
-/* use N_VDestroyVectorArray */
-SUNDIALS_DEPRECATED_EXPORT void N_VDestroyVectorArray_Parallel(N_Vector* vs, int count);
```
gardner48, line 241: Remove implementations in source

include/nvector/nvector_parhyp.h
```diff
@@ -211,24 +211,6 @@ SUNDIALS_EXPORT int N_VEnableScaleAddMultiVectorArray_ParHyp(N_Vector v,
...
-                                                                N_Vector w);
-
-/* use N_VCloneVectorArrayEmpty */
-SUNDIALS_DEPRECATED_EXPORT N_Vector *N_VCloneVectorArrayEmpty_ParHyp(int count,
-                                                                     N_Vector w);
-
-/* use N_VDestroyVectorArray */
-SUNDIALS_DEPRECATED_EXPORT void N_VDestroyVectorArray_ParHyp(N_Vector *vs,
-                                                             int count);
-
```
gardner48, line 231: Remove implementations in source

include/nvector/nvector_petsc.h
```diff
@@ -195,21 +195,6 @@ SUNDIALS_EXPORT int N_VEnableLinearCombinationVectorArray_Petsc(N_Vector v, bool
...
- */
-
-/* use N_VCloneVectorArray */
-SUNDIALS_DEPRECATED_EXPORT N_Vector *N_VCloneVectorArray_Petsc(int count, N_Vector w);
-
-/* use N_VCloneVectorArrayEmpty */
-SUNDIALS_DEPRECATED_EXPORT N_Vector *N_VCloneVectorArrayEmpty_Petsc(int count, N_Vector w);
-
-/* use N_VDestroyVectorArray */
-SUNDIALS_DEPRECATED_EXPORT void N_VDestroyVectorArray_Petsc(N_Vector *vs, int count);
```
gardner48, line 211: Remove implementations in source

include/nvector/nvector_pthreads.h
```diff
@@ -243,24 +243,6 @@ SUNDIALS_EXPORT int N_VEnableScaleAddMultiVectorArray_Pthreads(N_Vector v,
...
-                                                                  N_Vector w);
-
-/* use N_VCloneVectorArrayEmpty */
-SUNDIALS_DEPRECATED_EXPORT N_Vector* N_VCloneVectorArrayEmpty_Pthreads(int count,
-                                                                       N_Vector w);
-
-/* use N_VDestroyVectorArray */
-SUNDIALS_DEPRECATED_EXPORT void N_VDestroyVectorArray_Pthreads(N_Vector* vs,
-                                                               int count);
-
```
gardner48, line 263: Remove implementations in source

include/nvector/nvector_serial.h
```diff
@@ -184,21 +184,6 @@ SUNDIALS_EXPORT int N_VEnableWrmsNormMaskVectorArray_Serial(N_Vector v, booleant
...
- */
-
-/* use N_VCloneVectorArray */
-SUNDIALS_DEPRECATED_EXPORT N_Vector* N_VCloneVectorArray_Serial(int count, N_Vector w);
-
-/* use N_VCloneVectorArrayEmpty */
-SUNDIALS_DEPRECATED_EXPORT N_Vector* N_VCloneVectorArrayEmpty_Serial(int count, N_Vector w);
-
-/* use N_VDestroyVectorArray */
-SUNDIALS_DEPRECATED_EXPORT void N_VDestroyVectorArray_Serial(N_Vector* vs, int count);
```
gardner48, line 200: Remove implementations in source

src/arkode/arkode_adapt.c
```diff
@@ -136,10 +136,18 @@ int arkAdapt(void* arkode_mem, ARKodeHAdaptMem hadapt_mem,
...
+                     "ARKODE::arkAdapt", "error-history",
+                     "ecur = %"RSYM", ehist[0] = %"RSYM", ehist[0] = %"RSYM,
+                     ecur, hadapt_mem->ehist[0], hadapt_mem->ehist[1]);
+
+  SUNLogger_QueueMsg(ARK_LOGGER, SUN_LOGLEVEL_DEBUG,
+                     "ARKODE::arkAdapt", "step-history",
+                     "hcur = %"RSYM", hhist[0] = %"RSYM", hhist[0] = %"RSYM,
+                     hcur, hadapt_mem->hhist[0], hadapt_mem->hhist[1]);
+
+  SUNLogger_QueueMsg(ARK_LOGGER, SUN_LOGLEVEL_DEBUG,
```
gardner48, line 150: It would be good to add a CI test that loops over the different logging levels and just builds the code to catch logging level specific errors like this one
```suggestion
```

src/kinsol/kinsol.c
```diff
@@ -2375,17 +2382,7 @@ void KINInfoHandler(const char *module, const char *function,
...
-#ifndef NO_FPRINTF_OUTPUT
-  if (kin_mem->kin_infofp != NULL &&
-      kin_mem->kin_infofp != kin_mem->kin_sunctx->logger->info_fp) {
-    fprintf(kin_mem->kin_infofp,"\n[%s] %s\n",module, function);
-    fprintf(kin_mem->kin_infofp,"   %s\n",msg);
-  }
-#endif
 #if SUNDIALS_LOGGING_LEVEL >= SUNDIALS_LOGGING_INFO
   SUNLogger_QueueMsg(KIN_LOGGER, SUN_LOGLEVEL_INFO,
     "KINSOL::KINInfoHandler", function, "%s", msg);
```
gardner48, line 2388: Looks like the `module` input is lost, should this should this be something like `[KINSOL::<module>]`

gardner48: Also this could probably be moved into `KINPrintInfo` and `kin_ihfun` removed since the user can't override it any more

src/kinsol/kinsol.c
```diff
@@ -301,18 +301,9 @@ void *KINCreate(SUNContext sunctx)
   kin_mem->kin_omega_max        = OMEGA_MAX;
 
   kin_mem->kin_errfp            = stderr;
-  kin_mem->kin_infofp           = stdout;
-  kin_mem->kin_debugfp          = stdout;
-  kin_mem->kin_printfl          = PRINTFL_DEFAULT;
```
gardner48, line 306: Remove from `kinsol_impl.h`

gardner48 (changes requested):

src/arkode/arkode_erkstep_io.c
```diff
@@ -45,8 +45,6 @@ int ERKStepSetErrFile(void *arkode_mem, FILE *errfp) {
   return(arkSetErrFile(arkode_mem, errfp)); }
 int ERKStepSetUserData(void *arkode_mem, void *user_data) {
   return(arkSetUserData(arkode_mem, user_data)); }
-int ERKStepSetDiagnostics(void *arkode_mem, FILE *diagfp) {
-  return(arkSetDiagnostics(arkode_mem, diagfp)); }
```
gardner48, line 49: Remove `arkSetDiagonstics` in `arkode_io.c` and remove `report` and `diagfp` variables from `arkode_impl.h`

cmake/SundialsBuildOptionsPre.cmake
```diff
@@ -76,12 +76,12 @@ sundials_option(SUNDIALS_BUILD_WITH_PROFILING BOOL "${DOCSTR}" OFF)
 # ---------------------------------------------------------------
 
 set(DOCSTR "Build with logging capabilities enabled (0 = no logging, 1 = errors, 2 = +warnings, 3 = +info, 4 = +debug, 5 = +extras")
-sundials_option(SUNDIALS_LOGGING_LEVEL STRING "${DOCSTR}" 0
+sundials_option(SUNDIALS_LOGGING_LEVEL STRING "${DOCSTR}" 3
```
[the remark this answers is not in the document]
balos1: I think info level is the right level but we need to be careful what we put in the info level (this is why I changed a lot of the ARKODE output to `DEBUG` level). If we switch only do errors and warnings, then things that previously were always accessible, like the nonlinear solver iteration info would not be usable by default.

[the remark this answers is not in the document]
balos1: Also, the info level has no additional output compared to what we currently have in v6 now that I changed most (or all) of the ARKODE diagnostic output to be at the debug level.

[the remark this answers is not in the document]
gardner48: Would there still be additional function call overhead with level `3` vs level `2` even if it's not written out? If we use the info level for getting the history of statistics over time then this could also involve additional conditionals for what needs to be output.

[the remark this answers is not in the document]
balos1: The diagnostic output in ARKODE is guarded with an ifdef that checks the level, so no there is no extra overhead for those. If you left out the ifdef's then, yes, there would still be the additional conditionals. There should be zero overhead between level 2 and 3 with _this_ branch.  What we add to level 3 in the future could introduce overhead, of course, which is why I said we will want to carefully consider what is "INFO" and what is for debugging. 

I am not sure what exactly you mean by a "history of statistics over time", but if/when we go to implement it, we would want to consider if it should be at the INFO level or DEBUG level.

There are two downsides to choosing level `2` as the default:
1. KINSOL's long-standing info functions which would not be usable by default
2. The newer (introduced maybe in version 5?) SUNNonlinearSolver and SUNLinearSolver "info" functions would not be usable by default. 
Users of these functions would have to choose level `3`, so defaulting to `2` may break things for them. 

Since we are doing a major release, I suppose thats fine. We just need to make sure that we are OK with it. If we are, then level `2` should be fine as the default.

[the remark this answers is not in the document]
gardner48: By "history of statistics over time" I mean things like step size history, nonlinear/linear iterations per step, nonlinear/linear residuals per iteration, etc. To me these seem like output that would live at the INFO level as they're useful for understanding the behavior/performance of a method. This is potentially more output than we would want enabled be default and is why I'm thinking a default of `2`. This would break the two cases you note and that's something we should discuss. 

For now I think this PR can leave the default unchanged so we can discuss further without delaying the other changes in the PR.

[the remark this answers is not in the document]
balos1: By "this PR can leave the default unchanged" do you mean leaving it as `0` or leaving it as-is in the PR right now (`3`) ?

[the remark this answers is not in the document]
gardner48: I mean leave it as it was before, `0`

src/arkode/arkode_adapt.c
balos1: I agree, but I think this needs to be done later in a new PR.

src/kinsol/kinsol.c
balos1: The `module` input is not really relevant anymore though, is it? Its also not consistent with the rest logging in the other packages.

gardner48: `module` in this case is really the function that logged the message i.e., `KINFullNewton` so we'd replace `"KINSOL::KINInfoHandler"` with `"KINSOL::KINFullNewton"` which would align with the other packages e.g. `"ARKODE::arkEvolve"`

gardner48: Looks like I'm confusing the inputs. You're right `module` can be ignored. It should be `"KINSOL::<function>"` to align with the other messages and I guess where `function` is now would be empty.

balos1: Yes, so we just need to drop `::KINInfoHandler`.  The function will get added since the function name gets passed to QueueMsg which appends it.

cmake/SundialsBuildOptionsPre.cmake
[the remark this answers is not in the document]
balos1: I think if we are going to do that, might as well make it at least `2`. It has to be at least `2` for the error handling.  Ill switch it to `2` for now, but we will need to discuss it on Tuesday.

src/arkode/arkode_adapt.c
gardner48: The CI test can be added later but this line needs to be removed to compile with debug logging.

gardner48:
You're fixes to the C++ interfaces to add `SUNComm` reminded me we should probably remove things like
```
/* We include this here for backwards compatibility
   (the contents used to be defined here directly) */
#include <sundials/sundials_context.hpp>
```
in `sundials_context.h` and have C++ users include the `.hpp` file. I think this only occurs in `sundials_context.h` and `sundials_profiler.h`.

src/arkode/arkode_adapt.c
balos1: Ah, I see now. I quickly tweaked one of the GitHub action workflows to cycle through the logging options.

balos1:
> Your fixes to the C++ interfaces to add `SUNComm` reminded me we should probably remove things like
> 
> ```
> /* We include this here for backwards compatibility
>    (the contents used to be defined here directly) */
> #include <sundials/sundials_context.hpp>
> ```
> 
> in `sundials_context.h` and have C++ users include the `.hpp` file. I think this only occurs in `sundials_context.h` and `sundials_profiler.h`.

Lets leave this for the error handling PR because i rearranged some includes there

balos1:
GitLab CI is reporting as failed since Corona is (or was) dow.

balos1:
> GitLab CI is reporting as failed since Corona is (or was) down, i'll restart them.

OK, all (minus extended) are passing now.

drreynolds (approved):

gardner48 (approved):

PR #362, opened 2023-11-02
Bugfix: ARKODE ONE_STEP Description

gardner48:
Fix the description of `ONE_STEP` mode in ARKODE. The solution at the end of the step is always returned and interpolation is not performed.

### DIFF
--- a/doc/arkode/guide/source/Usage/ARKStep_c_interface/User_callable.rst
+++ b/doc/arkode/guide/source/Usage/ARKStep_c_interface/User_callable.rst
@@ -726,11 +726,8 @@ the user has set a stop time (with a call to the optional input function
         in :numref:`ARKODE.Mathematics.Interpolation`).
 
         The *ARK_ONE_STEP* option tells the solver to only take a
-        single internal step :math:`y_{n-1} \to y_{n}` and then return
-        control back to the calling program.  If this step will
-        overtake *tout* then the solver will again return an
-        interpolated result; otherwise it will return a copy of the
-        internal solution :math:`y_{n}` in the vector *yout*.
+        single internal step, :math:`y_{n-1} \to y_{n}`, and return the solution
+        at that point, :math:`y_{n}`, in the vector *yout*.
 
    **Return value:**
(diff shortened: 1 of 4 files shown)


doc/arkode/guide/source/Usage/ARKStep_c_interface/User_callable.rst
```diff
@@ -726,11 +726,8 @@ the user has set a stop time (with a call to the optional input function
         in :numref:`ARKODE.Mathematics.Interpolation`).
 
         The *ARK_ONE_STEP* option tells the solver to only take a
-        single internal step :math:`y_{n-1} \to y_{n}` and then return
-        control back to the calling program.  If this step will
-        overtake *tout* then the solver will again return an
-        interpolated result; otherwise it will return a copy of the
-        internal solution :math:`y_{n}` in the vector *yout*.
+        single internal step, :math:`y_{n-1} \to y_{n}`, and return the solution
```
balos1, line 729: So if I am understanding correctly, we actually do not interpolate if the step will overtake *tout* in one step mode?

gardner48: Right, one step mode always takes a single step and returns the result even if it overtakes `tout`. This is how it's always been implemented but the documentation was incorrect (CVODE and IDA document the behavior correctly).

balos1 (approved):

doc/arkode/guide/source/Usage/ARKStep_c_interface/User_callable.rst
cswoodward: We have a lot of apps using one-step mode now. We should have the ability to interpolate back. I expect this will be more needed with ARKODE than CVODE or IDA.

gardner48: An interpolated output at the desired time can still be obtained using the `GetDky` functions which is what Normal mode does.

PR #372, opened 2023-11-11
Feature/rename free to destroy

balos1:


examples/arkode/F2003_custom/ark_brusselator1D_f2003.f90
```diff
@@ -393,7 +393,7 @@ program main
   ! clean up
   call FARKStepFree(arkode_mem)
```
drreynolds, line 394: Should this be changed to 
```
call FARKStepDestroy(arkode_mem)
```
?

examples/arkode/F2003_custom/ark_brusselator1D_f2003.f90
```diff
@@ -393,7 +393,7 @@ program main
...
   call FN_VDestroy(sunvec_y)
-  call FSUNMatDestroy(sunmat_A)
+  ierr = FSUNMatDestroy(sunmat_A)
   ierr = FSUNLinSolFree(sunls)
   ierr = FSUNContext_Free(sunctx)
```
drreynolds, line 398: Should these be changed to 
```
ierr = FSUNLinSolDestroy(sunls)
ierr = FSUNContext_Destroy(sunctx)
```
?

drreynolds (changes requested):
I'm sorry, but I'd really like for someone to *convince* me of the need for such intrusive changes throughout the entire code suite.  SUNDIALS has lived with some modules using `Free` and others using `Destroy` since it was originally released 20 years ago, and as far as I can tell the only folks who take issue are us developers.  While I totally understand deprecating unused/broken/ill-designed modules and routines, deprecating for the sake of uniformity seems more like we're overburdening our users just so that we can have things tidier throughout the repository.


That said, if the goal really is for uniformity, then this PR still needs a lot of work:

1. It looks like the majority of our modules actually used `Free` and only a minority used `Destroy`, so why go the more-intrusive route, instead of naming them all `Free`?
2. Why do some modules use `_Destroy` while others have no underscore?
3. It seems like the only changes in the example files had to do with the now-integer return value from `SUNMatDestroy`.  However, all examples still seem to call the now-deprecated `Free` routines, and should be converted to use the new versions.

balos1:
> I'm sorry, but I'd really like for someone to convince me of the need for such intrusive changes throughout the entire code suite. SUNDIALS has lived with some modules using Free and others using Destroy since it was originally released 20 years ago, and as far as I can tell the only folks who take issue are us developers.

It's an annoying inconsistency in the API in both naming and also that some take a pointer to the object and others just take the object. I agree that this is very intrusive change with limited benefit. I don't recall most of the details discussions when it was decided that we would actually make this change, but everyone agreed to it at that moment. Regardless, I am now of the opinion (after seeing how intrusive it is) that deprecating and eventually removing the Free functions seems frivolous. I see three options:

1. We don't do anything and discard this PR. Moving forward we stick to either `Free` or `Destroy` by convention. I do remember that after some debate `Destroy` was chosen to be the preferred name.
2. We keep this PR but don't deprecate the Free functions. I.e., only add `Destroy` functions (which will all take a pointer to an object) and support both indefinitely. This way we have a consistent API available but don't force users to make the frivolous change. Supporting both is not hard, the `Free` functions just become wrappers (as they are in this PR already) to `Destroy`. 
3. We keep the PR as is.

I'll answer the other comments on some of the details of the implementation in this PR once we decide on a path forward.

drreynolds:
> I see three options:
> 
> 1. We don't do anything and discard this PR. Moving forward we stick to either `Free` or `Destroy` by convention. I do remember that after some debate `Destroy` was chosen to be the preferred name.
> 2. We keep this PR but don't deprecate the Free functions. I.e., only add `Destroy` functions (which will all take a pointer to an object) and support both indefinitely. This way we have a consistent API available but don't force users to make the frivolous change. Supporting both is not hard, the `Free` functions just become wrappers (as they are in this PR already) to `Destroy`.
> 3. We keep the PR as is.
> 
> I'll answer the other comments on some of the details of the implementation in this PR once we decide on a path forward since the answer may change depending on what is decided.

If uniformity is the goal, then I recommend that we just add Free/Destroy routines to each base class that will wrap the existing base class routine, thereby allowing a consistent API without forcing so many backwards-incompatible changes on users, and without even increasing the size of our vtable or requiring changes to user-supplied modules.

balos1:
Closing for now.

PR #379, opened 2023-12-04
Feature/error handling just the core formatted

balos1:
Other than applying clang-format and clang-tidy, there are changes/additions in the following files:

- .clang-format
- .clang-tidy
- .github/actions/free-space-ubuntu/action.yml
- .github/workflows/build-llvm-container.yml
- .github/workflows/check-clang-format.yml
- scripts/format.sh
- cmake/SundialsBuildOptionsPre.cmake
- cmake/SundialsSetupCompilers.cmake

Todo after merging:

- [ ] Setup .git-blame-ignore-revs file [so blame still works](https://stackoverflow.com/questions/53502654/how-do-i-run-a-code-formatter-over-my-source-without-modifying-git-history)


scripts/format.sh
drreynolds, line 1: Is this file really supposed to be empty?

balos1: No, it is not.

drreynolds (approved):

scripts/format.sh
balos1: I am not sure why GitHub says the file is empty. It is not.

gardner48 (commented):
Overall this looks good. Something in the last commit is causing format check action to fail.

PR #387, opened 2023-12-06
Bugfix/arkode vector requirements

drreynolds:
Updated table showing N_Vector requirements from ARKODE.  This is in support of some SciDAC partnerships, who are beginning to create custom N_Vector implementations to interface with ARKODE, and who want an up-to-date list of the minimum N_Vector routines they'll need to implement.


balos1 (approved):

balos1:
@gardner48 Can you add this PR to Jenkins?

PR #394, opened 2023-12-12
Changes in feature/error-handling-staging 

balos1:
These are just the changes in feature/error-handling-staging that the other branches are based on.


src/sundials/sundials_errors.c
```diff
@@ -77,10 +75,38 @@ void SUNAbortErrHandlerFn(int line, const char* func, const char* file,
...
+{
+  va_list ap;
+  char* log_msg       = NULL;
+  char* file_and_line = NULL;
+
+  va_start(ap, err_code);
+
+  file_and_line = sunCombineFileAndLine(__LINE__, __FILE__);
+  sunCreateLogMessage(SUN_LOGLEVEL_ERROR, 0, file_and_line,
+                      __func__, "The SUNDIALS SUNContext was corrupt or NULL when an error occurred. As such, error messages have been printed to stderr.",
```
drreynolds, line 97: Is it really *always* the case that if `SUNGlobalFallbackErrHandler` is called, then the SUNDIALS SUNContext was corrupt or NULL when the error occurred, or is it possible to use this fallback function just because no other error handlers were specified?  If this could be used in any other situation, then I recommend that this line be adjusted.

balos1: It will only ever be called if the `SUNContext` is `NULL` or the object holding the `SUNContext` is `NULL` and any use of it otherwise should be avoided. If there is no error handler attached, errors will be ignored (by design).

src/sundials/sundials_logger_impl.h
```diff
@@ -59,4 +60,37 @@ struct SUNLogger_
   SUNErrCode (*destroy)(SUNLogger* logger);
 };
 
+static void sunCreateLogMessage(SUNLogLevel lvl, int rank, const char* scope,
```
gardner48, line 63: Does this need to be define in the header?

src/sundials/sundials_mpi_errors.c
```diff
@@ -18,7 +18,7 @@
 #include <sundials/sundials_mpi_types.h>
 #include <unistd.h>
 
-static inline char* combineFileAndLine(int line, const char* file)
+static inline char* sunCombineFileAndLine(int line, const char* file)
```
gardner48, line 21: Why not use the version from `sundials_utils.h`?

gardner48 (changes requested):

scripts/format.sh
```diff
@@ -13,11 +13,13 @@
...
 #
 # Usage:
-#    ./format.sh <path to directory to format>
+#    ./format.sh <paths to directories or files to format>
 # 
 # We require clang-format 17.0.4. Other versions may produce different styles!
 # ---------------------------------------------------------------------------------
 
-find $1 -iname '*.h' -o -iname '*.hpp' -o \
+paths="$@"
```
gardner48, line 21: In `tidy_and_format.sh` if you reverse the input order so it's build path then directories format that would make it easier to pass multiple path to that script as well if that's helpful.

scripts/format.sh
```diff
@@ -13,11 +13,13 @@
...
+
+find $paths -iname '*.h' -o -iname '*.hpp' -o \
```
gardner48, line 23: shellcheck suggests something like (I added the input check because I forgot when using the script)
```suggestion
if [ $# -lt 1 ]; then
    echo "ERROR: At least one path to format required"
    exit 1
fi

paths=( "$@" )

find "${paths[@]}" -iname '*.h' -o -iname '*.hpp' -o \
```

src/sundials/fmod/CMakeLists.txt
```diff
@@ -19,8 +19,6 @@ set(sundials_SOURCES
   fsundials_adaptcontroller_mod.f90
   fsundials_context_mod.c
   fsundials_context_mod.f90
-  fsundials_errors_mod.c
-  fsundials_errors_mod.f90
```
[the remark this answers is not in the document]
balos1: Yes, thats right. I wanted to avoid forcing users to have to `use` yet another module. I actually think we should combine all of the core modules into one to mimic how you would link to the libraries or include headers.  I have a branch where it is 95% done.

scripts/format.sh
balos1: I actually am going to remove `tidy_and_format.sh` because the clang-tidy part doesn't work too well (I had to do a lot of extra stuff to make it work). The easier way to apply clang-tidy is by enabling it in CMake (in the formatting PR I added this option too).

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+78770ef9b40fb163bf6d0fd582b6a449f0298585
+e75d0c03345b230def5713cda2e1b2748171634b
```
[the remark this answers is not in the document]
balos1: The second hash is for the `realtype --> sunrealtype` change  (e.g.  and the change for the related macros e.g. `RCONST`).

include/sundials/priv/sundials_errors_impl.h
```diff
@@ -27,6 +27,10 @@
 #include "sundials/sundials_logger.h"
 #include "sundials/sundials_types.h"
 
+#ifdef __cplusplus /* wrapper to enable C++ usage */
+extern "C" {
+#endif
```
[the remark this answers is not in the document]
balos1: Since enums do not have linkage, it does not matter.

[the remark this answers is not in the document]
balos1: I moved the end of the extern to the bottom though, just for clarity.

src/sundials/sundials_logger_impl.h
balos1: Need to be - no, but it doesn't hurt for this utility I don't think.

src/sundials/sundials_mpi_errors.c
balos1: It should use it

src/sundials/sundials_logger_impl.h
gardner48: It doesn't hurt but we generally don't define things in a header unless they need to be there (i.e., to inline) here and this was in a `.c` file before

balos1: Oops, I read this as `inline` already. It would need `inline` to stay in the header. Ill move it.

include/sundials/priv/sundials_errors_impl.h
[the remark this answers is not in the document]
balos1: > Looking through the headers I noticed `sundials_context.h` still includes `sundials/sundials_context.hpp`, I thought that was removed in an earlier PR.

Looks like we [deferred it](https://github.com/LLNL/sundials/pull/360#issuecomment-1806533562) to the "error handling PR". Making this change does require updating a few dozen examples though, this is why the build is failing now.

examples/cvode/hip/cvAdvDiff_kry_hip.cpp
```diff
@@ -41,7 +41,8 @@
 #include <nvector/nvector_hip.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <sundials/sundials_math.h>    /* definition of ABS and EXP   */
+#include <sundials/sundials_core.hpp>
+#include <sundials/sundials_core.hpp>  /* definition of ABS and EXP   */
```
gardner48, line 45: Included twice
```suggestion
```

doc/superbuild/source/developers/style_guide/SourceCode.rst
```diff
@@ -395,46 +338,81 @@ Coding Conventions and Rules
...
+         .doSomething()       //
+         .doAnotherThing()    //
+         .doSomethingElse();  //
+
+There are other scenarios (e.g., a function call with a lot of parameters) where doing this type of line break is useful for readability too.
+
+..
+   .. code-block::
+..    cd scripts
+..    ./tidy_and_format.sh <path to directory to tidy and format> <path to builddir>
```
gardner48, line 384: `tidy_and_format.sh` was removed
```suggestion
..    ./format.sh <path to directory to tidy and format> <path to builddir>
```

doc/superbuild/source/developers/style_guide/SourceCode.rst
```diff
@@ -395,46 +338,81 @@ Coding Conventions and Rules
...
+.. code-block:: cpp
+
+   MyObject::callAFunctionOfSorts()
+         .doSomething()       //
+         .doAnotherThing()    //
+         .doSomethingElse();  //
+
+There are other scenarios (e.g., a function call with a lot of parameters) where doing this type of line break is useful for readability too.
+
+..
```
gardner48, line 381: Did you intend for the text below to be commented out?

scripts/tidy_and_format.sh
gardner48, line 1: Why was this removed? Was there an issue with using `clang-format`? Do the `GitHub` actions need to be updated?

gardner48 (changes requested):
Minor stuff

scripts/tidy_and_format.sh
gardner48: I an earlier comment you said that the clang-tidy part didn't work too well. Did you mean clang-format?

balos1: `clang-format` - formats the code
`clang-tidy` - lints the code, I used it to add missing braces and nothing else

The GitHub action uses `format.sh` which just does `clang-format` only, so no the action does not need to be updated.

I can explain in more detail about the difficulties of using `clang-tidy` offline.

doc/superbuild/source/developers/style_guide/SourceCode.rst
balos1: Yes, I did not want to lose what I had wrote but I also did not want it visible in the developer docs at this moment.

gardner48 (approved):

balos1:
OK, now that this is approved, I am going to close it since the stuff is already in feature/error-handling-staging.

PR #407, opened 2024-01-10
Import installed CMake target for KLU from SuiteSparse

mmuetzel:
Newer versions of SuiteSparse install import targets for its libraries.
Use the import target of KLU as installed by SuiteSparse if it is available and the user didn't set flags to select a different one.

This fixes an issue if the SuiteSparse headers are installed at their new default location `$prefix/include/suitesparse`.

~~Also, modify the CI rules for MinGW to also install and build with SuiteSparse. (I hope it is ok to have that in the same PR.)~~


cmake/tpl/FindKLU.cmake
```diff
@@ -30,6 +30,18 @@
 #   KLU_LIBRARIES   - all of the libraries needed for KLU
 # ---------------------------------------------------------------
 
+if (NOT (KLU_INCLUDE_DIR OR KLU_LIBRARY_DIR OR KLU_LIBRARY))
+  # prefer import target from upstream SuiteSparse if available and
+  # user didn't point to a specific (different) version
+  find_package(KLU CONFIG)
+
+  if(TARGET SuiteSparse::KLU)
+    set(KLU_FOUND ON)
```
mmuetzel, line 39: Setting `KLU_FOUND` to `ON` manually is only needed because of a bug in SuiteSparse 7.4.0. It should be fixed in SuiteSparse 7.5.0 (which will probably be released shortly).

mmuetzel:
I stripped the change to the CI rules from the PR. Maybe, that will make it a bit easier to review.

cmake/tpl/FindKLU.cmake
```diff
@@ -30,6 +30,18 @@
 #   KLU_LIBRARIES   - all of the libraries needed for KLU
 # ---------------------------------------------------------------
 
+if (NOT (KLU_INCLUDE_DIR OR KLU_LIBRARY_DIR OR KLU_LIBRARY))
+  # prefer import target from upstream SuiteSparse if available and
+  # user didn't point to a specific (different) version
+  include ( CMakeFindDependencyMacro )
+  find_dependency(KLU CONFIG)
```
balos1, line 37:
```suggestion
  find_package(KLU REQUIRED)
```
We prefer to use `find_package`.

balos1 (changes requested):

cmake/tpl/FindKLU.cmake
mmuetzel: I can switch to `find_dependency` (but that won't propagate flags like `SILENT` from the parent).
Adding `REQUIRED` would break the configuration if the installed SuiteSparse is older than version 7.4.0.
Removing `CONFIG` might not fetch the `KLUConfig.cmake` as installed by SuiteSparse 7.4.0 or later.

cmake/tpl/FindKLU.cmake
```diff
@@ -30,6 +30,18 @@
...
+
+  if(TARGET SuiteSparse::KLU)
+    add_library(SUNDIALS::KLU ALIAS SuiteSparse::KLU)
```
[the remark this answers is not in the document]
mmuetzel: I can add that. 
But could you please explain in which case that check would be necessary?

cmake/tpl/FindKLU.cmake
```diff
@@ -30,6 +30,18 @@
 #   KLU_LIBRARIES   - all of the libraries needed for KLU
 # ---------------------------------------------------------------
 
+if (NOT (KLU_INCLUDE_DIR OR KLU_LIBRARY_DIR OR KLU_LIBRARY))
+  # prefer import target from upstream SuiteSparse if available and
```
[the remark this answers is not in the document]
mmuetzel: I meant to refer to the "import target" that is provided by the `KLUConfig.cmake` file in newer versions of SuiteSparse.

cmake/tpl/FindKLU.cmake
balos1: OK, we can remove `REQUIRED` to allow for the older versions, but we will need to use `find_package` for consistency with how we treat other TPLs (we use `find_dependency` in the SUNDIALSConfig.cmake, but not in our custom Find modules).

cmake/tpl/FindKLU.cmake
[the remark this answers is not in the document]
balos1: If we happen to do `find_package(KLU)` twice, this ensures that we don't recreate the target.  It is unlikely to happen, but we do this in all of our Find modules so this makes it consistent.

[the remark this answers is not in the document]
mmuetzel: Would an early return be better for that case?

cmake/tpl/FindKLU.cmake
```diff
@@ -30,6 +30,18 @@
...
+  find_dependency(KLU CONFIG)
+
+  if(TARGET SuiteSparse::KLU)
+    add_library(SUNDIALS::KLU ALIAS SuiteSparse::KLU)
+    return()
+  endif()
+endif()
+
 # Set library prefixes for Windows
 if(WIN32)
```
mmuetzel, line 46: Unrelated to this change. But for MinGW, setting the library prefix is not necessary and setting the library suffix to d.lib is incorrect. The static library suffix is `.a` for MinGW.
``` suggestion
if(MSCV)
```

Would it be ok to change that in the same PR?

cmake/tpl/FindKLU.cmake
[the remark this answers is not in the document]
mmuetzel: I changed that comment to a "proper sentence". Is that ok?

cmake/tpl/FindKLU.cmake
mmuetzel: Prepending those values to the list of library prefixes and suffixes might not actually cause an issue if no libraries are found that match those patterns. But slight ABI differences between MSVC and MinGW could lead to unexpected behavior when mixing binaries of both targets.

balos1: Yeah, that change makes sense to include here.

cmake/tpl/FindKLU.cmake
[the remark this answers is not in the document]
balos1: Yes

balos1:
Please sign your last commit as indicated in the [Contributing guide](https://github.com/LLNL/sundials/blob/main/CONTRIBUTING.md).

mmuetzel:
OK like this?

cmake/tpl/FindKLU.cmake
```diff
@@ -30,6 +30,19 @@
 #   KLU_LIBRARIES   - all of the libraries needed for KLU
 # ---------------------------------------------------------------
 
+if (NOT (KLU_INCLUDE_DIR OR KLU_LIBRARY_DIR OR KLU_LIBRARY))
+  # Prefer the import target from upstream SuiteSparse if it is available
+  # and the user didn't point to a specific (different) version.
+  find_package(KLU CONFIG)
```
balos1, line 36: I haven't tried yet, but is it right? I am wondering if should it be `find_package(SuiteSparse CONFIG COMPONENTS KLU)`?

cmake/tpl/FindKLU.cmake
balos1, line 1: [This place in the docs](https://sundials.readthedocs.io/en/latest/Install_link.html#building-with-klu) should be updated to note that this option exists too.

balos1 (changes requested):

cmake/tpl/FindKLU.cmake
mmuetzel: SuiteSparse doesn't install one big Config file which contains information for the libraries in the form of components. Instead, the goal was to allow building each of the libraries as stand-alone projects.
To achieve that goal, SuiteSparse installs separate Config files for each of the libraries.

A similar change is used by MSYS2 to build SUNDIALS with newer versions of SuiteSparse:
https://github.com/msys2/MINGW-packages/commit/1dfb64a6ef97cfe5dc295bdf038a1d730b687081

So, that command is working.

cmake/tpl/FindKLU.cmake
mmuetzel: I don't know where that documentation gets generated (if it gets generated at all).

mmuetzel: I think I found it.

English isn't my native language. So, please rephrase if the documentation change sounds strange.

balos1: It is in `doc/shared/Install.rst`. If you click "edit on GitHub" in the top right corner on readthedocs, it will take you there (well sort of, its indirect since our documentation structure uses symlinks).

mmuetzel: When it comes to the existing documentation, parts of it is describing internals to the build system of SUNDIALS (like which CMake variables might be set).
I tried to keep describing the internals on a similar level. I don't know if any of that would be interesting to a user of SUNDIALS though...

mmuetzel:
Added another commit that imports the necessary target for dependent projects that link to a SUNDIALS target.

doc/shared/sundials/Install.rst
```diff
@@ -1336,11 +1336,22 @@ Texas A&M University and is available from the `SuiteSparse GitHub repository
...
+will result in populating the following variables: ``AMD_LIBRARY``,
 ``AMD_LIBRARY_DIR``,  ``BTF_LIBRARY``, ``BTF_LIBRARY_DIR``,
 ``COLAMD_LIBRARY``, ``COLAMD_LIBRARY_DIR``, and ``KLU_LIBRARY``.
 
+For SuiteSparse 7.4.0 and newer, the necessary information can also be gathered
+from a CMake import target.  If SuiteSparse is installed in a non-default
+prefix, the path to the CMake Config file can be set using
+``CMAKE_PREFIX_PATH``.  In that case, the CMake configure step won't populate
+the previously mentioned variables.  It is still possible to set
+``KLU_INCLUDE_DIR`` and ``KLU_LIBRARY_DIR`` which takes precedence over a
```
drreynolds, line 1349:
```suggestion
``KLU_INCLUDE_DIR`` and ``KLU_LIBRARY_DIR`` which take precedence over a
```

drreynolds (commented):
I noted a minor grammar fix in the documentation change, but otherwise this PR looks good to me.

Issue #329, opened 2023-08-24
Providing default constructors for structs breaks use of designated initializers


ianabel:
Using Sundials in a C++20 program results in no clean way to instantiate custom SUNLinearSolvers (and other obejcts).

The usual 
```
struct _generic_SUNLinearSolver_Ops LSOps =·
{
.gettype = SunLinSolWrapper::LSGetType,
.getid = SunLinSolWrapper::LSGetID,
.setatimes = nullptr, ...
};
```

Isn't permitted because the default constructor makes _generic_SUNLinearSolver_Ops a non-aggregate type. 

From include/sundials/sundials_linearsolver.h:127
```
#ifdef __cplusplus
  _generic_SUNLinearSolver_Ops() = default;
#endif
```

Removing this default constructor results in a pure aggregate type and declaring the default constructor should not be necessary.  

I know this is low priority, but flagging this as usage of C++20 will only increase.

ianabel:
This appears to have been added in the Ginko interface commit 86996b3 last october by @balos1 -- was it needed to get ginko working?

drreynolds:
Thanks for bringing this up @ianabel. @balos1 is on travel this week, but hopefully he'll chime in soon.

ianabel:
Right, I guess the question is -- was the default constructor needed for something specific. 

By my reading the C++ standard defines the behaviour of constructing an aggregate to be exactly what the default constructor would have done (in this case, with simple types). So it should be fine to just remove, but if there's a particular reason it was there I'm happy to debug whether removing it is an issue.

balos1:
I cannot recall exactly why the default constructor was added. Removing it does not seem to break anything in my limited testing so far. If you are able to try it out as well, that would be great. Here is the branch where I removed it https://github.com/LLNL/sundials/tree/bugfix/cpp20-aggregate.

PR #413, opened 2024-02-05
remove explicit default destructor from core classes

balos1:
Fixes https://github.com/LLNL/sundials/issues/329


drreynolds (dismissed):

gardner48 (changes requested):
Was providing the default constructor unnecessary before?
Need to update the change log to note the bug fix for C++20 users.

balos1:
Yes, it seems it was unnecessary before.

gardner48 (approved):

PR #429, opened 2024-02-27
Maintenance: update changelog

gardner48:
* Note a subset of C99 is now required
* Move addition to F2003 examples to v7


drreynolds (dismissed):
Nice catch with the F77 -> F2003 text!  Since our original plan had been to merge that after the release, I had it in the `vX.X.X` future block instead.

balos1 (approved):

balos1 (approved):

PR #432, opened 2024-02-29
CI (mingw): Install and build with SuiteSparse library

mmuetzel:
MSYS2 follows a rolling release model. I.e., most of its packages are quite new most of the time.
Adding the SuiteSparse libraries in the CI tests for MSYS2 might help to detect potentially API changes early.


mmuetzel:
> @mmuetzel Can you sign your commit as mentioned [in the contributing guide](https://github.com/LLNL/sundials/blob/main/CONTRIBUTING.md)?

I rebased on a current head of the `develop` branch and added the Signed-off-by at the end of the commit message.
Is that correct?

balos1:
> > @mmuetzel Can you sign your commit as mentioned [in the contributing guide](https://github.com/LLNL/sundials/blob/main/CONTRIBUTING.md)?
> 
> I rebased on a current head of the `develop` branch and added the Signed-off-by at the end of the commit message. Is that correct?

Yes, it looks good now.

mmuetzel:
It looks like the CI failed because GCC 14 is stricter when it comes to automatically converting pointer types than previous versions.
That is probably worth fixing independently.

mmuetzel:
I opened #477 which hopefully addresses that issue.
(IIUC, that is a real issue indeed if `SUNDIALS_INT64_T` is defined on Windows or other platforms that use an LLP64 data model.)

balos1 (approved):

PR #437, opened 2024-03-13
Bugfix: HIP_PLATFORM default

gardner48:
`HIP_PLATFORM` default to `amd` as the previous default, `hcc`, is no longer recognized in ROCm 5.7.0 or newer. The new default is also valid in older version of ROCm (at least back to version 4.3.1).


drreynolds (changes requested):
LGTM, except for a few typos.

drreynolds (approved):

PR #438, opened 2024-03-13
CMake: Link sundials_sunlinsolsuperlumt to blas

jschueller:
else it cannot be loaded:
/lib/x86_64-linux-gnu/libsundials_sunlinsolsuperlumt.so.4: undefined symbol: dtrsm_
probably because superlumt is usually static

### DIFF
--- a/src/sunlinsol/superlumt/CMakeLists.txt
+++ b/src/sunlinsol/superlumt/CMakeLists.txt
@@ -47,4 +47,8 @@ sundials_add_library(sundials_sunlinsolsuperlumt
   ${sunlinsollib_SOVERSION}
 )
 
+if (BLAS_FOUND)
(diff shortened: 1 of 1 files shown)


gardner48:
Hi @jschueller it looks like you had some trouble building with SuperLU_MT enabled. Try specifying the BLAS library SuperLU_MT depends on using the `SUPERLUMT_LIBRARIES` CMake variable and see if that resolves the problem.

It seems this variable is missing from the list of CMake variables in the install guide but there is some information about it [here](https://sundials.readthedocs.io/en/latest/sundials/Install_link.html#building-with-superlu-mt) (I'll make sure the install guide is updated accordingly).

jschueller:
yes, it works to set SUPERLUMT_LIBRARIES, thanks

PR #439, opened 2024-03-14
New Low Order MRI Methods

Steven-Roberts:
- Adds new 1st-3rd order MRI methods. Most are from the literature, but the rest are trivial to derive, e.g., Euler-based schemes.
- Adds some of the base ERK and IRK method tables too. Note that some cannot be equipped with reasonable embeddings.
- Fixes an MRI reallocation bug when switching coupling coefficients to one with more stages
- Fixes a workspace size bug for MRIStep
- Makes `MRIStepSetOrder` consistent with `ARKStepSetOrder` and `ERKStepSetOrder` so that a negative value sets the default order and a positive value out of range prints an error (when method initialization occurs not when `SetOrder` is called).

Before I update the docs, stability plots, changelog, etc, I want to make sure everyone's ok with these new methods.


src/arkode/arkode_mristep.c
```diff
@@ -1007,7 +1005,7 @@ int mriStep_Init(void* arkode_mem, int init_type)
   ARKodeMem ark_mem;
   ARKodeMRIStepMem step_mem;
   int retval, j;
-  sunbooleantype reset_efun;
```
drreynolds, line 1010: I'd like to keep this in, since the logic below will get even more complicated when MRI adaptivity is enabled.

src/arkode/arkode_mristep.c
```diff
@@ -1021,10 +1019,7 @@ int mriStep_Init(void* arkode_mem, int init_type)
   {
     /* enforce use of arkEwtSmallReal if using a fixed step size for
        an explicit method and an internal error weight function */
-    reset_efun = SUNTRUE;
-    if (step_mem->implicit_rhs) { reset_efun = SUNFALSE; }
-    if (ark_mem->user_efun) { reset_efun = SUNFALSE; }
-    if (reset_efun)
+    if (!step_mem->implicit_rhs && !ark_mem->user_efun)
```
drreynolds, line 1022: Please revert this change, since this logic will be even more complicated when MRI adaptivity is enabled.

src/arkode/arkode_mristep.c
```diff
@@ -1058,22 +1053,23 @@ int mriStep_Init(void* arkode_mem, int init_type)
...
     step_mem->stages = step_mem->MRIC->stages;
     step_mem->q      = step_mem->MRIC->q;
     step_mem->p      = step_mem->MRIC->p;
 
     /* allocate/fill derived quantities from MRIC structure */
 
     /* stage map */
-    if (step_mem->stage_map)
+    step_mem->stage_map = realloc(step_mem->stage_map,
+                                  step_mem->stages * sizeof(*step_mem->stage_map));
```
drreynolds, line 1065: I'm okay with changing to `realloc`, but (a) please cast the result to the correct type (`int*`), and (b) `memset` the new memory to all zeros to retain `calloc`'s initialization feature.

src/arkode/arkode_mristep.c
```diff
@@ -1082,42 +1078,65 @@ int mriStep_Init(void* arkode_mem, int init_type)
                       "Error in coupling table");
       return (ARK_ILL_INPUT);
     }
-
+    
     /* stage types */
-    if (step_mem->stagetypes)
+    step_mem->stagetypes = realloc(step_mem->stagetypes,
+                                   step_mem->stages * sizeof(*step_mem->stagetypes));
```
drreynolds, line 1084: I'm okay with changing to `realloc`, but (a) please cast the result to the correct type (`int*`), and (b) `memset` the new memory to all zeros to retain `calloc`'s initialization feature.

src/arkode/arkode_mristep.c
```diff
@@ -1082,42 +1078,65 @@ int mriStep_Init(void* arkode_mem, int init_type)
...
     for (j = 0; j < step_mem->stages; j++)
     {
       step_mem->stagetypes[j] = mriStepCoupling_GetStageType(step_mem->MRIC, j);
     }
-
+    
     /* explicit RK coefficient row */
-    if (step_mem->Ae_row)
+    step_mem->Ae_row = realloc(step_mem->Ae_row,
+                               step_mem->stages * sizeof(*step_mem->Ae_row));
```
drreynolds, line 1099: I'm okay with changing to `realloc`, but (a) please cast the result to the correct type (`sunrealtype*`), and (b) `memset` the new memory to all zeros to retain `calloc`'s initialization feature.

src/arkode/arkode_mristep.c
```diff
@@ -1082,42 +1078,65 @@ int mriStep_Init(void* arkode_mem, int init_type)
...
     }
-    step_mem->Ae_row = (sunrealtype*)calloc(step_mem->stages,
-                                            sizeof(sunrealtype));
-    ark_mem->lrw += step_mem->stages;
+    ark_mem->lrw += delta_stages;
 
     /* implicit RK coefficient row */
-    if (step_mem->Ai_row)
+    step_mem->Ai_row = realloc(step_mem->Ai_row,
+                               step_mem->stages * sizeof(*step_mem->Ai_row));
```
drreynolds, line 1110: I'm okay with changing to `realloc`, but (a) please cast the result to the correct type (`sunrealtype*`), and (b) `memset` the new memory to all zeros to retain `calloc`'s initialization feature.

src/arkode/arkode_mristep.c
```diff
@@ -1082,42 +1078,65 @@ int mriStep_Init(void* arkode_mem, int init_type)
...
+      arkProcessError(ark_mem, ARK_MEM_FAIL, __LINE__, __func__, __FILE__,
+                      MSG_ARK_MEM_FAIL);
+      return (ARK_MEM_FAIL);
+    }
+    ark_mem->lrw += delta_stages;
+
+    /* Allocate reusable arrays for fused vector interface */
+    step_mem->nfusedopvecs = 2 * step_mem->stages + 2;
+    step_mem->cvals = realloc(step_mem->cvals,
+                              step_mem->nfusedopvecs * sizeof(*step_mem->cvals));
```
drreynolds, line 1122: I'm okay with changing to `realloc`, but (a) please cast the result to the correct type (`sunrealtype*`), and (b) `memset` the new memory to all zeros to retain `calloc`'s initialization feature.

src/arkode/arkode_mristep.c
```diff
@@ -1082,42 +1078,65 @@ int mriStep_Init(void* arkode_mem, int init_type)
...
+    if (step_mem->cvals == NULL)
+    {
+      arkProcessError(ark_mem, ARK_MEM_FAIL, __LINE__, __func__, __FILE__,
+                      MSG_ARK_MEM_FAIL);
+      return (ARK_MEM_FAIL);
+    }
+    ark_mem->lrw += 2 * delta_stages;
+
+    step_mem->Xvecs = realloc(step_mem->Xvecs,
+                              step_mem->nfusedopvecs * sizeof(*step_mem->Xvecs));
```
drreynolds, line 1132: I'm okay with changing to `realloc`, but (a) please cast the result to the correct type (`N_Vector*`), and (b) `memset` the new memory to all `NULL` to retain `calloc`'s initialization feature.

src/arkode/arkode_mristep_io.c
```diff
@@ -431,7 +431,7 @@ int MRIStepSetOrder(void* arkode_mem, int ord)
   if (retval) { return (retval); }
 
   /* check for illegal inputs */
-  if (ord < 3 || ord > 4) { step_mem->q = 3; }
+  if (ord <= 0) { step_mem->q = 3; }
```
drreynolds, line 434: What about `ord > 4`?

test/unit_tests/arkode/CXX_serial/ark_test_dahlquist_mri.cpp
```diff
@@ -418,10 +414,11 @@ int run_tests(MRISTEP_METHOD_TYPE type, sunrealtype t0, int nsteps,
 
     cout << "\nComparing Solver Statistics:\n";
 
+    int nstages_evaled = nstages_stored - stiffly_accurate;
```
drreynolds, line 417: I'm not sure what "evaled" means -- can this be changed to something clearer (e.g., "evaluated")?

drreynolds (changes requested):

src/arkode/arkode_mristep.c
```diff
@@ -93,16 +93,14 @@ void* MRIStepCreate(ARKRhsFn fse, ARKRhsFn fsi, sunrealtype t0, N_Vector y0,
   }
 
   /* Allocate ARKodeMRIStepMem structure, and initialize to zero */
-  step_mem = NULL;
-  step_mem = (ARKodeMRIStepMem)malloc(sizeof(struct ARKodeMRIStepMemRec));
+  step_mem = calloc(1, sizeof(*step_mem));
```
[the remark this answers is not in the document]
Steven-Roberts: Why is the cast necessary? I guess it's needed when using a c++ compiler. Maybe it's pedantic, but I think [no cast is preferred](https://stackoverflow.com/a/605858).

src/arkode/arkode_mristep.c
Steven-Roberts: The `calloc` is not necessary, since the `stage_map` will be filled in by `mriStepCoupling_GetStageMap`. I'll concede this could be a bit fragile if that function changes, but it's also easier to detect issues using valgrind in the current form. I don't have a strong preference, so I'll add the `memset` if you think it's necessary.

src/arkode/arkode_mristep_io.c
Steven-Roberts: This edit was done to be consistent with what's done for ARK and ERK:
https://github.com/LLNL/sundials/blob/2abd63bd6cbc354fb4861bba8e98d0b95d65e24a/src/arkode/arkode_arkstep_io.c#L1047
https://github.com/LLNL/sundials/blob/2abd63bd6cbc354fb4861bba8e98d0b95d65e24a/src/arkode/arkode_erkstep_io.c#L497

Across ARKODE, we check for illegal values, i.e. `ord > 4` during initialization. I would argue that we should fail early and raise an error in setters, not initialization. That is probably a more sweeping change that shouldn't be address in this PR, so I just made the behavior consistent across ARKODE.

src/arkode/arkode_mristep.c
Steven-Roberts: `memset` is not need for any of the `realloc`s introduced in this PR. I checked each usage and valgrind doesn't report any issues. Certainly in other parts of the code outside this PR, we may need the `memset` if we switch to `realloc`. For this PR, are you ok with leaving out the `memset`?

gardner48: I'm not sure how much of a benefit there is from `realloc` in this case since we'll overwrite the existing data in the array. If `realloc` can grow or shrink the allocation then we save the `free` call and if not then there is an unnecessary copy.

drreynolds: I recall that we've had problems in the past with using `malloc` without `memset`, but only on some compilers and platforms (I seem to recall Windows and MS Visual Studio being particularly problematic), which is why we're being careful to use `calloc` nowadays.  It's been my experience that GCC on Linux does what we "expect," so I'm not surprised that valgrind didn't report any issues.

Given that this purpose of this PR is to add things in a completely different location (new MRI tables), I would prefer that it not fiddle around with *anything* in this function.  However, your overall `calloc` -> `realloc` approach is reasonable, so I'm fine with including it in this PR, as long as we can guarantee that these changes will work on all platforms/compilers.

Steven-Roberts: Good point David, I completely forgot about `realloc` copying the old data. I'll revert back to `free`+`calloc` while keeping the fix for `cvals` and `Xvecs` reallocation. That will be consistent and safe and we can tackle the optimal allocation question later.

drreynolds (approved):

drreynolds (changes requested):
All of this looks good to me, but this is missing documentation of the new tables, as well as comments in the change log and recent changes files.

include/arkode/arkode_mristep.h
```diff
@@ -52,20 +52,35 @@ typedef enum
...
-/* Default MRI coupling tables for each order */
-
-static const int MRISTEP_DEFAULT_3         = ARKODE_MIS_KW3;
-static const int MRISTEP_DEFAULT_EXPL_3    = ARKODE_MIS_KW3;
-static const int MRISTEP_DEFAULT_EXPL_4    = ARKODE_MRI_GARK_ERK45a;
-static const int MRISTEP_DEFAULT_IMPL_SD_2 = ARKODE_MRI_GARK_IRK21a;
-static const int MRISTEP_DEFAULT_IMPL_SD_3 = ARKODE_MRI_GARK_ESDIRK34a;
-static const int MRISTEP_DEFAULT_IMPL_SD_4 = ARKODE_MRI_GARK_ESDIRK46a;
-static const int MRISTEP_DEFAULT_IMEX_SD_3 = ARKODE_IMEX_MRI_GARK3b;
-static const int MRISTEP_DEFAULT_IMEX_SD_4 = ARKODE_IMEX_MRI_GARK4;
```
gardner48, line 67: This is inconsistent with how the defaults are defined in ERKStep and ARKStep. If we're going to define the defaults in the enum we should make it consistent across the steppers.

src/arkode/arkode_mristep.c
```diff
@@ -93,16 +93,14 @@ void* MRIStepCreate(ARKRhsFn fse, ARKRhsFn fsi, sunrealtype t0, N_Vector y0,
   }
 
   /* Allocate ARKodeMRIStepMem structure, and initialize to zero */
-  step_mem = NULL;
-  step_mem = (ARKodeMRIStepMem)malloc(sizeof(struct ARKodeMRIStepMemRec));
+  step_mem = (ARKodeMRIStepMem)calloc(1, sizeof(*step_mem));
```
gardner48, line 96: Related to the discussion in #451 we should settle on the a style for `malloc`/`calloc` calls i.e., `sizeof(struct ARKodeMRIStepMemRec)` vs `sizeof(*step_mem)`.

test/unit_tests/arkode/CXX_serial/ark_test_dahlquist_mri.cpp
```diff
@@ -418,10 +414,11 @@ int run_tests(MRISTEP_METHOD_TYPE type, sunrealtype t0, int nsteps,
 
     cout << "\nComparing Solver Statistics:\n";
 
+    int nstages_evaluated = nstages_stored - stiffly_accurate;
```
gardner48, line 417: Since `stiffly_accurate` is a bool, I think the following might be better. 
```suggestion
    int nstages_evaluated = nstages_stored;
    if (stiffly_accurate) nstages_evaluated--;
```

gardner48 (changes requested):

include/arkode/arkode_mristep.h
Steven-Roberts: Fixed. I think this would be good to do across SUNDIALS in another PR. Or at least change `int` to the enum type.

Steven-Roberts:
Since `ARKODE_FORWARD_EULER_1_1` and `ARKODE_BACKWARD_EULER_1_1` can be paired as an ARK method, should I change the names to `ARKODE_ARK1_ERK_1_1` and `ARKODE_ARK1_DIRK_1_1` for consistency with other ARK methods? This seems a bit less descriptive.

Note that pair is not internally consistent. An internally consistent version is
```
0 0 | 0 0
1 0 | 0 1
---------
1 0 | 0 1
```
but I not sure the current implementation would avoid the redundant $f^I(y_n)$ evaluation and exploit the FSAL property of the explicit part.

drreynolds:
> Since `ARKODE_FORWARD_EULER_1_1` and `ARKODE_BACKWARD_EULER_1_1` can be paired as an ARK method, should I change the names to `ARKODE_ARK1_ERK_1_1` and `ARKODE_ARK1_DIRK_1_1` for consistency with other ARK methods? This seems a bit less descriptive.
> 
> Note that pair is not internally consistent. An internally consistent version is
> 
> ```
> 0 0 | 0 0
> 1 0 | 0 1
> ---------
> 1 0 | 0 1
> ```
> 
> but I not sure the current implementation would avoid the redundant fI(yn) evaluation and exploit the FSAL property of the explicit part.

Since these tables weren't explicitly derived as an ARK pair, there is no need to rename them.  However, I suppose that we could add those ARK names to the existing ones if desired.

I agree that the current ARKStep implementation would likely not avoid the redundant evaluation or exploit FSAL.  David added a "stage map" array to MRIStep that determines the algorithm structure based on the MRI coupling table at the start of a run.  I wonder if we should expand on this idea in MRIStep, and employ something similar in ARKStep, to do more to identify and leverage method structure automatically.

doc/arkode/guide/source/Butcher.rst
```diff
@@ -150,6 +150,37 @@ specified via a unique ID and name:
...
+
+.. index:: Forward-Euler-1-1 ERK method
+
+Accessible via the constant ``ARKODE_FORWARD_EULER_1_1`` to
+:c:func:`ARKStepSetTableNum`, :c:func:`ERKStepSetTableNum` or
+:c:func:`ARKodeButcherTable_LoadERK`.
+Accessible via the string ``"ARKODE_FORWARD_EULER_1_1"`` to
+:c:func:`ARKStepSetTableName`, :c:func:`ERKStepSetTableName` or
+:c:func:`ARKodeButcherTable_LoadERKByName`.
+This is the default 1st order explicit method (from :cite:p:`Euler:68`)..
```
drreynolds, line 166:
```suggestion
This is the default 1st order explicit method (from :cite:p:`Euler:68`).
```

drreynolds (approved):
One tiny typo, otherwise this looks great.

drreynolds (dismissed):
Once all of the CI tests complete, I think that this is good to merge.

Steven-Roberts:
Should be ready to merge now pending the jenkins test and any final reviews

drreynolds (approved):

doc/arkode/guide/source/Constants.rst
gardner48, line 1: Update the ERK and DIRK list above with the new methods

doc/shared/sundials.bib
```diff
@@ -1795,6 +1795,15 @@ @article{DorPri:80
   doi     = {10.1016/0771-050X(80)90013-3}
 }
 
+@book{Euler:68
+  author    = {Leonhard Euler},
+  title     = {Institutiones calculi integralis},
+  volume    = {Volumen Primum},
+  year      = {1768},
```
gardner48, line 1802: Love that we have a 250+ year old reference! Does this have a DOI? :laughing:

doc/shared/sundials.bib
```diff
@@ -1943,6 +1952,42 @@ @article{Mclachlan:92
...
+}
+
+@Article{Runge:95,
+  author    = {Runge, C.},
+  title     = {Ueber die numerische Aufl{\"o}sung von Differentialgleichungen},
+  journal   = {Mathematische Annalen},
+  volume    = {46},
+  number    = {2},
+  pages     = {167-178},
+  year      = {1895},
```
gardner48, line 1986: Only a 129 years old, so new! 😆

gardner48 (changes requested):
Some minor doc updates (I'll make the changes in a sec), otherwise this looks good.

gardner48 (approved):
Updated the constants table with the new methods and added default first order methods in ERKStep and ARKStep.

PR #445, opened 2024-03-22
Bugfix/cmake version compatibility

balos1:
Fixes the issue observed [here](https://github.com/AMReX-Codes/amrex/pull/3835).


drreynolds (approved):

gardner48:
Since we're not fully backward compatible across major versions I don't think switching to `AnyNewerVersion` is the right solution. Starting with CMake 3.19 user's can specify a version range but unfortunately this does not work as one would hope i.e., it requires an upper bound and does not work with ranges that span different major versions. I see two solutions:
* Consuming packages do not specify a version or version range and check the `SUNDIALS_VERSION` variables for compatibility after `find_package` returns
* Write our own `SUNDIALSConfigVersion.cmake` rather than use `write_basic_package_version_file`

balos1:
Yes, the CMake version range does not work as you would expect when `SameMajorVersion` is set. If its set to `AnyNewerVersion`, then the range works as you might expect. 

I don't think that this CMake feature is the right mechanism to use to inform other packages if two of our major versions are compatible or not because it is a nuanced question that really depends on the package using SUNDIALS not SUNDIALS itself. E.g., in AMReX, they have added code to make AMReX compile with SUNDIALS 6 or 7, so from their point of view both versions are compatible and they really would like to just be able to do `find_package(SUNDIALS 6.0.0)`.  I think it is fine that a user who calls `find_package(SUNDIALS 6.0.0)` and gets SUNDIALS 7 might then find out that their code doesnt compile with SUNDIALS 7.  I don't like that a package that knows that SUNDIALS 6 or 7 will work has to do something more complicated in their own build system.  

Using `AnyNewerVersion` allows for the consuming package to decide on what versions of SUNDIALS are compatible for them, so I prefer it as the solution.

gardner48:
Just specifying the minimum is easier. I think we should explain in this [section](https://sundials.readthedocs.io/en/latest/sundials/Install_link.html#using-sundials-as-a-third-party-library-in-other-cmake-projects) that starting with version `x.z.y.` the version number provided to `find_package` will be treated as the minimum required and any newer version found will be marked as compatible. If users need to limit the range, then should use the version range feature to limit the range of acceptable versions.

doc/shared/sundials/Install.rst
```diff
@@ -1761,6 +1768,13 @@ configuration file to build against SUNDIALS in their own CMake project.
   target_link_libraries(myexec PUBLIC SUNDIALS::cvode SUNDIALS::nvecpetsc)
 
 
+.. note::
+
+   Starting with SUNDIALS v7.1.0, a single version provided to ``find_package``
+   denotes the minimum version of SUNDIALS to look for, and any version equal
+   or newer than what is specified will match.
```
gardner48, line 1775:
```suggestion
   .. versionchanged:: x.y.z

      A single version provided to ``find_package`` denotes the minimum version
      of SUNDIALS to look for, and any version equal or newer than what is
      specified will match. In prior versions ``SUNDIALSConfig.cmake`` required
      the version found to have the same major version number as the single
      version provided to ``find_package``.
```

gardner48 (changes requested):
Minor doc update, otherwise this looks good.

doc/shared/sundials/Install.rst
balos1: Question about leaving this as `x.y.z`, will the version update script find these and replace them? Or do we have to notice it in the release PR?

gardner48: The version update script will find and replace them.

gardner48 (approved):

Issue #423, opened 2024-02-16
API inconsistency: some SUNDIALS "Space" functions use `sunindextype` and some use `long int`


balos1:
For example, we have

```c
N_VSpace(N_Vector x, sunindextype *lenrw, sunindextype *leniw);
SUNMatSpace(SUNMatrix A, long  int *lenrw, long int *liniw);
```

balos1:
This is sort of related to conversation in #409. We use `long` in many places where we should now switch to either `long long` or a fixed-width type like `int64_t` since we are now requiring C99.

gardner48:
The use of `sunindextype` in `N_VSpace` was an erroneous update when `sunindextype` was added. It would be good to address in 7.0 (at least for consistency).

balos1:
Looks like `sunindextype` is used for `ARKodeButcherTable_Space` too.

balos1:
This was fixed in 7.0

balos1:
Actually, I was mistaken, this was not fixed in 7.0.0

drreynolds:
This will be superseded by https://github.com/LLNL/sundials-private/issues/93.  Should this one be closed?

gardner48:
Closing in favor LLNL/sundials-private#93

PR #447, opened 2024-03-26
Enable Fortran interface for 32-bit sunindextype

balos1:
This tweaks the SWIG generation, and CMake build system to support the Fortran interfaces for 32-bit or 64-bit sunindextype as suggested in https://github.com/LLNL/sundials-private/issues/60.  Most of the files 'changed' are just the moved or new SWIG generated files.  Unfortunately this does make the repository larger in terms of files and bytes.


doc/superbuild/source/developers/releases/Checklist.rst
```diff
@@ -73,6 +73,7 @@ web pages.
    that they are deprecated).
 
 #. Regenerate the Fortran 2003 interfaces. It is possible nothing will be updated.
+   This is done by running ``make all32 all64`` in the ``swig/`` directory.
```
drreynolds, line 76: It looks like a simple `make` or even `make all` would also suffice, since you have the first explicit target 
```
all: all32 all64
```

swig/Makefile
```diff
@@ -20,7 +20,7 @@ CVODES=fcvodes_mod
 IDA=fida_mod
 IDAS=fidas_mod
 KINSOL=fkinsol_mod
-GENERIC=fsundials_core_mod
+CORE=fsundials_core_mod
```
drreynolds, line 23: In most of the `src/X/X/fmod_int32/CMakeLists.txt` and `src/X/X/fmod_int64/CMakeLists.txt` files you have the line
```
    PUBLIC sundials_fcore_mod
```

Should all of those instead be
```
    PUBLIC fsundials_core_mod
```
?

swig/README.md
```diff
@@ -52,8 +52,8 @@ $ make install # optional
 ## How to regenerate the interfaces
 
 To regenerate the interfaces that have already been created. Simply run
-`make` from the `sunrepo/swig` directory. **This will replace all the
-generated files in `sunrepo/src`.**
+`make all32 all64` from the `sundials/swig` directory.
```
drreynolds, line 55: Alternately, could this just be `make` or `make all`?

swig/sunmatrix/fsunmatrix_sparse_mod.i
drreynolds, line 1: I recommend that we take one of the sparse KLU F2003 examples and set it up either to use 32-bit integers, or to work with both 32-bit and 64-bit integers, so that we can test the new support (and show an example).

drreynolds (commented):
I still need to test this out, but for now I have only a few comments/questions.

doc/superbuild/source/developers/releases/Checklist.rst
balos1: I thought so too but I could not get it to work. If you happen to have an idea on how to fix it, it would be appreciated.

balos1: When you run `make`  or `make all` it only runs the all32 target.

swig/Makefile
balos1: No. The latter is the name of the module but  the former is the name of the CMake target / library (confusing - yes).

swig/README.md
balos1: See my other comment. For some reason `make` / `make all` only run the `all32` target. I have tried many different tweaks to the makefile with no luck.

balos1:
In a few functions I promoted `int` pertaining to vector arrays to `sunindextype` to maintain some consistency across these functions. Since its a promotion, it is backwards compatible despite being a change in the function signatures.

balos1:
Related to #423.

swig/README.md
drreynolds: I figured out how to do this, but it's not as clean.  The issue is that `make` seems to determine the dependency tree based on the target names, without realizing that your `.SETINT32` and `.SETINT64` changes those targets.  The way I got this to work was by duplicating each target name, e.g., instead of one `core` target I have two, `core32` and `core64`.  I then group all of the `32` targets together into `modules32`, and all of the `64` targets together into `modules64`, which are used to create `all32` and `all64`.

While my approach will build both sets of interfaces with a single `make` command, it nearly doubles the size of the `Makefile`.  Moreover, it could be confusing when people add new modules, and is not extensible if we wanted to include more variations (e.g., `single`, `double` and `long double` for `sunrealtype`).  I recommend that we stick with @balos1's current approach, even if it requires `make all32 all64`.

examples/idas/F2003_serial/idasAkzoNob_ASAi_dns_f2003.f90
```diff
@@ -33,8 +33,19 @@ module dae_mod
...
+
+  ! Since SUNDIALS can be compiled with 32-bit or 64-bit sunindextype
+  ! we set the integer kind used for indices in this example based
+  ! on the the index size SUNDIALS was compiled with so that it works
+  ! in both configurations. This is not a requirement for user codes.
+#if defined(SUNDIALS_INT32_T)
+  integer, parameter :: sunindextype = selected_int_kind(8)
+#elif defined(SUNDIALS_INT64_T)
+  integer, parameter :: sunindextype = selected_int_kind(16)
+#endif
```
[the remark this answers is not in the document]
balos1: I suppose we could put it in there, but users should not use it anyways. We need it in our examples so we can compile them with both options for testing.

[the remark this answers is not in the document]
drreynolds: I suppose that's true, since we also tell users not to worry about using `sunindextype` or `sunrealtype` in their codes, just to install SUNDIALS in a compatible manner to their built-in types.

drreynolds (dismissed):

examples/idas/F2003_serial/idasAkzoNob_ASAi_dns_f2003.f90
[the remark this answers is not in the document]
gardner48: Would having `sunindextype` in `fsundials_core_mod` allow us to have one set of sources for `int32` and `int64` interfaces? 

In this example, maybe we should change the name to something like `myindextype` so the "user" isn't redefining a SUNDIALS type.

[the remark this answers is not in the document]
balos1: > Would having sunindextype in fsundials_core_mod allow us to have one set of sources for int32 and int64 interfaces?

No, because SWIG still does not understand it.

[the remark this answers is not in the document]
balos1: > In this example, maybe we should change the name to something like myindextype so the "user" isn't redefining a SUNDIALS type.

👍

examples/cvode/F2003_parallel/cv_diag_kry_f2003.out
balos1, line 1: There was a typo in the CMakeLists.txt that prevented this example from ever running before. As such, I am not sure how the original `.out` file was even generated and why it is so far off from the one generated on the test box. I assume it must have been generated on a different machine.

examples/kinsol/F2003_parallel/kin_diagon_kry_f2003.out
balos1, line 1: There was a typo in the CMakeLists.txt that prevented this example from ever running before.

examples/ida/F2003_parallel/ida_heat2D_kry_bbd_f2003.out
balos1, line 1: There was a typo in the CMakeLists.txt that prevented this example from ever running before.

balos1:
@drreynolds @gardner48 All checks are green, yay! I think this is ready now.

drreynolds (changes requested):
One tiny item, otherwise this looks great to me.

drreynolds (dismissed):

examples/nvector/C_openmp/test_fnvector_openmp_mod.f90
```diff
@@ -22,16 +22,16 @@ module test_nvector_openmp
   use test_utilities
   implicit none
 
-  integer(c_long), parameter :: N = 100 ! vector length
-  integer(c_int),  parameter :: nv = 3  ! length of vector arrays
-  integer(c_int),  parameter :: ns = 2  ! number of vector arrays
+  integer(kind=myindextype), parameter :: N = 100 ! vector length
+  integer(kind=myindextype), parameter :: nv = 3  ! length of vector arrays
+  integer(kind=myindextype), parameter :: ns = 2  ! number of vector arrays
```
gardner48, line 27: Should these be `c_int` since they're converged to `int(4)` when used below?

examples/nvector/mpimanyvector/test_fnvector_mpimanyvector_mod.f90
```diff
@@ -25,37 +25,37 @@ module test_nvector_mpimanyvector
...
+    real(c_double)            :: nvarr(nv)              ! array of nv constants to go with vector array
+    type(N_Vector), pointer   :: x, y, z, tmp           ! N_Vectors
+    type(c_ptr)               :: subvecs                ! MPIManyVector subvectors
+    type(c_ptr)               :: xvecs, zvecs           ! C pointer to array of MPIManyVectors
 
     !===== Setup ====
     subvecs = FN_VNewVectorArray(nsubvecs, sunctx)
     tmp  => FN_VMake_Serial(N1, x1data, sunctx)
-    call FN_VSetVecAtIndexVectorArray(subvecs, 0, tmp)
+    call FN_VSetVecAtIndexVectorArray(subvecs, 0_myindextype, tmp)
```
gardner48, line 54: I guess we were inconsistent with how `VectroArray`s are index (`int`) and how subvectors are indexed (`sunindextype`). These should probably be the same and I guess it's easier to promote `int` to `sunindextype` even if we'd probably never need even max int numbers of subvectors.

examples/nvector/test_nvector.f90
```diff
@@ -18,12 +18,10 @@
 ! to an NVECTOR module implementation. It does not test every
 ! function. It tests the N_VMake constructor, one standard vector
 ! operation (N_VConst), N_VGetArrayPointer, and one fused operation.
-! -----------------------------------------------------------------
+! -----------------------------------------------------------------c
```
gardner48, line 21:
```suggestion
! -----------------------------------------------------------------
```

gardner48 (changes requested):

doc/shared/nvectors/NVector_Description.rst
```diff
@@ -185,7 +185,7 @@ The functions :c:func:`N_VCloneVectorArray` and
 variables of type :c:type:`N_Vector`, each of the same type as an existing
 ``N_Vector`` input:
 
-.. c:function:: N_Vector *N_VCloneVectorArray(int count, N_Vector w)
+.. c:function:: N_Vector *N_VCloneVectorArray(sunindextype count, N_Vector w)
```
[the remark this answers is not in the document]
drreynolds: I must have missed these N_Vector API changes when I reviewed earlier, but I am against changing the `count` type in this and the other "VectorArray" routines.  While these are optional, many of our users and collaborators have created custom N_Vector implementations, and I would see this change as potentially breaking.

cmake/tpl/FindKLU.cmake
```diff
@@ -33,7 +33,11 @@
 if (NOT (KLU_INCLUDE_DIR OR KLU_LIBRARY_DIR OR KLU_LIBRARY))
   # Prefer the import target from upstream SuiteSparse if it is available
   # and the user didn't point to a specific (different) version.
+  find_package(AMD CONFIG)
+  find_package(BTF CONFIG)
+  find_package(COLAMD CONFIG)
   find_package(KLU CONFIG)
+  find_package(SuiteSparse_config CONFIG)
```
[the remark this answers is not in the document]
balos1: Right, if you set `CMAKE_MODULE_PATH` it will find them, but if you don't you need these other lines.

examples/arkode/F2003_custom/CMakeLists.txt
```diff
@@ -17,18 +17,22 @@
 # Example lists are tuples "name\;type" where the type is
 # 'develop' for examples excluded from 'make test' in releases
 
-# Examples using SUNDIALS linear solvers
-set(FARKODE_examples
-  "ark_brusselator1D_f2003\;develop"
-  "ark_analytic_complex_f2003\;develop"
-  )
+if(SUNDIALS_INDEX_SIZE MATCHES "64")
```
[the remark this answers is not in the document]
balos1: Its probably good to have some examples that are 64-bit only just for demonstration purposes, but I imagine in the future we could update more of them to work with both sizes.

cmake/tpl/FindKLU.cmake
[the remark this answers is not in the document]
mmuetzel: Could you please show in which case you need to add these lines. It definitely wasn't intended that downstream projects would need to manually include dependencies of a SuiteSparse library.
Is this an issue with the CMake Config files installed by SuiteSparse?

doc/shared/nvectors/NVector_Description.rst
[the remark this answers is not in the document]
balos1: I was able to change this back with modifications to the examples.

drreynolds (dismissed):

balos1:
@gardner48  I think this is ready to merge, if you approve.

cmake/tpl/FindKLU.cmake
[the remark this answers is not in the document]
gardner48: @mmuetzel The issue seems to be with using `-DKLU_DIR=<suite_sparse_install_path>/lib64/cmake/KLU` alone to provide the config file location. Using either `-DKLU_ROOT=<suite_sparse_install_path>` or `-DCMAKE_PREFIX_PATH=<suite_sparse_install_path>` works without the above additions.

@balos1 are you also setting `-D<dependency>_DIR` variables for the other packages? Using `-DKLU_DIR` alone didn't work for me with the above additions. I had add `-D<dependency>_DIR` variables for each dependency which also make the config work without the above additions. 

This issue is tangential to the topic of this PR so I suggest these changes get broken out into a separate PR.

doc/shared/RecentChanges.rst
```diff
@@ -93,6 +93,8 @@ Fixed a memory leak when an error handler was added to a
 :c:type:`SUNContext`. Fixes `GitHub Issue #466
 <https://github.com/LLNL/sundials/issues/466>`_.
 
+Enabled the Fortran interfaces to build with 32-bit ``sunindextype``.
```
gardner48, line 96: Move to "New Features and Enhancements" section above

doc/shared/nvectors/NVector_Description.rst
gardner48, line 1: Revert changes in this file

doc/superbuild/source/developers/style_guide/SourceCode.rst
```diff
@@ -338,6 +341,17 @@ Coding Conventions and Rules
    x;`` to ``return(x);``. Note, however, lots of older SUNDIALS source code
    uses ``return(x);``.
 
+#. Always use ``sunindextype`` for variables that are related to problem dimensions.
+   E.g., use it for the length of a vector, dimensions of a matrix, or for the number of vectors.
```
gardner48, line 345: I don't think we've settled on this yet so let's leave it out of the conventions until we do
```suggestion
#. Always use ``sunindextype`` for variables that are related to problem dimensions
   e.g., use it for the length of a vector or dimensions of a matrix.
```

examples/arkode/F2003_parallel/CMakeLists.txt
```diff
@@ -17,13 +17,16 @@
...
-    "ark_brusselator1D_task_local_nls_f2003\;--monitor --global-nls\;1\;4\;develop\;2"
-    "ark_brusselator1D_task_local_nls_f2003\;--monitor --explicit --tf 3\;1\;4\;develop\;2"
-    "ark_diag_kry_bbd_f2003\;\;1\;4\;develop\;2"
-    "ark_diag_non_f2003\;\;1\;4\;develop\;2"
-    "ark_heat2D_f2003\;\;1\;4\;develop\;2")
+
+if(SUNDIALS_INDEX_SIZE MATCHES "64")
+  set(FARKODE_examples
+      "ark_brusselator1D_task_local_nls_f2003\;--monitor\;1\;4\;develop\;2"
+      "ark_brusselator1D_task_local_nls_f2003\;--monitor --global-nls\;1\;4\;develop\;2"
```
gardner48, line 24: Looks like this example was updated to work with 32-bit indices should they be outside the "if 64"?

examples/cvode/F2003_serial/CMakeLists.txt
```diff
@@ -21,33 +21,33 @@
...
-  "cv_roberts_dns_f2003\;develop"
   )
 
-# Examples using LAPACK linear solvers
 if(SUNDIALS_INDEX_SIZE MATCHES "64")
 
+  # Examples using SUNDIALS linear solvers
+  set(FCVODE_examples
+    "cv_advdiff_bnd_f2003\;develop"
+    "cv_analytic_fp_f2003\;develop"
```
gardner48, line 31:
```suggestion
  list(APPEND FCVODE_examples
    "cv_analytic_fp_f2003\;develop"
```

examples/cvodes/F2003_serial/CMakeLists.txt
```diff
@@ -16,13 +16,15 @@
...
-  "cvs_analytic_fp_f2003\;\;develop"
-  "cvsAdvDiff_FSA_non_f2003\;-sensi stg t\;develop"
-  "cvsAdvDiff_FSA_non_f2003\;-sensi sim t\;develop"
-  )
+  "cvs_analytic_fp_f2003\;\;develop")
+
+if(SUNDIALS_INDEX_SIZE MATCHES "64")
+  list(APPEND FCVODES_examples
+    "cvsAdvDiff_FSA_non_f2003\;-sensi stg t\;develop"
+    "cvsAdvDiff_FSA_non_f2003\;-sensi sim t\;develop"
```
gardner48, line 25: Looks like these example were updated to work with 32- and 64-bit ints too

gardner48 (changes requested):

balos1:
@gardner48 All checks passing 😄

gardner48 (approved):

cmake/tpl/FindKLU.cmake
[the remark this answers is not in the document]
mmuetzel: It looks like the changes in this file made it into the final version that was merged.

Should the changes to this file be reverted?
(Iiuc, it is unclear which issue they are addressing.)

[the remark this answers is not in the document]
gardner48: @balos1 in my testing it seemed like these updates were not needed could you check if setting `-D<dependency>_DIR` without these changes is sufficient?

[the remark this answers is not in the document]
balos1: I cant replicate what I had seen before, so I think we can roll this back in a separate PR.

Issue #443, opened 2024-03-21
Update SUNDIALS source tree figure in documentation


drreynolds:
Figure 1.2, Directory structure of the SUNDIALS source tree (at https://sundials.readthedocs.io/en/latest/Organization_link.html) still shows boxes for the old Fortran77 interfaces -- these should be removed.

PR #450, opened 2024-03-27
Docs: Updating landing page, remove file tree figure

gardner48:
* Move high level description of SUNDIALS to docs landing page
* Remove file tree figure from docs (addresses #443)
* Remove organization section in shared docs


drreynolds (dismissed):

doc/superbuild/source/Organization_link.rst
balos1, line 1: Removing this makes it so that the overall SUNDIALS organization figure `sunorg1.png` is not included anywhere in the superbuild for the HTML docs (only in the PDF docs for the individual packages).   You would need to remove the `ifconfig` [here](https://github.com/LLNL/sundials/blob/main/doc/arkode/guide/source/Organization.rst?plain=1#L21) (and in the corresponding places in the other package docs), but then the organization figure would be repeated for each package in the HTML build. So maybe we should keep this file? If not, you need to remove it from `doc/superbuild/source/index.rst`.

balos1 (changes requested):

doc/superbuild/source/Organization_link.rst
gardner48: I've moved `sunorg1.png` so it appears at the start of the "Using SUNDIALS" chapter in all versions.

doc/arkode/guide/source/sundials/index.rst
```diff
@@ -16,11 +16,15 @@
 Using SUNDIALS
 **************
 
-As discussed in :numref:`ARKODE.Organization`, the six solvers packages
-(CVODE(S), IDA(S), ARKODE, KINSOL) that make up SUNDIALS are built upon common
-classes/modules for vectors, matrices, and algebraic solvers. In addition, the
-six packages all leverage some other common infrastructure, which we discuss
-in this section.
+.. _Organization.Sundials.HighLevelDiagram:
```
balos1, line 19: After previewing the docs, I think it might be better to have the figure at the bottom of the page.

balos1 (changes requested):

balos1 (approved):

PR #451, opened 2024-03-27
Split hashmap into .c and .h file

Steven-Roberts:
This fixes compilation warnings I had with `gcc` 8.5 using `-Wall`. Many of the static functions in `sundials_hashmap.h` were unused by files that included the header. After discussion with @balos1 moving the implementations into a .c file made more sense than adding the `inline` keyword to all the functions.


src/sundials/sundials_hashmap_impl.h
```diff
@@ -0,0 +1,56 @@
...
+ * the hash map upon its destruction.
+ * -----------------------------------------------------------------*/
+
+#ifndef _SUNDIALS_HASHMAP_IMPL_H
+#define _SUNDIALS_HASHMAP_IMPL_H
+
+#include <stdlib.h>
+#include <sundials/sundials_types.h>
+
+typedef struct _SUNHashMapKeyValue* SUNHashMapKeyValue;
```
gardner48, line 26:
```suggestion
typedef struct SUNHashMapKeyValue_* SUNHashMapKeyValue;
```

src/sundials/sundials_hashmap_impl.h
```diff
@@ -0,0 +1,56 @@
...
+
+struct _SUNHashMapKeyValue
```
gardner48, line 28:
```suggestion
struct SUNHashMapKeyValue_
```

src/sundials/sundials_hashmap_impl.h
```diff
@@ -0,0 +1,56 @@
...
+{
+  const char* key;
+  void* value;
+};
+
+typedef struct _SUNHashMap* SUNHashMap;
```
gardner48, line 34:
```suggestion
typedef struct SUNHashMap* SUNHashMap_;
```

src/sundials/sundials_hashmap_impl.h
```diff
@@ -0,0 +1,56 @@
...
+
+struct _SUNHashMap
```
gardner48, line 36:
```suggestion
struct SUNHashMap_
```

src/sundials/sundials_hashmap.c
```diff
@@ -17,15 +17,13 @@
  * the hash map upon its destruction.
  * -----------------------------------------------------------------*/
```
gardner48, line 18: Remove "header-only" in the comment block above.

src/sundials/sundials_hashmap_impl.h
```diff
@@ -0,0 +1,56 @@
...
+ * See the top-level LICENSE and NOTICE files for details.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ * SUNDIALS Copyright End
+ * -----------------------------------------------------------------
+ * A simple header-only hashmap implementation for char* keys and
+ * void* values. Uses linear probing to resolve collisions.
+ * The values can be anything, but will be freed by
+ * the hash map upon its destruction.
+ * -----------------------------------------------------------------*/
```
gardner48, line 18: Remove "header-only" in the comment block above

gardner48 (changes requested):
Minor fixes to comments and avoid C/C++ reserved naming patters

src/sundials/sundials_hashmap.c
```diff
@@ -74,14 +55,14 @@ struct _SUNHashMap
...
-static SUNErrCode SUNHashMap_New(int max_size, SUNHashMap* map)
+SUNErrCode SUNHashMap_New(int max_size, SUNHashMap* map)
 {
   int i;
 
   if (max_size <= 0) { return SUN_ERR_ARG_OUTOFRANGE; }
 
   *map = NULL;
-  *map = (SUNHashMap)malloc(sizeof(struct _SUNHashMap));
+  *map = (SUNHashMap)malloc(sizeof(**map));
```
balos1, line 65: Personally, I find this to be less readable in this case.  I like to do `sizeof(type)` in these contexts because IMO it more clearly expresses the intent to allocate memory for `type`.

src/sundials/sundials_hashmap.c
```diff
@@ -224,7 +203,7 @@ static int SUNHashMap_Insert(SUNHashMap map, const char* key, void* value)
   }
 
   /* Create the key-value pair */
-  kvp = (SUNHashMapKeyValue)malloc(sizeof(struct _SUNHashMapKeyValue));
+  kvp = (SUNHashMapKeyValue)malloc(sizeof(*kvp));
```
balos1, line 206: Again, I find the former code to be more readable.

src/sundials/sundials_hashmap.c
Steven-Roberts: The `sizeof(type)` can be fragile since it's easy to forget to change that if the type of `map` is changed. Also it makes more work when renaming a type like this. https://stackoverflow.com/q/17258647

balos1: Yeah, in general, but why would we ever change the type of `SUNHashMap` in the malloc constructor for `SUNHashMap`?

balos1: Generally, I understand the point that `sizeof(expr)` would be less fragile to a chnage of the type, but that does not seem likely to ever happen here.

balos1: It looks like the rest of sundials is a bit of a mixed bag on this stylistic point, so there isnt really a precedent.

Steven-Roberts: > Yeah, in general, but why would we ever change the type of SUNHashMap in the malloc inside the constructor for SUNHashMap?

Yeah, I can't imagine that here. Certainly other places, types could change and cause issues, e.g., switching `long int` to `sunindextype`.

> Generally, I understand the point that sizeof(expr) would be less fragile to a chnage of the type, but that does not seem likely to ever happen here. Furthermore, if it did happen, we would also have to still update the cast.

Yes, but the cast is unnecessary and can be removed.

> It looks like the rest of sundials is a bit of a mixed bag on this stylistic point, so there isnt really a precedent.

I'll go with `sizeof(type)` for now, and we can discuss how to do this more consistently at our next meeting. My 2 cents is we should use `data = malloc(sizeof(*data))` everywhere.

gardner48: In the past we've had some users that pull in the sundials source and build it as part of their C++ application with a C++ compiler and while we don't endorse this use case the cast is required there.

balos1: Yeah, I think the cast needs to stay for that reason. As far as the `sizeof`, it doesn't really matter but perhaps we can choose a convention to stick with.

balos1: After thinking about it a bit more, always using `data = malloc(sizeof(*data))` might make any future scripted change to use an allocator easier.

gardner48: Scripting-wise either style is fine as `data = malloc(sizeof(...)` will become something like `sunMalloc(sizeof(...))`

gardner48 (approved):

src/sundials/sundials_hashmap.c
gardner48: I slightly prefer the more verbose style
```
*map = (SUNHashMap)malloc(sizeof(struct SUNHashMap_));
```
to the alternative
```
*map = (SUNHashMap)malloc(sizeof(**map));
```
however now that C99 is allowed, something like
```
SUNHashMap map_out = (SUNHashMap)malloc(sizeof(*map));
```
could also work.

balos1:
This looks good to me, but the formatter needs to be applied so that the clang check passes.

src/sundials/sundials_hashmap.c
Steven-Roberts: Ok (2) it is. When we upgrade to `sunMalloc` we could do something like this to avoid casts

```
#ifdef __cplusplus
template<class T>
T* sunMalloc(size_t size, T* data) { return (T*) malloc(size * sizeof(T)); }
#else
void *sunMalloc(size_t size, void* data) { return malloc(size * sizeof(*data)); }
#endif

sunrealtype *data = sunMalloc(10, data);
```

Steven-Roberts: Hmm, `sizeof(*data)` is not right, but the point stands that we can probably split up the C/C++ implementation to avoid casts.

gardner48: We don't explicitly support building with a C++ compiler so I don't think we should worry about splitting this up into C and C++ versions.

balos1 (approved):

PR #456, opened 2024-04-16
Make MRI step fail when inner stepper has a recoverable error

Steven-Roberts:
Fixes a bug where an MRI method would use the (likely corrupted) result of an inner stepper which returned with a recoverable error. Once adaptivity is supported, this could be handled similarly to ARK methods.


gardner48 (changes requested):
Looks good, please update the change logs to note the fix

gardner48 (approved):

drreynolds (approved):

gardner48 (approved):

PR #457, opened 2024-04-17
Feature: Option to disable interpolation in ARKODE

gardner48:
Add support for disabling interpolated output in ARKODE steppers by passing `ARK_INTERP_NONE` to the `SetInterpolantType` functions. Disabling interpolation can reduce the memory footprint of an integrator by two or more state vectors (depending on the interpolation type and degree) which can be beneficial when interpolation is not needed e.g., when integrating to a final time without output in between or using an explicit fast time scale integrator with MRIStep.

When interpolation is disabled, rootfinding is not supported, implicit methods must use the trivial predictor, interpolation at stop times cannot be used, and calls to `Evolve` in `ARK_NORMAL` mode will return at or past the requested output time.


gardner48:
Replaced by #474

PR #375, opened 2023-11-16
Add ARKODE unit test for number of mass matrix solves

Steven-Roberts:
Adds a unit test that will allow us to monitor changes to and ensure the correct number of mass matrix solves in ARKODE. I've included comments in each test which explain where the solves arise and if it can be sharpened. The tests cover a range of methods and settings, but let me know if there other combinations worth including. Here is the test output: 

```
Mass Matrix Solve Count Test

Result | Implicit Method               | Explicit Method               | Time-Dep | Deduce Im RHS | Actual | Expected
-------+-------------------------------+-------------------------------+----------+---------------+--------+---------
  Pass | ARKODE_SDIRK_2_1_2            | ARKODE_ERK_NONE               | No       | No            | 2      | 2
  Pass | ARKODE_SDIRK_2_1_2            | ARKODE_ERK_NONE               | No       | Yes           | 2      | 2
  Pass | ARKODE_SDIRK_2_1_2            | ARKODE_ERK_NONE               | Yes      | No            | 2      | 2
  Pass | ARKODE_SDIRK_2_1_2            | ARKODE_ERK_NONE               | Yes      | Yes           | 0      | 0
  Pass | ARKODE_SDIRK_5_3_4            | ARKODE_ERK_NONE               | No       | No            | 1      | 1
  Pass | ARKODE_SDIRK_5_3_4            | ARKODE_ERK_NONE               | No       | Yes           | 1      | 1
  Pass | ARKODE_SDIRK_5_3_4            | ARKODE_ERK_NONE               | Yes      | No            | 5      | 5
  Pass | ARKODE_SDIRK_5_3_4            | ARKODE_ERK_NONE               | Yes      | Yes           | 0      | 0
  Pass | ARKODE_ESDIRK324L2SA_4_2_3    | ARKODE_ERK_NONE               | No       | No            | 2      | 2
  Pass | ARKODE_ESDIRK324L2SA_4_2_3    | ARKODE_ERK_NONE               | No       | Yes           | 2      | 2
  Pass | ARKODE_ESDIRK324L2SA_4_2_3    | ARKODE_ERK_NONE               | Yes      | No            | 4      | 4
  Pass | ARKODE_ESDIRK324L2SA_4_2_3    | ARKODE_ERK_NONE               | Yes      | Yes           | 1      | 1
  Pass | ARKODE_DIRK_NONE              | ARKODE_SHU_OSHER_3_2_3        | No       | No            | 5      | 5
[...]

PR #458, opened 2024-04-18
Add ARKODE unit test for number of mass matrix solves

Steven-Roberts:
Replaces #375 which has become too outdated to easily merge develop into.


balos1 (approved):

balos1:
@drreynolds Are you OK if we merge this PR before #459  since it only adds a new unit test?

drreynolds:
> @drreynolds Are you OK if we merge this PR before #459 since it only adds a new unit test?

That should be fine, since it will be just one file where I will need to follow on and make multiple edits.

PR #463, opened 2024-04-29
SUNStepper basics based on MRIStepInnerStepper

balos1:
I am opening this PR as a draft to gather feedback. This renames/moves `MRIStepInnerStepper` to `SUNStepper` so that @Steven-Roberts and I can extend it for other purposes.


src/arkode/arkode.c
```diff
@@ -3318,6 +3318,90 @@ int arkAccessHAdaptMem(void* arkode_mem, const char* fname, ARKodeMem* ark_mem,
   return (ARK_SUCCESS);
 }
 
+/* Allocate MRI forcing and fused op workspace vectors if necessary */
+int arkAllocSUNStepperForcing(SUNStepper stepper, int count, N_Vector tmpl)
```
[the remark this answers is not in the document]
balos1: With `MRIStepInnerStepper`, we were using `arkAllocVectors` to allocate the vectors. I was unsure yet if it was important that `arkAllocVectors` was used and it is on my todo to determine if it is.  If we dont need to use `arkAllocVectors` then these routines can be included as `SUNStepper` functions.

src/sundials/sundials_stepper.c
```diff
@@ -0,0 +1,173 @@
...
+  memset(*stepper, 0, sizeof(**stepper));
+
+  (*stepper)->ops = (SUNStepper_Ops)malloc(sizeof(*((*stepper)->ops)));
+  SUNAssert((*stepper)->ops, SUN_ERR_MALLOC_FAIL);
+
+  memset((*stepper)->ops, 0, sizeof(*((*stepper)->ops)));
+
+  (*stepper)->priv_ops =
+    (SUNStepper_PrivOps)malloc(sizeof(*((*stepper)->priv_ops)));
+  SUNAssert((*stepper)->priv_ops, SUN_ERR_MALLOC_FAIL);
```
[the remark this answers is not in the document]
balos1: Just separation of private ops and non-private. Not really necessary and will go away if we do replace the alloc/free forcing ops with SUNStepper routines.

balos1:
@drreynolds @gardner48 I have made the changes we discussed on Tuesday. Let me know what you think.

src/arkode/arkode_arkstep.c
```diff
@@ -21,13 +21,19 @@
...
 #include "arkode_impl.h"
 #include "arkode_interp_impl.h"
 
 #define FIXED_LIN_TOL
 
+/* TryStep step result flags */
+#define TRYSTEP_FAILED  -1
+#define TRYSTEP_SUCCESS +0
+#define TRYSTEP_ADAPT   +1
+
```
drreynolds, line 36: Should these be in `include/sundials/sundials_stepper.h` instead?

src/arkode/xbraid/arkode_xbraid.c
```diff
@@ -438,51 +438,5 @@ int ARKBraid_Access(braid_App app, braid_Vector u, braid_AccessStatus astatus)
...
-  if (tmp_flag > 0)
-  {
-    *ark_flag = STEP_ADAPT;
-    return ARK_SUCCESS;
-  }
-
-  /* Step was successful and passed the error test */
-  *ark_flag = STEP_SUCCESS;
-  return ARK_SUCCESS;
+  return arkStep_TryStep(arkode_mem, tstart, tstop, y, ark_flag);
```
drreynolds, line 441: This is a nice simplification!

src/sundials/sundials_stepper.c
```diff
@@ -0,0 +1,179 @@
...
+    stepper->fused_vectors = N_VNewVectorArray(stepper->nforcing, SUNCTX_);
+    SUNCheckLastErr();
+  }
+
+  if (!stepper->fused_scalars)
+  {
+    stepper->fused_scalars =
+      (sunrealtype*)calloc(count + 1, sizeof(*stepper->fused_scalars));
+    SUNAssert(stepper->fused_scalars, SUN_ERR_MEM_FAIL);
+  }
```
drreynolds, line 133: Shouldn't both of these also check to see whether the length of the already-allocated `fused_scalars` and `fused_vectors` arrays have length at least `count`?

drreynolds (commented):
I like how clean this is now.  I noticed a few items for comment (see below).

src/arkode/arkode_arkstep.c
```diff
@@ -3654,6 +3858,63 @@ int arkStep_GetOrder(ARKodeMem ark_mem)
   return step_mem->q;
 }
 
+/* -----------------------------------------------------------------------------
+ * arkStep_TryStep
+ *
+ * Attempts one internal time step using ARKStepEvolve
+ * ---------------------------------------------------------------------------*/
+
+int arkStep_TryStep(void* arkode_mem, sunrealtype tstart, sunrealtype tstop,
```
gardner48, line 3867: The body of this function was taken from `ARKBraid_TakeStep` and some of the behavior is specific to XBraid's use (having to reset before every step and use a specific step size) and I'm not sure if it's how we want to `SUNStepper` to behave. This is related to my question above about having a stepper advance from its last reached state.

1. If `TryStep` is called in a loop (i.e., `SUNStepper` could provide an implementation of `Advance` that calls `TryStep` as needed) the call to `Reset` could lead to unnecessary RHS evaluations.
2. Do we want the step size to be determined by `tstart` and `tstop` or internally by the stepper?

src/sundials/sundials_stepper.c
```diff
@@ -0,0 +1,214 @@
...
+
+SUNErrCode SUNStepper_Create(SUNContext sunctx, SUNStepper* stepper)
+{
+  SUNFunctionBegin(sunctx);
+
+  *stepper = NULL;
+  *stepper = (SUNStepper)malloc(sizeof(**stepper));
+  SUNAssert(stepper, SUN_ERR_MALLOC_FAIL);
+
+  memset(*stepper, 0, sizeof(**stepper));
```
gardner48, line 20: Could make the `malloc` above a `calloc` and remove the `memset`

src/sundials/sundials_stepper.c
```diff
@@ -0,0 +1,214 @@
...
+
+  (*stepper)->ops = (SUNStepper_Ops)malloc(sizeof(*((*stepper)->ops)));
+  SUNAssert((*stepper)->ops, SUN_ERR_MALLOC_FAIL);
+
+  memset((*stepper)->ops, 0, sizeof(*((*stepper)->ops)));
```
gardner48, line 25: Could make the `malloc` above a `calloc` and remove the `memset`

src/sundials/sundials_stepper.c
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  free(stepper->ops);
+  free(stepper);
+  *stepper_ptr = NULL;
+
+  return SUN_SUCCESS;
+}
+
+SUNErrCode SUNStepper_Advance(SUNStepper stepper, sunrealtype t0,
```
gardner48, line 55: If `TryStep` advances from the last time reached and if we add a `SetStopTime` function, there could be a default implementation of `Advance` that calls `TryStep`

src/sundials/sundials_stepper.c
```diff
@@ -0,0 +1,214 @@
...
+{
+  SUNFunctionBegin(stepper->sunctx);
+  if (stepper->ops->advance)
+  {
+    return stepper->ops->advance(stepper, t0, tout, y, tret, stop_reason);
+  }
+  return SUN_ERR_NOT_IMPLEMENTED;
+}
+
+SUNErrCode SUNStepper_Step(SUNStepper stepper, sunrealtype t0, sunrealtype tout,
```
gardner48, line 67: Like `Advance` we could have a default implementation that calls `TryStep`

include/sundials/sundials_stepper.h
```diff
@@ -0,0 +1,94 @@
...
+                                   sunrealtype tout, N_Vector y,
+                                   sunrealtype* tret, int* stop_reason);
+
+typedef int (*SUNStepperOneStepFn)(SUNStepper stepper, sunrealtype t0,
+                                   sunrealtype tout, N_Vector y,
+                                   sunrealtype* tret, int* stop_reason);
+
+typedef int (*SUNStepperTryStepFn)(SUNStepper stepper, sunrealtype t0,
+                                   sunrealtype tout, N_Vector y,
+                                   sunrealtype* tret, int* stop_reason);
```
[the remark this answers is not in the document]
balos1: How would someone set `t0` to begin with? Are you assuming this would be done with the package create/init routines before it is wrapped as a stepper?

include/sundials/sundials_stepper.h
```diff
@@ -0,0 +1,94 @@
...
+#define _SUNDIALS_STEPPER_H
+
+#include <sundials/sundials_core.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef _SUNDIALS_STRUCT_ SUNStepper_s* SUNStepper;
+
```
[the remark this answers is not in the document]
Steven-Roberts: I also like 3.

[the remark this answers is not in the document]
balos1: Yes, or a root return.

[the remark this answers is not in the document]
balos1: I like option 3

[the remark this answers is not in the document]
gardner48: Do you need to differentiate between between a tstop return and a root found return or is a positive return sufficient to indicate a successful return that's different from a normal successful (zero) return?

src/arkode/arkode_arkstep.c
```diff
@@ -3219,6 +3226,203 @@ int arkStep_ComputeSolutions_MassFixed(ARKodeMem ark_mem, sunrealtype* dsmPtr)
...
+  retval = SUNStepper_GetForcingData(stepper, &tshift, &tscale, &forcing,
+                                     &nforcing);
+  if (retval != ARK_SUCCESS) { return (retval); }
+
+  /* set the inner forcing data */
+  retval = arkStep_SetInnerForcing(arkode_mem, tshift, tscale, forcing, nforcing);
+  if (retval != ARK_SUCCESS) { return (retval); }
+
+  /* set the stop time */
+  retval = ARKStepSetStopTime(arkode_mem, tout);
```
Steven-Roberts, line 3297: Is this needed? Maybe it could be an option. For operator splitting, it would be preferable not to have this.

include/sundials/sundials_stepper.h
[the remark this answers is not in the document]
gardner48: I would expect this to be done with the stepper `Create` or `Init` function or alternatively `Reset` could be called in the first step if we don't require a stepper to be initialized with the starting state.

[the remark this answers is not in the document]
balos1: > I would expect this to be done with the stepper Create or Init

Do you mean the package "steppers"? The `SUNStepper` create does not currently have `t0` (but it could) and it has no `Init` function. 

> alternatively Reset could be called in the first step if we don't require a stepper to be initialized with the starting state

So we would assume `tR = t0 = 0` then?

include/sundials/sundials_stepper.h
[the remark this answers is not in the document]
balos1: I could imagine (but dont have a concrete example at the moment) that a root return might need special handling when checkpointing and/or backwards integration.

src/sundials/sundials_stepper.c
```diff
@@ -0,0 +1,214 @@
...
+
+#include "sundials/sundials_errors.h"
+#include "sundials/sundials_nvector.h"
+#include "sundials_stepper_impl.h"
+
+SUNErrCode SUNStepper_Create(SUNContext sunctx, SUNStepper* stepper)
+{
+  SUNFunctionBegin(sunctx);
+
+  *stepper = NULL;
```
Steven-Roberts, line 16: Unnecessary

src/sundials/sundials_stepper.c
```diff
@@ -0,0 +1,214 @@
...
+
+  return SUN_SUCCESS;
+}
+
+SUNErrCode SUNStepper_GetForcingData(SUNStepper stepper, sunrealtype* tshift,
+                                     sunrealtype* tscale, N_Vector** forcing,
+                                     int* nforcing)
+{
+  sunrealtype tau, taui;
+  int i;
```
Steven-Roberts, line 206: These seem unnecessary

src/sundials/sundials_stepper.c
```diff
@@ -0,0 +1,214 @@
...
+  *stepper_ptr = NULL;
+
+  return SUN_SUCCESS;
+}
+
+SUNErrCode SUNStepper_Advance(SUNStepper stepper, sunrealtype t0,
+                              sunrealtype tout, N_Vector y, sunrealtype* tret,
+                              int* stop_reason)
+{
+  SUNFunctionBegin(stepper->sunctx);
```
Steven-Roberts, line 59: Do all the functions in this file need to check if `stepper` is `NULL`?

balos1: No, we do not need to check that. If the `stepper` is NULL we cannot handle the error properly anyways.

src/sundials/sundials_stepper.c
balos1: Yeah they are leftover from copy/pasting from `MRIStepInnerStepper`

src/arkode/arkode_arkstep.c
balos1: In our team discussions we defined the `Advance` function to operate in this way. That is, we said `Advance` would evolve to precisely `tout` as opposed to `Evolve` which would potentially step past `tout` and interpolate. `MRIStep` needs this.  It sounds like you need an `Evolve` function for the operator splitting stepper.

Steven-Roberts: Ah I didn't realize we would have both `Advance` and `Evolve`. I that case, this is not an issue.

include/sundials/sundials_stepper.h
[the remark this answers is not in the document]
gardner48: > Do you mean the package "steppers"? The `SUNStepper` create does not currently have `t0` (but it could) and it has no `Init` function.

Yes, I was thinking the constructor for an implementation of a `SUNStepper` (i.e., `MyStepper(t0, y0, ctx)`) could take the initial condition. The package steppers are a little different since they wrap an integrator that was already created with the initial condition set. Alternatively having `SUNStepper_Init(t0, y0)` of `SUNStepper_Init(t0, y0, yp0)` might not be a bad idea.

> So we would assume `tR = t0 = 0` then?

Just `tR = t0` since `t0` is not necessarily `0`

include/sundials/sundials_stepper.h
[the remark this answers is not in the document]
gardner48: It would be nice it this could just encoded in the return value but having a `SUNErrCode` also indicate a status would be mixing purposes.

include/sundials/sundials_stepper.h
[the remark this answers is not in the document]
balos1: > Just tR = t0 since t0 is not necessarily 0

Well, yes, but my point was that if there was no `Init` that took `t0` (or `Create`) then we would not know `t0`. 

>The package steppers are a little different since they wrap an integrator that was already created with the initial condition set.

CVODE/IDA do not take the initial condition and `t0` until their `Init` routines are called.  So you could have the case where a stepper is created with a integrator that does not yet have the IC/t0 set. With that in mind, I think having ` SUNStepper_Init` routine is cleaner.

src/arkode/arkode_arkstep.c
gardner48: In the last discussion we dropped `Evolve` because it seemed like we did not need it. @Steven-Roberts for the splitting methods, would `Advance` still be sufficient or do you need to step past the time the method is supposed to integrate to?

include/sundials/sundials_stepper.h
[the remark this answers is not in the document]
balos1: Indeed, I think its best to have the error code and status separated.

include/sundials/sundials_stepper.h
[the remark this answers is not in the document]
gardner48: > CVODE/IDA do not take the initial condition and `t0` until their `Init` routines are called. So you could have the case where a stepper is created with a integrator that does not yet have the IC/t0 set. With that in mind, I think having ` SUNStepper_Init` routine is cleaner.

I think have an `Init` function is reasonable. As far as wrapping CVODE or IDA as a stepper, creating a stepper from a CVODE / IDA instance that was not initialized would be a user-error as the RHS / RES function would also not be set.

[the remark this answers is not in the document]
drreynolds: I think that David's alternate option of each consuming package calling `Reset` is perhaps the most appropriate.  This already occurs in MRIStep (or it would take a minimal change to do so), and I anticipate that the operator-splitting stepper would also want to control the initial condition for any sub-stepper.

include/sundials/sundials_stepper.h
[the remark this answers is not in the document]
drreynolds: Agreed.

src/arkode/arkode_arkstep.c
Steven-Roberts: `Advance` alone would be sufficient, but then I don't see a way to handle symmetric operator splitting methods most efficiently. As an example of the functionality needed, consider integration a stepper from `t=0.5` to `t=1.5` with output needed at `t=1`. Two calls to `Evolve` would be ideal for this, while `Advance` might add more function evaluations by forcing the first call to end exactly at `t=1`. Originally, I thought adding an interpolation function to the stepper was the way to fix this, but on second thought I don't think it's reasonable to have interpolation data across the entire tspan passed to `Advance`.

Steven-Roberts: That's another way to look at it. Multiple outputs with one `Advance` call would work, but the function signature would start to look very different than `ARKStepEvolve` in normal mode. An alternative is to add a `SetTStop` function to `SUNStepper`.

Steven-Roberts: > Adding a stop time would also work but could lead to extra RHS evaluations.

I could only see this leading to extra RHS evaluations if it was set unnecessarily.

gardner48: If going from `t = 0.5` to `t = 1.5` with an output at `t = 1` and using t-stop at `t = 1`, the integrator will need to adjust its step to stop exactly at `t = 1` and so will need an extra step compared to stepping over and interpolating the output at `t = 1`.

Steven-Roberts: I was thinking that situation would be handled like this
```
SetTStop(stepper, 1.5);
Advance(stepper, 0.5, 1, , y, ...);
// The inner solve may have overstepped t=1, but it returns interpolated solution at t=1
// Do something with y
Advance(stepper, 0.5, 1, , y, ...);
// This will end exactly at t=1.5 and not overstep
```

gardner48: Ah, I see what you mean now. Yes, that would work as well.

balos1: That is precisely how `ARKStepEvolve` or any other of the other package "evolve"-type functions work.

gardner48: Right, it's sounding like we need `Evolve` + `SetStopTime` rather than `Advance`

balos1: Originally we talked about having four functions:

- `Evolve`:  step to tout, possibly passing it and returning interpolated solution at t=tout 
- `Advance`: step to tout, stopping exactly at tout and return
- `Step`: take one _successful_ step (or fail trying to) and return
- `TryStep`: take exactly one internal step and return

We said we did not have a use for `Evolve` yet, so I removed it from the `SUNStepper` API for now. However, it seems the operator splitting stepper does necessitate an `Evolve` function PLUS a `SUNStepper_SetTStop` function. 

If we add a `SUNStepper_SetTStop` function, then it should only be used with `Evolve`, NOT `Advance` otherwise there is no reason to have `Advance` as a function since it would work just like `Evolve`.

gardner48: I'm leaning toward `Evolve` + `SetStopTime` since `Advance` (or even `Evolve`) with multiple outputs could just be a wrapper looping over the output times and calling the one output `Evolve` with the right output vector.

Steven-Roberts: > However, it seems the operator splitting stepper does necessitate an Evolve function PLUS a SUNStepper_SetTStop function.

It does need the `Evolve` function. `SetTStop` would only be needed if `Advance` was removed (as I could be achieved by combining `Evolve` and `SetTStop`).

So I think the question is which set of functions to use
1. `Evolve`, 'Advance`, `Step`, `TryStep` or
2. `Evolve`, `SetTStop`, `Step`, `TryStep`

drreynolds: I'm a bit late to the party, but it sounds to me like `Evolve` + `SetStopTime` is the way to go.  I'll note that like Steven, in the MERK multirate time-stepping methods I evolve over a longer duration, periodically storing the solution for some intermediate stages.  It would be perfectly handled by this approach, whereas right now I anticipate that the MRIStepInnerStepper may take a small number of extra steps as it stops to return intermediate solutions.  I'll defer to the rest of the group on the need for `Step` and `TryStep` usage modes.

src/sundials/sundials_stepper.c
balos1: I dont think we know the `count` for these existing arrays.

include/sundials/sundials_stepper.h
```diff
@@ -0,0 +1,125 @@
...
+SUNDIALS_EXPORT
+SUNErrCode SUNStepper_AddForcing(SUNStepper stepper, sunrealtype t, N_Vector f);
+
+SUNDIALS_EXPORT
+SUNErrCode SUNStepper_GetForcingData(SUNStepper stepper, sunrealtype* tshift,
+                                     sunrealtype* tscale, N_Vector** forcing,
+                                     int* nforcing);
+
+SUNDIALS_EXPORT
+SUNErrCode SUNStepper_Advance(SUNStepper stepper, sunrealtype t0,
```
Steven-Roberts, line 100:
```suggestion
SUNErrCode SUNStepper_Evolve(SUNStepper stepper, sunrealtype t0,
```

src/sundials/sundials_stepper.c
```diff
@@ -0,0 +1,242 @@
...
+}
+
+SUNErrCode SUNStepper_SetStopTimeFn(SUNStepper stepper, SUNStepperSetStopTimeFn fn)
+{
+  SUNFunctionBegin(stepper->sunctx);
+  stepper->ops->setstoptime = fn;
+  return SUN_SUCCESS;
+}
+
+SUNErrCode SUNStepper_SetForcing(SUNStepper stepper, int count, N_Vector tmpl)
```
Steven-Roberts, line 170: Missing header declaration

include/sundials/sundials_stepper.h
```diff
@@ -0,0 +1,125 @@
...
+#include <sundials/sundials_core.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* TryStep step result flags */
+#define SUNSTEPPER_TRYSTEP_FAILED  -1
+#define SUNSTEPPER_TRYSTEP_SUCCESS +0
+#define SUNSTEPPER_TRYSTEP_ADAPT   +1
```
Steven-Roberts, line 25: Can't we just forward flags from the inner method, e.g., `tmp_flag` in `arkStep_TryStep`? If so, then these are unnecessary.

Steven-Roberts (commented):
Initial pass at latest changes

Steven-Roberts:
After some more thought and discussion with David on IDA and passing around `yp`, the primary issue I see is keeping it consistent with `y`. With operator splitting, for example, `y` jumps around between calls to evolve, and I don't see how to keep `yp` consistent. With MRI, the change in the forcing between stages can cause a similar issue. This is not something the outer methods can do and needs to be done in the inner SunStepper integrator.

I'd like to propose the following to make SunStepper self starting (all you need is `y` to reset then evolve):

- Remove `yp` as arguments
- To support an IDA SunStepper, have a constructor like
```
int IDACreateSUNStepper(void* inner_ida_mem, ConsistentYPrimeFn f, SUNStepper* stepper)
```
where there's a function which can be specified by users (we could have a default that uses `IDACalcIC`) on how to get `yp` give `y`.

src/sundials/sundials_stepper_impl.h
```diff
@@ -0,0 +1,61 @@
...
+  SUNStepperFullRhsFn fullrhs;
+  SUNStepperResetFn reset;
+  SUNStepperSetStopTimeFn setstoptime;
+};
+
+struct SUNStepper_
+{
+  /* stepper specific content and operations */
+  void* content;
+  SUNStepper_Ops ops;
```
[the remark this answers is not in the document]
balos1: We do not need the ops struct, but that is how we have usually done things. @gardner48  should we stick with ops being its own struct?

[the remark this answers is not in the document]
gardner48: I think we should keep it to be consistent with the other objects but we should consider this change for the next major release if it has a noticeable performance impact.

src/sundials/sundials_stepper_impl.h
```diff
@@ -0,0 +1,61 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct SUNStepper_Ops_* SUNStepper_Ops;
+
+struct SUNStepper_Ops_
+{
+  SUNStepperEvolveFn evolve;
+  SUNStepperOneStepFn onestep;
```
Steven-Roberts, line 28: Is the convention for struct members snake case? E.g., `one_step`

gardner48: Yes but historically the members in an ops function have been one word, all lower case.

drreynolds (commented):
Finished reviewing the documentation.

doc/arkode/guide/source/Usage/ARKStep/User_callable.rst
```diff
@@ -4400,3 +4400,25 @@ wrap an ARKStep memory block as an :c:type:`MRIStepInnerStepper`.
...
+   Wraps an ARKStep memory block as a :c:type:`SUNStepper`.
+
+   **Arguments:**
+      * *arkode_mem* -- pointer to the ARKStep memory block.
+      * *stepper* -- the :c:type:`SUNStepper` object.
+
+   **Return value:**
+      * *ARK_SUCCESS* if successful
+      * *ARK_MEM_FAIL* if a memory allocation failed
+      * *ARK_ILL_INPUT* if an argument has an illegal value.
```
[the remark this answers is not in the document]
Steven-Roberts: Yes, I'm in the middle of updating the docs following the ARKstep->ARKode change. Aiming to have that done by the end of the day, so I'd recommend deferring further docs reviewing until then. The code is fine to review whenever.

drreynolds (commented):
This looks great.  

We may need to be careful with the `SUNStepper` / `MRIStepInnerStepper` compatibility since the MRI adaptivity PR #564 adds a couple more routines to the `MRIStepInnerStepper`, so a "unified" `SUNStepper` would need those two extra ops to be used in an adaptive MRI sense.  I also recommend that in a soon-to-follow release we should consolidate on only the `SUNStepper`, retaining relevant wrapping as necessary for backwards-compatibility with existing `MRIStepInnerStepper` objects.  

As I mentioned in a comment on the docs, I sincerely doubt that any of the potential "consuming" packages (e.g., SplitStep, ForcingStep, discrete adjoint integration, XBraid, MRIStep) *actually* require the full set of `SUNStepper` ops to be implemented.  If each consuming package clearly documented their requirements, then all "consumers" could standardize on `SUNStepper`.

doc/shared/sunstepper/SUNStepper_Description.rst
```diff
@@ -0,0 +1,348 @@
...
+
+      This function only frees memory allocated within the base class and the
+      base class structure itself. The user is responsible for freeing any
+      memory allocated for the member data (content).
+
+
+.. _SUNStepper.Description.BaseMethods.SteppingFunctions:
+
+Stepping Functions
+^^^^^^^^^^^^^^^^^^
```
Steven-Roberts, line 99: Note for reviewers: Unlike `MRIStepInnerStepper`, the functions in this section are publicly documented. They can be hidden if folks think that's preferable.

doc/shared/sunstepper/SUNStepper_Description.rst
```diff
@@ -0,0 +1,348 @@
...
+      responsible for evaluating ODE right-hand side function :math:`f(t, v)` as
+      well as computing and applying the forcing term :eq:`SUNStepper_forcing`
+      to obtain the full right-hand side of the ODE :eq:`SUNStepper_IVP`.
+
+.. _SUNStepper.Description.BaseMethods.Content:
+
+Attaching and Accessing the Content Pointer
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. c:function:: SUNErrCode SUNStepper_SetContent(SUNStepper stepper, void *content)
```
Steven-Roberts, line 178: Note for reviewers: these functions are new to `SUNStepper` compared to `MRIStepInnerStepper`. Since the `SUNStepper` struct is private, I think this is the proper way to access the `last_flag`.

include/sundials/sundials_stepper.h
```diff
@@ -0,0 +1,113 @@
...
+SUNErrCode SUNStepper_GetLastFlag(SUNStepper stepper, int* last_flag);
+
+SUNDIALS_EXPORT
+SUNErrCode SUNStepper_SetEvolveFn(SUNStepper stepper, SUNStepperEvolveFn fn);
+
+SUNDIALS_EXPORT
+SUNErrCode SUNStepper_SetOneStepFn(SUNStepper stepper, SUNStepperOneStepFn fn);
+
+SUNDIALS_EXPORT
+SUNErrCode SUNStepper_SetFullRhsFn(SUNStepper stepper, SUNStepperFullRhsFn fn);
```
Steven-Roberts, line 97: Note to reviewers: there is no analogous `SUNStepper_FullRhs`. This is following what MRI does where it was presumably excluded because the full RHS functions are private in ARKODE. I can make a public `SUNStepper_FullRhs` function if folks want.

doc/shared/sunstepper/SUNStepper_Description.rst
```diff
@@ -0,0 +1,467 @@
...
+
+.. _SUNStepper.Description.ImplMethods:
+
+Implementation Specific Methods
+-------------------------------
+
+This section describes the virtual methods defined by the :c:type:`SUNStepper`
+abstract base class.
+
+An :c:type:`SUNStepper` *must* provide implementations of the following
```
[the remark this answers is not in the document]
Steven-Roberts: I've updated the docs to indicate they are optional and determined by the "consumer" of the `SUNStepper`. I'll need to specify the required functionality in the operator splitting PR.

swig/sundials/fsundials_stepper.i
```diff
@@ -0,0 +1,26 @@
...
+// ---------------------------------------------------------------
+// Swig interface file
+// ---------------------------------------------------------------
+
+%{
+#include "sundials/sundials_stepper.h"
+%}
+
+// TODO(SBR): Can someone confirm this is correct
+%apply void* { SUNStepper };
```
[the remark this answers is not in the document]
balos1: This is correct unless we want users to be able to make a custom SUNStepper in Fortran.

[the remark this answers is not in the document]
balos1: Additional note... since we do not make the `SUNStepper` struct public, users cannot fully customize a `SUNStepper` anyways (only the ops), so this makes sense.

doc/arkode/guide/source/Usage/MRIStep/Custom_Inner_Stepper/Description.rst
```diff
@@ -79,6 +79,32 @@ Creating and Destroying an Object
...
+   **Example usage:**
+
+   .. code-block:: C
+
+      SUNStepper sunstepper = NULL;
+      SUNStepper_Create(ctx, &sunstepper);
+      /* Attach content and functions to the SUNStepper... */
+
+      MRIStepInnerStepper inner_stepper = NULL;
+      flag = MRIStepInnerStepper_Create(sunstepper, &inner_stepper);
```
drreynolds, line 103: How critical is it that the user fully configure the SUNStepper *before* they wrap it as an MRIStepInnerStepper?  If this copies existing function pointers then the order is important, in which case this documentation should contain a "note" to clarify the assumed order.  If the routine just attaches wrapper routines, and the SUNStepper could be modified after creating the MRIStepInnerStepper, then the current documentation should be sufficient.

doc/arkode/guide/source/Usage/User_callable.rst
```diff
@@ -4685,3 +4685,26 @@ rescale the upcoming time step by the specified factor.  If a value
...
+   Wraps an ARKODE integrator as a :c:type:`SUNStepper`.
+
+   :param arkode_mem: pointer to the ARKODE memory block.
+   :param stepper: the :c:type:`SUNStepper` object.
+
+   :retval ARK_SUCCESS: the function exited successfully.
+   :retval ARK_MEM_FAIL: a memory allocation failed.
+   :retval ARK_SUNSTEPPER_ERR: the :c:type:`SUNStepper` initialization failed.
+
+   .. versionadded:: x.y.z
```
drreynolds, line 4710: Are there any assumed requirements about the ARKODE solver module?  Specifically, since some steppers "support" all ARKODE functionality, whereas other steppers only support a tiny subset, then it would be good to clarify which steppers this function would work for.

doc/shared/sunstepper/SUNStepper_Description.rst
```diff
@@ -0,0 +1,348 @@
...
+functions for creating and destroying a generic base class object, attaching and
+retrieving the derived class ``content`` pointer, and setting function pointers
+to derived class method implementations.
+
+.. _SUNStepper.Description.BaseMethods.CreateDestroy:
+
+Creating and Destroying an Object
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In additional to creating an empty :c:type:`SUNStepper` using
```
drreynolds, line 53:
```suggestion
In addition to creating an empty :c:type:`SUNStepper` using
```

doc/shared/sunstepper/SUNStepper_Description.rst
```diff
@@ -0,0 +1,348 @@
...
+
+   :param stepper: a stepper object.
+   :param content: a pointer to set to the stepper member data pointer.
+   :return: A :c:type:`SUNErrCode` indicating success or failure.
+
+
+Handling Warnings and Errors
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+An implementation of a :c:type:`SUNStepper` may have system of warning and error
```
drreynolds, line 201:
```suggestion
An implementation of a :c:type:`SUNStepper` may have a system of warning and error
```

doc/shared/sunstepper/SUNStepper_Implementing.rst
```diff
@@ -0,0 +1,56 @@
...
+   See the top-level LICENSE and NOTICE files for details.
+
+   SPDX-License-Identifier: BSD-3-Clause
+   SUNDIALS Copyright End
+   ----------------------------------------------------------------
+
+.. _SUNStepper.Implementing:
+
+Implementing a SUNStepper
+=========================
```
drreynolds, line 19: It would be great if we had an example that implemented a custom SUNStepper, in which case this file could also mention an example where users can see these instructions "in action."  I'm imagining a simple operator-splitting example where one partition has an analytical solution.

src/arkode/arkode_sunstepper.c
```diff
@@ -0,0 +1,164 @@
...
+  if (arkode_mem == NULL)
+  {
+    arkProcessError(NULL, ARK_MEM_NULL, __LINE__, __func__, __FILE__,
+                    MSG_ARK_NO_MEM);
+    return (ARK_MEM_NULL);
+  }
+  ARKodeMem ark_mem = (ARKodeMem)arkode_mem;
+
+  SUNErrCode err = SUNStepper_Create(ark_mem->sunctx, stepper);
+  if (err != SUN_SUCCESS) { return ARK_SUNSTEPPER_ERR; }
```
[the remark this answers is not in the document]
drreynolds: I don't feel strongly on this question either way.  While it is an ARKODE-provided routine (and so `arkProcessError` makes sense), it is a utility routine provided for `SUNStepper` users, so I could equivalently see it returning a `SUNErrCode` instead of an `int`, and "feeling" more similar to the SUNStepper routines.

drreynolds (commented):
A few minor items, but once those are resolved I think this is good to merge.

doc/arkode/guide/source/Usage/MRIStep/Custom_Inner_Stepper/Description.rst
Steven-Roberts: `MRIStepInnerStepper_CreateFromSUNStepper` wraps the `SUNStepper` and makes no copies of function pointers. The order in which `MRIStepInnerStepper_CreateFromSUNStepper` and the `SUNStepper_Set*` functions are called shouldn't matter.

doc/shared/sunstepper/SUNStepper_Implementing.rst
Steven-Roberts: In the operator splitting PR, currently only a unit test creates a custom SUNStepper. I can see about modifying the advection-diffusion-reaction example so we have some something to point to.

doc/arkode/guide/source/Usage/User_callable.rst
Steven-Roberts: The only functionality not supported by all of ARKODE is `SUNStepper_SetForcing`. I'll make note of that here and in ARKStep, ERKStep, etc.

doc/shared/sunstepper/SUNStepper_Implementing.rst
drreynolds: Or this could be added in the near future (not required for approval of this PR).

doc/superbuild/source/sunstepper/index.rst
gardner48, line 1: See edits in `doc/arkode/guide/source/sunstepper/index.rst`

include/sundials/sundials_stepper.h
```diff
@@ -0,0 +1,113 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef int (*SUNJacFn)(sunrealtype t, N_Vector y, N_Vector fy, SUNMatrix Jac,
+                        void* user_data, N_Vector tmp1, N_Vector tmp2,
+                        N_Vector tmp3);
+
+typedef int (*SUNJacTimesFn)(N_Vector v, N_Vector Jv, sunrealtype t, N_Vector y,
+                             N_Vector fy, void* user_data, N_Vector tmp);
```
gardner48, line 27: In case we need similar functions for DAEs in the future
```suggestion
typedef int (*SUNODEJacFn)(sunrealtype t, N_Vector y, N_Vector fy, SUNMatrix Jac,
                           void* user_data, N_Vector tmp1, N_Vector tmp2,
                           N_Vector tmp3);

typedef int (*SUNODEJacTimesFn)(N_Vector v, N_Vector Jv, sunrealtype t, N_Vector y,
                                N_Vector fy, void* user_data, N_Vector tmp);
```

doc/shared/sunstepper/SUNStepper_Description.rst
```diff
@@ -0,0 +1,348 @@
...
+
+   :param stepper: the stepper object.
+   :param t0: the initial time for the integration.
+   :param tout: the final time for the integration.
+   :param vout: on output, the state at time *tout*.
+   :param tret: the final time corresponding to the output value *vout*.
+   :return: A :c:type:`SUNErrCode` indicating success or failure.
+
+
+.. c:function:: SUNErrCode SUNStepper_OneStep(SUNStepper stepper, sunrealtype t0, sunrealtype tout, N_Vector vout, sunrealtype* tret)
```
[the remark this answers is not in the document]
gardner48: It looks like this is not used in splitting or adjoint PRs. Can this function be removed?

include/sundials/sundials_stepper.h
gardner48: I think this could make sense, especially since something using a `SUNStepper` might need the full RHS for some reason

src/arkode/arkode_impl.h
```diff
@@ -454,6 +460,10 @@ struct ARKodeMemRec
   ARKTimestepGetMassMemFn step_getmassmem;
   ARKMassMultFn step_mmult;
 
+  /* Time stepper module -- forcing */
+  sunbooleantype step_supports_forcing;
```
gardner48, line 464: Will be removed in #564
```suggestion
```

src/arkode/arkode_mristep.c
```diff
@@ -2840,6 +2888,17 @@ int mriStepInnerStepper_FullRhs(MRIStepInnerStepper stepper, sunrealtype t,
   return stepper->last_flag;
 }
 
+int mriStepInnerStepper_FullRhsSUNStepper(MRIStepInnerStepper stepper,
+                                          sunrealtype t, N_Vector y, N_Vector f,
+                                          SUNDIALS_MAYBE_UNUSED int mode)
+{
+  SUNStepper sunstepper = (SUNStepper)stepper->content;
+  SUNErrCode err        = sunstepper->ops->fullrhs(sunstepper, t, y, f);
```
gardner48, line 2896: Should the SUNStepper FullRhs also have a mode input in case it can leverage that information to avoid a function evaluation?

src/sundials/sundials_stepper.c
```diff
@@ -0,0 +1,171 @@
+
```
gardner48, line 1: Add usual comment block at top of file

src/sundials/sundials_stepper.c
```diff
@@ -0,0 +1,171 @@
...
+}
+
+SUNErrCode SUNStepper_Destroy(SUNStepper* stepper_ptr)
+{
+  if (stepper_ptr != NULL)
+  {
+    free((*stepper_ptr)->ops);
+    free(*stepper_ptr);
+    *stepper_ptr = NULL;
+  }
```
gardner48, line 47: We should probably allow the user to attach their own `Destroy` implementation to free any stepper content

gardner48 (changes requested):

doc/shared/sunstepper/SUNStepper_Description.rst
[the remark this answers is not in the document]
Steven-Roberts: I think it can be removed unless @balos1 has an objection.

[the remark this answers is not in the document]
balos1: It can be removed. When we discussed the design, we included it for future reasons, but we can add it when its needed.

external/CMakeLists.txt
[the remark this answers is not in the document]
balos1: Im assuming the formatter did this?

include/sundials/sundials_stepper.h
balos1: We could do this, or we could just name the DAE ones `SUNResJacFn` etc if we add support for them later.

external/CMakeLists.txt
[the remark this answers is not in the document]
balos1: I think @Steven-Roberts must have setup the formatter locally and the incorrect configuration was applied to this file. We should revert the change to this file.

balos1:
I actually need the OneStep function in the adjoint work, so I have added it back into here.

drreynolds (approved):

drreynolds:
It looks like the only remaining issue with the CI is that the documentation updates have some label errors: 
```
/var/lib/jenkins/workspace/SUNDIALS_PR-463/doc/shared/RecentChanges.rst:5: WARNING: undefined label: 'arkode.usage.splittingstep'
/var/lib/jenkins/workspace/SUNDIALS_PR-463/doc/shared/RecentChanges.rst:5: WARNING: undefined label: 'arkode.usage.forcingstep'
/var/lib/jenkins/workspace/SUNDIALS_PR-463/doc/shared/sunstepper/SUNStepper_Implementing.rst:42: WARNING: c:func reference target not found: SplittingStepCreate
/var/lib/jenkins/workspace/SUNDIALS_PR-463/doc/shared/sunstepper/SUNStepper_Implementing.rst:42: WARNING: c:func reference target not found: ForcingStepCreate
/var/lib/jenkins/workspace/SUNDIALS_PR-463/doc/shared/sunstepper/SUNStepper_Implementing.rst:55: WARNING: c:func reference target not found: SplittingStepCreate
/var/lib/jenkins/workspace/SUNDIALS_PR-463/doc/shared/sunstepper/SUNStepper_Implementing.rst:55: WARNING: c:func reference target not found: ForcingStepCreate
/var/lib/jenkins/workspace/SUNDIALS_PR-463/doc/shared/RecentChanges.rst:5: WARNING: undefined label: 'arkode.usage.splittingstep'
/var/lib/jenkins/workspace/SUNDIALS_PR-463/doc/shared/RecentChanges.rst:5: WARNING: undefined label: 'arkode.usage.forcingstep'
```

Steven-Roberts:
Yeah, wasn't sure how to best handle that. Those are references to parts of the docs that will be added in the operator splitting PR (which is merged after this). I can temporarily comment out or remove the references out to get the CI passing.

drreynolds:
> Yeah, wasn't sure how to best handle that. Those are references to parts of the docs that will be added in the operator splitting PR (which is merged after this). I can temporarily comment out or remove the references out to get the CI passing.

I forget, but is the "merge" plan to approve/merge this PR before #572?  If that's the case, then perhaps we should move these bits of documentation into #572 so that both "squashed" and merged PRs can pass all CI tests?

Steven-Roberts:
> I forget, but is the "merge" plan to approve/merge this PR before https://github.com/LLNL/sundials/pull/572?

Yes

> If that's the case, then perhaps we should move these bits of documentation into https://github.com/LLNL/sundials/pull/572 so that both "squashed" and merged PRs can pass all CI tests?

Ok, I'll do that

doc/shared/sunstepper/SUNStepper_Description.rst
```diff
@@ -244,16 +280,6 @@ determined by the "consumer" of the :c:type:`SUNStepper`.
    :return: A :c:type:`SUNErrCode` indicating success or failure.
 
 
-.. c:function:: SUNErrCode SUNStepper_SetOneStepFn(SUNStepper stepper, SUNStepperOneStepFn fn)
```
gardner48: Since the `OneStep` function was added back, this documentation needs to be added back as well

doc/shared/sunstepper/SUNStepper_Description.rst
```diff
@@ -303,30 +340,21 @@ This section describes the virtual methods defined by the :c:type:`SUNStepper`
...
 
 
-.. c:type:: SUNErrCode (*SUNStepperEvolveFn)(SUNStepper stepper, sunrealtype t0, sunrealtype tout, N_Vector v, sunrealtype* tret, int* stop_reason)
+.. c:type:: SUNErrCode (*SUNStepperEvolveFn)(SUNStepper stepper, sunrealtype tout, N_Vector vret, sunrealtype* tret)
 
    This type represents a function with the signature of
    :c:func:`SUNStepper_Evolve`.
 
 
-.. c:type:: SUNErrCode (*SUNStepperOneStepFn)(SUNStepper stepper, sunrealtype t0, sunrealtype tout, N_Vector v, sunrealtype* tret, int* stop_reason)
```
gardner48: Add back the `SUNStepperOneStepFn` documentation

doc/shared/sunstepper/SUNStepper_Description.rst
```diff
@@ -346,3 +374,9 @@ abstract base class.
    This type represents a function with the signature of
    :c:func:`SUNStepper_SetForcing`.
 
+
+.. c:type:: SUNErrCode (*SUNStepperDestroyFn)(SUNStepper stepper)
```
gardner48, line 378:
```suggestion
.. c:type:: SUNErrCode (*SUNStepperDestroyFn)(SUNStepper* stepper)
```

include/sundials/sundials_stepper.h
```diff
@@ -50,19 +57,25 @@ typedef SUNErrCode (*SUNStepperSetForcingFn)(SUNStepper stepper,
                                              sunrealtype tscale,
                                              N_Vector* forcing, int nforcing);
 
+typedef SUNErrCode (*SUNStepperDestroyFn)(SUNStepper stepper);
```
gardner48, line 60:
```suggestion
typedef SUNErrCode (*SUNStepperDestroyFn)(SUNStepper* stepper);
```

gardner48 (changes requested):

doc/shared/sunstepper/SUNStepper_Description.rst
Steven-Roberts: This user-specified function is only responsible for freeing the content. It can' be responsible for freeing ops and the stepper itself since those details are not public. To reduce the risk of users freeing the wrong things, I think we should keep the argument as `SUNStepper stepper`.

gardner48: Ah, yes this is a bit different since the definition of the `SUNStepper` is private. I think this is fine as is then.

drreynolds (approved):

src/arkode/arkode_sunstepper.c
```diff
@@ -0,0 +1,209 @@
...
+  err = SUNStepper_SetEvolveFn(*stepper, arkSUNStepperEvolve);
+  if (err != SUN_SUCCESS)
+  {
+    arkProcessError(ark_mem, ARK_SUNSTEPPER_ERR, __LINE__, __func__, __FILE__,
+                    "Failed to set SUNStepper evolve function");
+    return ARK_SUNSTEPPER_ERR;
+  }
+
+  err = SUNStepper_SetOneStepFn(*stepper, arkSUNStepperOneStep);
+  if (err != SUN_SUCCESS) { return ARK_SUNSTEPPER_ERR; }
```
gardner48, line 171:
```suggestion
  if (err != SUN_SUCCESS)
  {
    arkProcessError(ark_mem, ARK_SUNSTEPPER_ERR, __LINE__, __func__, __FILE__,
                    "Failed to set SUNStepper one step function");
    return ARK_SUNSTEPPER_ERR;
  }
```

gardner48 (changes requested):
One minor change (missing process error call) that I'll apply and approve in a sec.

gardner48 (approved):

PR #470, opened 2024-05-07
Maintenance/arkode cleanup

drreynolds:
Cleaned up ARKODE source code files so that function declarations/implementations are in more logical locations (following the restructuring of ARKODE's user-interface).


gardner48 (changes requested):
A few minor things, otherwise this looks good.

gardner48 (approved):

PR #474, opened 2024-05-12
Feature: Add option to disable interpolation in ARKODE

gardner48:
Add support for disabling interpolated output in ARKODE steppers by passing `ARK_INTERP_NONE` to `ARKodeSetInterpolantType`. Disabling interpolation can reduce the memory footprint of an integrator by two or more state vectors (depending on the interpolation type and degree) which can be beneficial when interpolation is not needed e.g., when integrating to a final time without output in between or using an explicit fast time scale integrator with MRIStep.

When interpolation is disabled, rootfinding is not supported, implicit methods must use the trivial predictor, interpolation at stop times cannot be used, and calls to `ARKodeEvolve` in `ARK_NORMAL` mode will return at or past the requested output time.


doc/arkode/guide/source/Usage/ARKStep_c_interface/User_callable.rst
gardner48, line 1: If everyone is good with this change I think we should propagate it to the other stepper-specific deprecated functions to simplify documentation updates until they are removed.

balos1 (changes requested):
Just a couple minor comments

doc/shared/RecentChanges.rst
```diff
@@ -39,43 +32,80 @@ Added the following MRI coupling tables
...
 
 Check if size of ``SuiteSparse_long`` is 8 if the size of ``sunindextype`` is 8
 when using KLU.
+
+**Deprecation Notices**
+
+Numerous ARKODE stepper-specific functions are now deprecated in favor of
+ARKODE-wide functions.
+
+Deprecated ``ARKStepSetOptimalParams`` function. This functions does not have an
```
drreynolds, line 109:
```suggestion
Deprecated the `ARKStepSetOptimalParams` function. Since this function does not have an
```

src/arkode/arkode_interp.c
```diff
@@ -311,29 +312,18 @@ void arkInterpPrintMem_Hermite(ARKInterp interp, FILE* outfile)
...
+    arkProcessError(ark_mem, ARK_INTERP_FAIL, __LINE__, __func__, __FILE__,
+                    "Illegal degree specified.");
+    return ARK_ILL_INPUT;
   }
 
-  /* on negative degree, check for allowable value and update stored degree */
-  degree = -degree;
-  if (degree > ARK_INTERP_MAX_DEGREE) { degree = ARK_INTERP_MAX_DEGREE; }
-  HINT_DEGREE(interp) = SUNMIN(HINT_DEGREE(interp), degree);
+  if (degree < 0) { degree = ARK_INTERP_MAX_DEGREE; }
```
drreynolds, line 322: I don't think that this routine can be called with a negative input anymore, since you removed these calls from the steppers.  I'll need to think through the new logic you've put together to determine whether the previous behavior is fully replicated (I anticipate that it's fine, now that you added the `ark_mem->interp_degree` variable); however, since a negative input can no longer occur, please update the comment block at the top of this function, and remove this line.
```suggestion
```

src/arkode/arkode_interp.c
```diff
@@ -1036,29 +1026,18 @@ void arkInterpPrintMem_Lagrange(ARKInterp I, FILE* outfile)
...
+    arkProcessError(ark_mem, ARK_INTERP_FAIL, __LINE__, __func__, __FILE__,
+                    "Illegal degree specified.");
+    return ARK_ILL_INPUT;
   }
 
-  /* on negative degree, check for allowable value and update stored degree */
-  degree = -degree;
-  if (degree > ARK_INTERP_MAX_DEGREE) { degree = ARK_INTERP_MAX_DEGREE; }
-  LINT_NMAX(I) = SUNMIN(LINT_NMAX(I), degree + 1);
+  if (degree < 0) { degree = ARK_INTERP_MAX_DEGREE; }
```
drreynolds, line 1036: Same comment here about the possibility of negative input values, and the comment block preceding the function.
```suggestion
```

drreynolds (approved):

balos1 (approved):

balos1 (approved):

PR #475, opened 2024-05-14
CMake: Fix dll install location

jschueller:
This fixes dll search on conda where dlls are expected to be installed in prefix/bin (pretty standard layout)
static or import libs still go to prefix/lib


mmuetzel:
The same issue is solved slightly differently with a patch in MSYS2:
https://github.com/msys2/MINGW-packages/blob/8ff52af093c2d67d665e147d79049819e902e2bb/mingw-w64-sundials/0003-sundials-use-default-installdirs.patch

But admittedly, I prefer your proposed change.
Looks good to me. 👍

jschueller:
didnt know that you could omit DESTINATION like that, but that's only since cmake 3.14

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Changes to SUNDIALS in release X.Y.Z
 
+Fixed the runtime library installation path for windows systems. This fix changes the
```
balos1, line 5:
```suggestion
Fixed the runtime library installation path for windows systems. This fix changes the 
default library installation path from `CMAKE_INSTALL_PREFIX/CMAKE_INSTALL_LIBDIR` to
`CMAKE_INSTALL_PREFIX/CMAKE_INSTALL_BINDIR`.
```

balos1: Markdown formatting should use single backticks for inline code. Please also add this to `docs/shared/RecentChanges.rst` as well (which expects the double backticks).

balos1 (changes requested):
Just a couple request w.r.t. the change notes. Please also sign your commit as per our [Contributing guide](https://github.com/LLNL/sundials/blob/main/CONTRIBUTING.md).

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
...
+default library installation path from ``CMAKE_INSTALL_PREFIX/CMAKE_INSTALL_LIBDIR`` to
+``CMAKE_INSTALL_PREFIX/CMAKE_INSTALL_BINDIR``.
```
balos1, line 7:
```suggestion
`CMAKE_INSTALL_PREFIX/CMAKE_INSTALL_BINDIR`.
```

CHANGELOG.md
jschueller: ok

balos1 (approved):

PR #479, opened 2024-05-15
use LC template for gitlab id tokens and limit github actions push events

balos1:
Addresses https://hpc.llnl.gov/technical-bulletins/bulletin-568 and also limits our github action workflows to only run on push to certain branches.

### DIFF
--- a/.github/workflows/double-precision.yml
+++ b/.github/workflows/double-precision.yml
@@ -6,7 +6,6 @@ on:
     branches:
       - main
       - develop
-      - 'releases/**'
   pull_request:
   merge_group:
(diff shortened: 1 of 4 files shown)


gardner48 (approved):

balos1:
@gardner48 piggy-backing on this PR to limit workflows to push event for specific branches

test/answers
gardner48, line 1: Was this an intentional update? If so we'll need to update the answers repo.

balos1: It was unintentional

gardner48 (approved):

PR #481, opened 2024-05-16
Fix missing statics and dead code identified with -Wmissing-declarations

Steven-Roberts:


src/nvector/manyvector/nvector_manyvector.c
```diff
@@ -597,7 +597,8 @@ MPI_Comm N_VGetCommunicator_MPIManyVector(N_Vector v)
 }
 #else
 /* This function retrieves the MPI Communicator from a ManyVector object. */
-SUNComm N_VGetCommunicator_ManyVector(N_Vector v) { return SUN_COMM_NULL; }
+//TODO: is this needed?
+//SUNComm N_VGetCommunicator_ManyVector(N_Vector v) { return SUN_COMM_NULL; }
```
Steven-Roberts, line 601: Unsure it this function is intended to be public

src/arkode/arkode_sprk.c
```diff
@@ -23,7 +23,7 @@
 #include "arkode/arkode_butcher.h"
 #include "arkode_impl.h"
 
-ARKodeSPRKTable ARKodeSymplecticEuler(void)
+static ARKodeSPRKTable ARKodeSymplecticEuler(void)
```
Steven-Roberts, line 26: Maybe `arkodeSymplecticEuler` now that it's private?

gardner48: Yes and similarly for the other internal functions

src/sundials/sundials_adaptcontroller.c
```diff
@@ -64,23 +64,6 @@ SUNAdaptController SUNAdaptController_NewEmpty(SUNContext sunctx)
...
-  if (C == NULL) { return; }
-
-  /* free non-NULL ops structure */
-  if (C->ops) { free(C->ops); }
-  C->ops = NULL;
-
-  /* free overall SUNAdaptController object and return */
-  free(C);
-  return;
-}
```
gardner48, line 82: I think the intent was for this to be public but it's missing from the header

src/nvector/manyvector/nvector_manyvector.c
gardner48: This is not needed, `N_VGetCommunicator` will return `SUN_COMM_NULL` if a `GetCommunicator` function is not attached.

gardner48 (changes requested):

src/sundials/sundials_adaptcontroller.c
drreynolds: I agree with @gardner48.  If I recall correctly, I'd included it as a utility for user-provided `SUNAdaptController` implementations (like `SUNAdaptController_NewEmpty`).  We should add it to the header.

src/sundials/sundials_band.c
drreynolds, line 1: This is great.  When we added the `SUNDlsMat_`-prefixed functions (around 5 years ago?), we kept these wrappers here for backward compatibility.  It looks like we never remembered to remove them.

This new use of `-Wmissing-declarations` will help us to avoid similar "half-removals" in the future.

drreynolds (dismissed):

include/sundials/sundials_adaptcontroller.h
```diff
@@ -91,6 +91,10 @@ struct _generic_SUNAdaptController
 SUNDIALS_EXPORT
 SUNAdaptController SUNAdaptController_NewEmpty(SUNContext sunctx);
 
+/* Function to free a generic SUNAdaptController (assumes content is already empty) */
+SUNDIALS_EXPORT
+void SUNAdaptController_DestroyEmpty(SUNAdaptController C);
```
gardner48, line 96: I think this needs documentation as well

gardner48 (changes requested):
A bit of documentation needed, otherwise this looks good to me

gardner48 (approved):

PR #483, opened 2024-05-16
Fix promotion of double to sunrealtype identified with -Wdouble-promotion

Steven-Roberts:


gardner48 (approved):

src/sunmatrix/sparse/sunmatrix_sparse.c
```diff
@@ -1298,7 +1298,7 @@ SUNErrCode format_convert(const SUNMatrix A, SUNMatrix B)
     sunindextype jj;
     for (jj = Ap[row]; jj < Ap[row + 1]; jj++)
     {
-      sunindextype col  = Aj[jj];
+      col               = Aj[jj];
       sunindextype dest = Bp[col];
```
drreynolds, line 1302: I was under the impression that in-loop variable declarations were not permitted in C (certainly not older versions), so I wonder how this function even compiled before the transition to C99.  Should we remove the other in-loop declarations (`temp` and `dest`), or are we good with this stylistic mix (I believe that the rest of the file consistently places all declarations at the beginning of the function)?

gardner48: This line must have come in after we allowed for C99 as the CI strictly enforced C90 before then. I think we should allow for mixed declaration, maybe not generally but at least in cases like this where the variable are only used in the local scope. That said, in this case the `col` inside the loop was shadowing the `col` outside the loop which can be error prone. I'll update the #484 to include `-Wshadow` to catch this.

balos1: > I was under the impression that in-loop variable declarations were not permitted in C  (certainly not older versions)
This impression is incorrect. Even under ANSI C it is perfectly fine to declare a local variable in a for loop body as long as it as the top of the scope. The rule is that declarations have to be at the beginning of the scope before the first processing statement.

balos1 (approved):

PR #484, opened 2024-05-16
CI: Update Compiler Warnings, Enable Warnings as Errors

gardner48:
* Additional warning flags from #481, #482, and #483
* Remove `-Wno-deprecated-declarations`
* Enable warnings as errors in GitHub actions


gardner48:
CI failures should be resolved as related PRs are merged

drreynolds (approved):

drreynolds (dismissed):

Steven-Roberts:
When building in single precision, `-Wdouble-promotion` causes tons of warnings from `printf` which can't be helped. Maybe it should be added conditionally.

gardner48:
> When building in single precision, `-Wdouble-promotion` causes tons of warnings from `printf` which can't be helped. Maybe it should be added conditionally.

Is there a configuration where we could have this flag and avoid the warnings from `printf`?

Steven-Roberts:
> Is there a configuration where we could have this flag and avoid the warnings from printf?

When precision is set to extended, `-Wdouble-promotion` is most useful. It's also fine to have the flag when precision is double, but it's not going to identify any promotion warnings.

gardner48:
> When precision is set to extended, `-Wdouble-promotion` is most useful. It's also fine to have the flag when precision is double, but it's not going to identify any promotion warnings.

Ah yes, the docs only talk about `float` to `double` promotion but it makes sense this would for `double` to `long double` promotion as well. I'll update this in a sec.

include/sunlinsol/sunlinsol_ginkgo.hpp
```diff
@@ -332,7 +332,7 @@ class LinearSolver : public ConvertibleTo<SUNLinearSolver>
     }
 
     iter_count_ = static_cast<int>(logger->get_num_iterations());
-    GkoExec()->get_master()->copy_from(gko::lend(GkoExec()), 1,
+    GkoExec()->get_master()->copy_from(GkoExec(), 1,
```
gardner48, line 335: `gko::lend` is deprecated but I'm not sure what version in which it was deprecated. We probably need a version specific macro for backward compatibility.

examples/sunlinsol/ginkgo/test_sunlinsol_ginkgo.cpp
```diff
@@ -143,10 +143,10 @@ __global__ void fill_kernel(sunindextype mat_rows, sunindextype mat_cols,
 }
 #endif
 
-void fill_matrix(gko::matrix::Csr<sunrealtype, sunindextype>* matrix)
+static void fill_matrix(std::shared_ptr<gko::matrix::Csr<sunrealtype, sunindextype>>& matrix)
```
gardner48, line 146: @balos1 Not sure if this is the right fix but I couldn't compile this test with Ginkgo 1.7 and GCC 13.2 without this update.

examples/sunlinsol/ginkgo/test_sunlinsol_ginkgo.cpp
```diff
@@ -501,7 +501,7 @@ int main(int argc, char* argv[])
     else
     {
       gko_matrix->fill(0.0);
-      fill_matrix(gko::lend(gko_matrix));
+      fill_matrix(gko_matrix);
```
gardner48, line 504: Same comment about `gko::lend` as earlier

examples/sunlinsol/ginkgo/test_sunlinsol_ginkgo.cpp
```diff
@@ -480,7 +480,7 @@ int main(int argc, char* argv[])
       gko_matdata.remove_zeros();
       gko_matrix->read(gko_matdata);
     }
-    else { fill_matrix(gko::lend(gko_matrix)); }
+    else { fill_matrix(gko_matrix); }
```
gardner48, line 483: Same comment about `gko::lend` as earlier

include/sunlinsol/sunlinsol_ginkgo.hpp
balos1: It appears that it was deprecated in 1.7.0

examples/sunlinsol/ginkgo/test_sunlinsol_ginkgo.cpp
balos1: Is the reference needed?

gardner48: Nope

include/sunlinsol/sunlinsol_ginkgo.hpp
gardner48: Turns out it's actually 1.6 where `gko::lend` was deprecated. The change log notes it more clearly in 1.7 but it's also mentioned in the 1.6 change log. Tested with 1.5 and 1.6 to confirm.

balos1 (approved):

PR #486, opened 2024-05-16
CI: Combine actions

gardner48:
Combine some actions so it's easier to trigger all the tests manually


.github/workflows/ubuntu-latest.yml
```diff
@@ -24,8 +24,21 @@ jobs:
...
+          - buildtype: Debug
+            precision: single
+          # - buildtype: Debug
+          #   precision: extended
+          - buildtype: Release
+            precision: single
+          # - buildtype: Release
+          #   precision: extended
+          - buildtype: RelWithDebInfo
+            precision: double
```
drreynolds, line 40: Since our "standard" installation uses both `double` and `RelWithDebInfo`, is that really a configuration that we want to "exclude" in this workflow?  I'm guessing that we just exclude some to limit the combinatorial explosion of tests (e.g., once extended precision is running, we'd have 9 possible combinations) -- is there some logic behind this selection of excluded combinations?

gardner48: This reproduces what we currently run for test cases with this workflow. There are other jobs that build `double` without setting the build type so I suspect that's why the original version of the file did not include this case. We probably we need to revisit this can refine our coverage.

drreynolds (approved):

PR #487, opened 2024-05-16
CI: Add Fortran Compiler Warnings

gardner48:
Add Fortran `-check` flags with `ENABLE_ALL_WARNINGS`


gardner48:
Testing if bounds checking works with the CI. Previously we ran into compiler bugs with the Jenkins tests but I think that was with an older GCC version than use now.

gardner48:
The `gfortran` version Jenkins uses still segfaults with the additional Fortran warning flags so these flags are only added when using the docker GitHub actions tests.

drreynolds (approved):

balos1 (approved):

PR #488, opened 2024-05-17
Loop over all Butcher tables to test order

Steven-Roberts:
Switches from a vector of table names (which was often not updated) to iterating over the enum of tables so methods get included in this test automatically.


Steven-Roberts:
Is there any need to do outfile checking for this test? Isn't the return code of the test enough?

test/unit_tests/arkode/CXX_serial/ark_test_butcher.cpp
```diff
@@ -123,12 +90,13 @@ int main()
 
   // loop over individual DIRK tables
   std::cout << "\nTesting individual DIRK methods:\n\n";
-  for (std::string table : Tables_DIRK)
+  for (int table = ARKODE_MIN_DIRK_NUM; table <= ARKODE_MAX_DIRK_NUM; table++)
   {
     std::cout << "Testing method " << table << ":";
```
drreynolds, line 95: I like that this file is somewhat more maintainable as we add new Butcher tables to ARKODE.  However, one thing that I particularly liked about the existing version is that the output displayed the table name itself, not an integer value.  Since there is no simple/fast way to look up the name from a given enum value (I do not consider having to open `arkode_butcher_erk.h` and manually count lines inside the enum as either simple or fast), this update obfuscates the connection between method performance and the name itself.

test/unit_tests/arkode/CXX_serial/ark_test_butcher.cpp
```diff
@@ -21,76 +21,43 @@
...
+     {"ARKODE_ARK324L2SA_4_2_3", "ARKODE_ARK324L2SA_ERK_4_2_3",
+      "ARKODE_ARK324L2SA_DIRK_4_2_3"},
+     {"ARKODE_ARK436L2SA_6_3_4", "ARKODE_ARK436L2SA_ERK_6_3_4",
+      "ARKODE_ARK436L2SA_DIRK_6_3_4"},
+     {"ARKODE_ARK437L2SA_7_3_4", "ARKODE_ARK437L2SA_ERK_7_3_4",
+      "ARKODE_ARK437L2SA_DIRK_7_3_4"},
+     {"ARKODE_ARK548L2SA_8_4_5", "ARKODE_ARK548L2SA_ERK_8_4_5",
+      "ARKODE_ARK548L2SA_DIRK_8_4_5"},
+     {"ARKODE_ARK548L2SAb_8_4_5", "ARKODE_ARK548L2SAb_ERK_8_4_5",
+      "ARKODE_ARK548L2SAb_DIRK_8_4_5"}};
```
drreynolds, line 48: This piece is no more easily "maintained" than what was there before, as any new ARK table will still require manual intervention to add it here.

drreynolds:
> Is there any need to do outfile checking for this test? Isn't the return code of the test enough?

That's a good question, since I also do not fully understand the use of `.out` files versus return flags in `test/unit_tests`.

test/unit_tests/arkode/CXX_serial/ark_test_butcher.cpp
Steven-Roberts: Yeah the output is not as informative. The function `arkButcherTableERKNameToID` does allow us to convert an ID to a name. I'll check if that can be accessed here to keep the original output style.

Steven-Roberts: I agree the output is not ideal.

> Since there is no simple/fast way to look up the name from a given enum value

I can create the reserve function of [`arkButcherTableERKNameToID`](https://github.com/LLNL/sundials/blob/b685654c6958699821fc68a822b66076fec8cae7/src/arkode/arkode_butcher_erk.c#L62). Then we could map the id from the loop to its name and have the original output style.

test/unit_tests/arkode/CXX_serial/ark_test_butcher.cpp
Steven-Roberts: Yeah, there not much that we can do for ARK methods. That said, I don't think this should discourage us in the ERK/DIRK case.

drreynolds (dismissed):
Once this passes the CI tests (it may need a new .out file, either in this repo or in the answers repo), this looks good to me.

balos1 (approved):

gardner48:
@Steven-Roberts A couple of follow-on items to this PR:
* Note the addition of `ARKodeButcherTable_ERKIDToName` and `ARKodeButcherTable_DIRKIDToName` in the change log and recent changes
* Add `.. versionadded: x.y.z` to the documentation for `ARKodeButcherTable_ERKIDToName` and `ARKodeButcherTable_DIRKIDToName`
* Update the `test/answers` commit to `2004a4042d8f730073bac084d601e083fca0c5bf`

gardner48:
> * Update the `test/answers` commit to `2004a4042d8f730073bac084d601e083fca0c5bf`

Actually, no need to update `test/answers` the commit will be updated by #474

Steven-Roberts:
Ok, looks like I'll need a separate PR to do that now that this is merged.

balos1:
> .. versionadded: x.y.z

I always forget about this nice sphinx feature

PR #492, opened 2024-05-21
CI (MinGW): Add runner for 32-bit (MINGW32)

mmuetzel:
32-bit and 64-bit Windows systems are different to 32-bit and 64-bit Linux systems, e.g., when it comes to the data model.
Add a runner to the CI that builds for Windows 32-bit for coverage.


balos1 (approved):
Thanks!

PR #495, opened 2024-05-23
Feature: Remove Implicit RHS Evaluation in Autonomous Problems With the Trivial Predictor

gardner48:
Add the function ``ARKodeSetAutonomous`` to indicate that the implicit right-hand side function does not explicitly depend on time. When using the trivial predictor, an autonomous problem may reuse implicit function evaluations across stage solves reducing the total number of function evaluations.


gardner48:
Opening draft PR for testing

doc/arkode/guide/source/Usage/User_callable.rst
```diff
@@ -1824,6 +1825,49 @@ Specify if the implicit RHS is deduced after a nonlinear solve  :c:func:`ARKodeS
...
+   :retval ARK_MEM_NULL: ``arkode_mem`` was ``NULL``.
+   :retval ARK_ILL_INPUT: an argument had an illegal value.
+   :retval ARK_STEPPER_UNSUPPORTED: implicit solvers are not supported by the
+                                    current time-stepping module.
+
+   .. warning::
+
+      Results may differ when combining :c:func:`ARKodeSetAutonomous` with
+      :c:func:`ARKodeSetDeduceImplicitRhs` and using a stiffly accurate method
+      with the trivial predictor as the deduced implicit right-hand side (RHS)
```
drreynolds, line 1853: This sentence was overly long and difficult to digest.  
```suggestion
      with the trivial predictor.  This is because the deduced implicit right-hand side (RHS)
```

doc/arkode/guide/source/Usage/User_callable.rst
```diff
@@ -1824,6 +1825,49 @@ Specify if the implicit RHS is deduced after a nonlinear solve  :c:func:`ARKodeS
...
+      value will be reused in the initial residual computation rather than
+      evaluating the implicit RHS function e.g.,
+      ``examples/arkode/C_serial/ark_brusselator.c``.
```
drreynolds, line 1856: I'm not sure that I understand why a code example is used to support this statement.  Perhaps some explanation here would help, such as
```suggestion
      evaluating the implicit RHS function.  This issue can be seen in the example problem 
      ``examples/arkode/C_serial/ark_brusselator.c``, where (explain what happens).
```

doc/arkode/guide/source/Usage/User_callable.rst
```diff
@@ -1824,6 +1825,49 @@ Specify if the implicit RHS is deduced after a nonlinear solve  :c:func:`ARKodeS
...
+
+      Similarly programs that save computations within the RHS function for
+      reuse elsewhere will need to be updated to account for the RHS function
+      reuse with :c:func:`ARKodeSetAutonomous`. For example,
+      ``examples/arkode/C_serial/ark_KrylovDemo_prec.c`` reuses rates computed
+      in the RHS function in the preconditioner setup. If
+      :c:func:`ARKodeSetAutonomous` was enabled in this example, the rates saved
+      initially would be from the RHS function evaluations used to determine the
+      initial step size rather than the RHS evaluation in the nonlinear residual
+      with the predictor.
```
drreynolds, line 1866: I anticipate that this statement will just confuse users.  How are they to know that they are reusing too-old of RHS values in their routines, and is this really a problem?  Some clear guidance on this is needed.  

For example, "Care must be taken in user-supplied functions that utilize stored values of the implicit RHS function (:c:type:`ARKLsJacFn`, :c:func:`ARKLsLinSysFn`, :c:func:`ARKLsJacTimesVecFn`, :c:func:`ARKLsJacTimesSetupFn`, :c:func:`ARKLsPrecSetupFn`, :c:func:`ARKLsPrecSolveFn`).  In these functions, the input vector of RHS function evaluations may come from a significantly earlier phase of the simulation than the current implicit solve (e.g., determination of the initial step size).  For problems where this may be problematic, we recommend that the user-supplied function re-evaluate the implicit RHS function internally when using :c:func:`ARKodeSetAutonomous`."

doc/shared/RecentChanges.rst
```diff
@@ -36,6 +36,12 @@ Added :c:func:`ARKodeButcherTable_ERKIDToName` and
 :c:func:`ARKodeButcherTable_DIRKIDToName` to convert a Butcher table ID to a
 string representation.
 
+Added the function :c:func:`ARKodeSetAutonomous` in ARKODE to indicate that the
+implicit right-hand side function does not explicitly depend on time. When using
+the trivial predictor, an autonomous problem may reuse implicit function
+evaluations across stage solves reducing the total number of function
+evaluations.
```
drreynolds, line 43:
```suggestion
evaluations across stage solves to reduce the total number of function
evaluations.
```

CHANGELOG.md
```diff
@@ -38,6 +38,12 @@ Added the following MRI coupling tables
 Added `ARKodeButcherTable_ERKIDToName` and `ARKodeButcherTable_DIRKIDToName` to
 convert a Butcher table ID to a string representation.
 
+Added the function ``ARKodeSetAutonomous`` in ARKODE to indicate that the
+implicit right-hand side function does not explicitly depend on time. When using
+the trivial predictor, an autonomous problem may reuse implicit function
+evaluations across stage solves reducing the total number of function
+evaluations.
```
drreynolds, line 45:
```suggestion
evaluations across stage solves reducing the total number of function
evaluations.
```
```suggestion
evaluations across stage solves to reduce the total number of function
evaluations.
```

examples/arkode/C_serial/ark_brusselator.out
```diff
@@ -8,24 +8,24 @@ Brusselator ODE test problem:
...
+   10.000000    0.305598    3.657381    3.499995
    -------------------------------------------
 
 Final Solver Statistics:
-   Internal solver steps = 250 (attempted = 262)
-   Total RHS evals:  Fe = 0,  Fi = 3282
-   Total linear solver setups = 118
+   Internal solver steps = 252 (attempted = 262)
+   Total RHS evals:  Fe = 0,  Fi = 1910
+   Total linear solver setups = 111
```
drreynolds, line 25: Do you have any idea as to why this problem now takes more time steps and more overall Newton iterations?

src/arkode/arkode_arkstep_io.c
```diff
@@ -781,6 +790,51 @@ int arkStep_SetNonlinear(ARKodeMem ark_mem)
...
+  if (autonomous)
+  {
+    /* Allocate tempv5 if needed */
+    if (!arkAllocVec(ark_mem, ark_mem->yn, &ark_mem->tempv5))
+    {
+      arkProcessError(ark_mem, ARK_MEM_FAIL, __LINE__, __func__, __FILE__,
+                      MSG_ARK_MEM_FAIL);
+      return ARK_MEM_FAIL;
+    }
+  }
```
drreynolds, line 833: Should this routine also free `ark_mem->tempv5` in the case that `autonomous` is false, e.g., if a user called this twice, first with `autonomous` set to true and then with `autonomous` set to false?

src/arkode/arkode_arkstep.c
```diff
@@ -1729,23 +1740,88 @@ int arkStep_TakeStep_Z(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
...
-  /* Call the full RHS if needed e.g., an explicit first stage. If this is the
-     first step then we may need to evaluate or copy the RHS values from an
-     earlier evaluation (e.g., to compute h0). For subsequent steps treat this
-     RHS evaluation as an evaluation at the end of the just completed step to
-     potentially reuse (FSAL methods) or save (stiffly accurate methods with an
-     implicit first stage using Hermite interpolation) RHS evaluations from the
-     end of the last step. */
+  /* For a stiffly accurate implicit or ImEx method with an implicit first
+     stage, save f(tn, yn) if using Hermite interpolation as Fi[0] will be
+     overwritten during the implicit solve */
```
drreynolds, line 1745: The "complexity" of this function is growing out of control.  I'd like us as a group to think diligently about what we can do to simplify this.  Some options include:
* Create multiple `TakeStep` functions that are more specialized to specific purposes.
* Pull out distinct pieces of `TakeStep` into separate functions, each of which can encapsulate the necessary logic.  We could even have some ARKSTEP-specific internal function pointers, with separate implementations that handle simple-vs-complex situations.  For example, we could pull out everything that happens before we loop over stages into a `StepSetup` function, with separate implementations for ERK, DIRK, and ARK methods.
* Create more `sunbooleantype` flags that we determine during initialization, and that are used instead of re-determining so many things at each step.
* Deprecating some of our less-used options that necessitate so many corner cases.
* Reorganize ARKSTEP so that it is optimal for the default/simplest case (e.g., trivial predictor with autonomous RHS and Lagrange interpolant), and then do additional things only when doing something more complicated (e.g., higher-order predictors).

doc/arkode/guide/source/Usage/User_callable.rst
gardner48: > input vector of RHS function evaluations may come from a significantly earlier phase of the simulation than the current implicit solve

The RHS input to the functions listed will still be consistent with the state input to these functions.

The problem in the KrylovDemo example is that the "user" assumes that the nonlinear residual will be called and will evaluate the RHS before the preconditioner setup is called. Based on this assumption, the user's RHS functions saves the rates computed in the RHS and reuses them in the preconditioner setup. Enabling `SetAutonomous` breaks this assumption. For this example, the solution is the user should not save these values but instead evaluate the rates with the state input to the preconditioner setup.

examples/arkode/C_serial/ark_brusselator.out
gardner48: This is the issue I noted in the documentation for `SetAutonomous` where the combination of a stiffly accurate method + deduce rhs + set autonomous may lead to slightly different solutions. 

When `SetAutonomous` is enabled the first Newton iteration uses the deduced RHS evaluation in the residual rather than re-evaluating the RHS because the method is stiffly accurate. This difference between the deduced RHS and evaluating the RHS leads to slightly different solutions over time.

src/arkode/arkode_arkstep_io.c
gardner48: Yes, it would be safe to free that vector here.

src/arkode/arkode_arkstep.c
gardner48: > The "complexity" of this function is growing out of control. I'd like us as a group to think diligently about what we can do to simplify this. Some options include:
> 
>     * Create multiple `TakeStep` functions that are more specialized to specific purposes.

Having separate `TaksStep` functions for ERK and ARK/DIRK would help make the ERK case simpler. It's possible for a user to switch from an ERK to DIRK table during the integration so we'd need to update the `TakeStep` pointer when ever a method is attached.

>     * Pull out distinct pieces of `TakeStep` into separate functions, each of which can encapsulate the necessary logic.  We could even have some ARKSTEP-specific internal function pointers, with separate implementations that handle simple-vs-complex situations.  For example, we could pull out everything that happens before we loop over stages into a `StepSetup` function, with separate implementations for ERK, DIRK, and ARK methods.


I like the idea of having a `PreStep` function (and maybe a `PostStep` function too) to encapsulate this logic. 

>     * Create more `sunbooleantype` flags that we determine during initialization, and that are used instead of re-determining so many things at each step.

Many of these boolean values could be set a initialization but we'd need a flag to update these values any time a relevant option is changed e.g., if the method or predictor is changed during the integration. This could be forced with `ReInit` but that's does more than just validate the current settings.

>     * Deprecating some of our less-used options that necessitate so many corner cases.

I'm not sure I'd want to deprecate any of the options related to this logic since they could all reduce RHS evaluations. 

>     * Reorganize ARKSTEP so that it is optimal for the default/simplest case (e.g., trivial predictor with autonomous RHS and Lagrange interpolant), and then do additional things only when doing something more complicated (e.g., higher-order predictors).

I think the `PreStep` function is a better solution since the differences are before the actual step computations which are the same between the different configurations.

doc/arkode/guide/source/Usage/User_callable.rst
gardner48: See [this comment below](https://github.com/LLNL/sundials/pull/495#discussion_r1613540758). I'll update this text with more details.

gardner48: I've updated this text (and added a comment in the example code) to clarify how this example be used to demonstrate the differences that can arise when combining `SetAutonomous` and `SetDedeuceImplicitRhs` with a stiffly accurate method.

doc/arkode/guide/source/Usage/User_callable.rst
```diff
@@ -1848,22 +1848,31 @@ Specify if the implicit RHS is deduced after a nonlinear solve  :c:func:`ARKodeS
...
-      :c:func:`ARKodeSetAutonomous` was enabled in this example, the rates saved
-      initially would be from the RHS function evaluations used to determine the
-      initial step size rather than the RHS evaluation in the nonlinear residual
-      with the predictor.
+      Results may differ when enabling both :c:func:`ARKodeSetAutonomous` and
+      :c:func:`ARKodeSetDeduceImplicitRhs` with a stiffly accurate implicit
+      method and using the trivial predictor. The differences are due to reusing
+      the deduced implicit right-hand side (RHS) value in the initial nonlinear
+      residual computation rather than evaluating the implicit RHS function. The
+      significance of the difference well depend on how well the deduced RHS
```
drreynolds:
```suggestion
      significance of the difference will depend on how well the deduced RHS
```

drreynolds (dismissed):

test/unit_tests/arkode/CXX_serial/ark_test_dahlquist_ark.cpp
```diff
@@ -779,10 +742,40 @@ int expected_rhs_evals(ProblemOptions& prob_opts, int stages, int order,
...
+  bool save_fn_for_residual = prob_opts.p_type == 0 &&
+                              prob_opts.m_type !=
+                                mass_matrix_type::time_dependent &&
+                              prob_opts.r_type != rk_type::expl;
+
   if (prob_opts.r_type == rk_type::impl || prob_opts.r_type == rk_type::imex)
   {
     flag = ARKodeGetNumNonlinSolvIters(arkstep_mem, &nni);
-    if (check_flag(&flag, "ARKodeGetNumNonlinSolvIters", 1)) { return 1; }
+    if (check_flag(&flag, "ARKStepGetNumNonlinSolvIters", 1)) return 1;
```
balos1, line 753:
```suggestion
    if (check_flag(&flag, "ARKodeGetNumNonlinSolvIters", 1)) { return 1; }
```

balos1 (changes requested):

balos1 (approved):

PR #498, opened 2024-05-31
Add macro for consistent sunrealtype formatting

Steven-Roberts:
Putting in draft stage since there's many more places to add the macro, but I want to get the ok before proceeding. Is `SUN_REAL_TYPE_FORMAT` preferable to `SUN_REAL_FORMAT`? Or `SUN_FORMAT_REAL`?

Example of the formatting used: https://godbolt.org/z/EEjderKab


src/cvode/cvode_impl.h
```diff
@@ -33,11 +33,9 @@ extern "C" {
 #endif
 
 #if defined(SUNDIALS_EXTENDED_PRECISION)
-#define RSYM  ".32Lg"
-#define RSYMW "19.32Lg"
+#define RSYM ".32Lg"
```
Steven-Roberts, line 36: This `RSYM` macro is duplicated a few times, so I'm planning replace it with the single `SUN_REAL_FORMAT`.

gardner48: Take a look at [this branch ](https://github.com/LLNL/sundials/tree/feature/output-format). I'm doing something similar there to clean up `PrintAllStats`. We should coordinate on updating the formatting.

src/sundials/sundials_utils.h
```diff
@@ -86,4 +89,56 @@ static inline void sunCompensatedSum(sunrealtype base, sunrealtype inc,
...
+    fprintf(fp, "%-*s = %ld\n", SUN_TABLE_WIDTH, name, value[0]);
+    for (size_t i = 1; i < count; i++) { fprintf(fp, ", %ld", value[i]); }
+    fprintf(fp, "\n");
+  }
+  else
+  {
+    if (!start) { fprintf(fp, ","); }
+    for (size_t i = 0; i < count; i++)
+    {
+      fprintf(fp, "%s %zu,%ld", name, i, value[i]);
```
Steven-Roberts, line 139: Note this format is slightly different than
https://github.com/LLNL/sundials/blob/b685654c6958699821fc68a822b66076fec8cae7/src/cvodes/cvodes_io.c#L2745

src/sundials/fmod_int64/fsundials_core_mod.f90
```diff
@@ -47,6 +47,8 @@ module fsundials_core_mod
  integer(C_INT), parameter, public :: SUN_COMM_NULL = 0_C_INT
 #endif
 
+ type(C_PTR), parameter, public :: SUN_REAL_FORMAT_E = "%.DBL_DIGe"_C_PTR
```
Steven-Roberts, line 50: Looks like swig is having trouble converting this macro. Is there a way to exclude this?

PR #499, opened 2024-05-31
Add a Python library for developer utilities with an initial module for log parsing

balos1:
I started this on another branch where I needed to parse the `SUNLogger` log files to understand (and plot) some CVode behavior.  The intent is to add more of our Python utilities for plotting stuff like step size over time etc.  We can also add other modules unrelated to logging.

### DIFF
--- a/doc/superbuild/source/developers/sundialsdev/index.rst
+++ b/doc/superbuild/source/developers/sundialsdev/index.rst
@@ -1,0 +1,23 @@
+..
+   -----------------------------------------------------------------------------
+   SUNDIALS Copyright Start
+   Copyright (c) 2002-2024, Lawrence Livermore National Security
+   and Southern Methodist University.
+   All rights reserved.
+
+   See the top-level LICENSE and NOTICE files for details.
+
+   SPDX-License-Identifier: BSD-3-Clause
+   SUNDIALS Copyright End
+   -----------------------------------------------------------------------------
+
+.. _SUNDIALS_DEV:
+
+sundialsdev Python Library
+==========================
+
+This is a Python library of utilities SUNDIALS developer may find useful.
+Right now it consists of the following modules:
+
+- ``logs``: this module has functions for parsing logs produced by `SUNLogger`.
+
--- a/scripts/sundialsdev/logs.py
+++ b/scripts/sundialsdev/logs.py
@@ -1,0 +1,120 @@
+#!/usr/bin/env python3
+# -----------------------------------------------------------------------------
+# Programmer(s): Cody Balos @ LLNL
+# -----------------------------------------------------------------------------
+# SUNDIALS Copyright Start
+# Copyright (c) 2002-2024, Lawrence Livermore National Security
+# and Southern Methodist University.
+# All rights reserved.
+#
+# See the top-level LICENSE and NOTICE files for details.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+# SUNDIALS Copyright End
+# -----------------------------------------------------------------------------
+# Module of Python functions that may be useful to SUNDIALS developers writing
+# scripts to parse logs produced by SUNLogger.
+# -----------------------------------------------------------------------------
+
+import re
+import numpy as np
+
+
+def parse_logfile_payload(payload, line_number, all_lines, array_indicator="(:)"):
+    """
+    This function parses the payload of in a SUNDIALS log file line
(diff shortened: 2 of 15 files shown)


scripts/sundialsdev/logs.py
```diff
@@ -0,0 +1,120 @@
...
+        line_dict["label"] = matches[0][3]
+        line_dict["payload"] = parse_logfile_payload(
+            matches[0][4], line_number, all_lines
+        )
+    return line_dict
+
+
+def log_file_to_list(filename, step_scope_txt):
+    """
+    This function takes a debug log file from CVODE and creates a list where
```
drreynolds, line 77:
```suggestion
    This function takes a debug log file from a SUNDIALS log file and creates a list where
```
(or is this function specific to CVODE)?

scripts/sundialsdev/logs.py
```diff
@@ -0,0 +1,120 @@
...
+    each list entry is a step attempt.
```
drreynolds, line 78:
```suggestion
    each list entry is an element of an array.
```

drreynolds (commented):
I like where this is going.  Are you confident that all SUNLogger lines consistently use the required format, i.e., that single items have the format
```
[loglvl][rank][scope][label] key1 = value, key2 = value
```
and array-valued items have the format
```
[loglvl][rank][scope][label] y(:)
y_1
y_2
...
```
?  It looks like you fixed one inconsistency in CVODE, but I think that all such inconsistencies should be resolved before this PR is eventually merged.

balos1:
>  Are you confident that all SUNLogger lines consistently use the required format
I am pretty confident that all single line entires do follow the required format. The vector valued ones do not yet (only CVODE where I updated it). 

> It looks like you fixed one inconsistency in CVODE, but I think that all such inconsistencies should be resolved before this PR is eventually merged.

My intent was for this PR to just put up what I have developed already as part of a different task. Others can extend it, and update lines in the other packages as necessary.  If I need to resolve all inconsistencies in the packages before merging this, I will not be able to finish up this PR for a bit. Since its just a developer tool, perhaps we can accept it as-is and others can fix things they find as needed?

scripts/sundialsdev/logs.py
balos1: The intent of the comment was to convey that each element of the list (i.e., each list entry) represents a step attempt.  Ill reword the comment like this:

> This function takes a debug log file from CVODE and creates a list where each list element represents an integrator step attempt.

balos1:
```suggestion
    each list element represents an integrator step attempt.
```

scripts/sundialsdev/logs.py
balos1: It should not be specific to CVODE (I refactored and this comment should be updated as you point out).

scripts/sundialsdev/logs.py
gardner48, line 1: I think we should move this file and `sundials_csv.py` to a new `tools` or `utilities` directory to separate them for the other files in `scripts` where are for development/release.

doc/superbuild/source/developers/sundialsdev/index.rst
gardner48, line 1: These utilities would be helpful to users as well. I think this should go in the user guide with some documentation on the functions provided.

scripts/sundialsdev/logs.py
balos1: If you mean move `scripts/sundialsdev` to `tools/sundialsdev` I think that would be good.

doc/superbuild/source/developers/sundialsdev/index.rst
balos1: Should we do that here or in #501?

scripts/sundialsdev/logs.py
gardner48: I was thinking just under `tools`. Doing that in #501 is good with me

doc/superbuild/source/developers/sundialsdev/index.rst
gardner48: Doing is in #501 is good with me.

scripts/sundialsdev/logs.py
balos1: I think it should remain under `sundialsdev` because it is setup to be a Python library that would be installable as `sundialsdev`.

balos1:
I think this is ready now @gardner48. The last two comments will be taken care of in #501.

gardner48 (approved):

PR #501, opened 2024-06-08
Feature: Logging Updates

gardner48:
Follow on to utilities added in #499 
* Make info logging output in the integrators more consistent 
* Update parser for MRI methods
* Add function to extract lists of data for plotting
* Add example scripts using log parser


gardner48:
> Checkout https://github.com/LLNL/sundials/blob/feature/nls-switching-v7-gs/src/sundials/sundials_logger_impl.h#L35. Now that we have C99 we can change all the logging statements in the code to use a variadic macro and it makes the logging statements much cleaner and gets rid of the repetitive preprocessor checks. My intent was to bring this over from that branch. I can do that here if you think it looks good.

I was going to ask if you still had the variadic macros from the original logging implementation on another branch. I think switching over to those would be great. 

It would also be nice to have an "If" version of the macros i.e., something like `SUNLogInfoIf` that takes an additional boolean input to change cases like
```
#if SUNDIALS_LOGGING_LEVEL >= SUNDIALS_LOGGING_INFO
      if (kflag != ARK_SUCCESS)
      {
        SUNLogInfo(ARK_LOGGER, __func__, "end-step-attempt", "status = failed solve, kflag = %i", kflag);
      }
#endif
```
into something like
```
SUNLogInfoIf(kflag != ARK_SUCCESS, ARK_LOGGER __func__, "end-step-attempt", "status = failed solve, kflag = %i", kflag);
```

balos1:
@gardner48 The "if" version sounds like a nice addition as well.

scripts/sundialsdev/logs.py
```diff
@@ -72,13 +82,12 @@ def parse_logfile_line(line, line_number, all_lines):
...
-def log_file_to_list(filename, step_scope_txt):
+def log_file_to_list(filename):
     """
-    This function takes a debug log file from a SUNDIALS log file and creates a list where
-    each list element represents an integrator step attempt.
+    This function takes a SUNDIALS log file and creates a list where each list
+    element represents an integrator step attempt.
 
     E.g.,
-    [
```
balos1, line 81: This function no longer captures all the log entries between "begin-step-attempt" and "end-step-attempt". If that was your intent, then it shuold be a different function.

balos1: This function was specifically to capture *all* log output and organize into a list of lists, where each inner list holds all the logging output lines (each one as a dictionary) for a step attempt.

balos1: For a nested integrator, I think it should be like:
```
[
  // first level is outer steps
   [
      // outer step 1
     { stuff },
     { stuff },
     [ 
        // inner step
     ],
   ],
]
```

gardner48: The intent is to handle all the entries but I haven't gotten there yet (working on making it easier to parse/plot the nonlinear/linear solver logging first). 

I originally tried the structure above for MRI output but it was difficult to associate the fast integrations with their corresponding stages at the slow time scale. The structure I ended up going with is a nesting lists of step attempt data with in lists of stage data.

For example an RK method the list is
```
[
  {
    'step' : 1,
    't_n' : 0,
    ...
    'stages' :
    [
      { 
         'stage' : 0
         ...
      },
      {
         'stage' : 1,
         ...
      }
    ],
   'dsm' : 2.363391824501069e-08
  },
 ...
]     
```
For an MRI method each dictionary in the `stages` list has a `time-level-#` entry that is a list of step attempt dictionaries just like above.

test/unit_tests/logging/CMakeLists.txt
```diff
@@ -0,0 +1,143 @@
...
+
+if(BUILD_IDA)
+  list(APPEND unit_tests "test_logging_ida.cpp\;1") # Newton + Dense
+  list(APPEND unit_tests "test_logging_ida.cpp\;0") # Newton + GMRES
+endif()
+
+if(BUILD_IDAS)
+  list(APPEND unit_tests "test_logging_idas.cpp\;1") # Newton + Dense
+  list(APPEND unit_tests "test_logging_idas.cpp\;0") # Newton + GMRES
+endif()
```
drreynolds, line 72: This is a *ton* of new tests; multiple of which seem redundant.  I'm unclear on whether the CI or Jenkins will run all examples/tests for all logging levels (and all precisions), but that seems to be even more redundant.

Would it be possible to get full "coverage" with a smaller set of tests so that we don't overwhelm the CI bots or Jenkins?  Could these just be run at the highest logging level (since that output includes all of the lower logging levels)?

drreynolds (changes requested):
This PR is very exciting!

I have two documentation-related requests, both in regard to the as-yet-untouched file `doc/shared/sundials/Logging.rst` (which is why I'm mentioning them here):

1. Add the integer values that correspond with each logging level (`SUN_LOGLEVEL_ALL`, `SUN_LOGLEVEL_NONE`, ...).  Currently, users are asked for a numerical logging level value by CMake, which can only be found hidden inside `include/sundials/sundials_logger.h`.
2. The end of this file has an incredibly brief section on "Example Usage" for the logger.  Could you add a few step-by-step "case studies" showing how a user could 
     * configure SUNDIALS with an "appropriate" logging level
     * set environment variables to hold the "relevant" log file(s)
     * run an existing SUNDIALS example
     * use one of `tools/log_example.py` or `tools/log_example_mri.py`, e.g., to plot the step size history as a function of time.

scripts/tarscript
```diff
@@ -271,7 +271,7 @@ find $tmpdir -name ".DS_Store" -delete
 
 # Remove ignored or untracked files that may have been added
 cd $sundialsdir
-for f in $(git ls-files --others --directory benchmarks cmake doc examples include scripts src test); do
+for f in $(git ls-files --others --directory benchmarks cmake doc examples external include src test tools); do
```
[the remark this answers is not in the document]
gardner48: `external` is still needed (it was moved rather than removed above so the file order matches the `ls` order).

tools/suntools/logs.py
```diff
@@ -0,0 +1,366 @@
...
+    Parse a line from a SUNDIALS log file it into a dictionary.
+
+    A log file line has the form:
+      [loglvl][rank][scope][label] key1 = value, key2 = value
+    The log line payload (everything after the brackets) can be multiline with
+    one value per line for keys corresponding to an array/vector:
+      [loglvl][rank][scope][label] y(:) =
+      y_1
+      y_2
+      ...
```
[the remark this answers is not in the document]
gardner48: Yes, `(:)` is needed to detect a vector output. Hopefully the merge will just be an easy "keep the changes here"

src/arkode/arkode_mristep.c
```diff
@@ -1759,8 +1743,8 @@ int mriStep_SetCoupling(ARKodeMem ark_mem)
     {
     case 1: table_id = MRISTEP_DEFAULT_IMEX_SD_1; break;
     case 2: table_id = MRISTEP_DEFAULT_IMEX_SD_2; break;
-    case 3: table_id = MRISTEP_DEFAULT_IMEX_SD_2; break;
-    case 4: table_id = MRISTEP_DEFAULT_IMEX_SD_2; break;
+    case 3: table_id = MRISTEP_DEFAULT_IMEX_SD_3; break;
+    case 4: table_id = MRISTEP_DEFAULT_IMEX_SD_4; break;
     }
```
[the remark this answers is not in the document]
gardner48: I needed this fix for testing so I'll keep the change here and resolve the merge conflict when the other PR is merged

test/unit_tests/logging/CMakeLists.txt
gardner48: Unfortunately, this is as lean as I think the set of tests can be. The different logging levels need to be tested separately to ensure output from a higher level is not accidentally output at a lower level e.g., debugging information printed when only logging errors. The GitHub logging test (`.github/workflows/ubuntu-clang-latest.yml`) was updated to build with each logging level and run only these logging tests (this PR also adds a `LABELS` option to `sundials_add_test` so we can run only the tests with a given label) to keep the testing time down. Testing each logging level takes about 3 minutes and can be run in parallel to each other. The other CI tests are unchanged and only run with the default logging level. The different combinations of intergrator/stepper and linear/nonlinear solver used I believe are the minimal set needed to cover the code different pathways and logging output nesting levels.

gardner48:
>     1. Add the integer values that correspond with each logging level (`SUN_LOGLEVEL_ALL`, `SUN_LOGLEVEL_NONE`, ...).  Currently, users are asked for a numerical logging level value by CMake, which can only be found hidden inside `include/sundials/sundials_logger.h`.

The [logging documentation](https://sundials--501.org.readthedocs.build/en/501/sundials/Logging_link.html#enabling-logging) directs users to the documentation for the CMake option [here](https://sundials--501.org.readthedocs.build/en/501/sundials/Install_link.html#cmakeoption-SUNDIALS_LOGGING_LEVEL) which lists the numeric values and associated logging level.
 
>     2. The end of this file has an incredibly brief section on "Example Usage" for the logger.  Could you add a few step-by-step "case studies" showing how a user could
>        
>        * configure SUNDIALS with an "appropriate" logging level
>        * set environment variables to hold the "relevant" log file(s)
>        * run an existing SUNDIALS example
>        * use one of `tools/log_example.py` or `tools/log_example_mri.py`, e.g., to plot the step size history as a function of time.

I'll update that second with additional information shortly.

test/unit_tests/logging/CMakeLists.txt
drreynolds: That sounds good; I was concerned that the total number of tests might multiply by a factor of 5, but with these labeled appropriately it isn't onerous.

test/unit_tests/logging/test_logging_cvode.cpp
```diff
@@ -0,0 +1,195 @@
...
+ * ---------------------------------------------------------------------------*/
+
+#include <cmath>
+#include <cstdio>
+#include <iomanip>
+#include <iostream>
+#include <limits>
+
+// Include desired integrators, vectors, linear solvers, and nonlinear solvers
+#include "cvode/cvode.h"
```
Steven-Roberts, line 24: The CVODES version of this essentially only differs in which header is used. You could factor this shared code into a separate file `test_logging_cvode_shared.h`, removing the CVODE(S) include.
I think this file would simply be
```
#include "cvode/cvode.h"
#include "test_logging_cvode_shared.h"
```
while the CVODES one would be
```
#include "cvodes/cvodes.h"
#include "test_logging_cvode_shared.h"
```
That would save nearly 400 lines of duplicated code across the CVODE and IDA tests. x-macros can be used to have the print statements for the method name be different if desired.

tools/log_example.py
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env python3
```
Steven-Roberts, line 1: In this `tools` directory, can we have a `requirements.txt` file or some way to get dependencies?

tools/log_example.py
```diff
@@ -0,0 +1,163 @@
...
+
+    parser.add_argument('--logy', action='store_true',
+                        help='Use log scale for y-axis')
+
+    parser.add_argument('--labels', type=str, nargs='+',
+                        help='Plot labels')
+
+    parser.add_argument('--save', type=str, nargs='?', const='fig.pdf',
+                        default=None, metavar='FILE_NAME',
+                        help='''Save figure to file''')
```
Steven-Roberts, line 61:
```suggestion
                        help='Save figure to file')
```

tools/suntools/logs.py
```diff
@@ -0,0 +1,366 @@
...
+    try:
+        return np.longlong(s)
+    except ValueError:
+        try:
+            return np.double(s)
+        except ValueError:
+            return s
+
+
+def parse_logfile_payload(payload, line_number, all_lines, array_indicator="(:)"):
```
Steven-Roberts, line 35: I think the current implementation is perfectly fine, but in the future if log files become more complex, something like [pyparsing](https://github.com/pyparsing/pyparsing) would be worth considering

Steven-Roberts (changes requested):

test/unit_tests/logging/test_logging_cvode.cpp
Steven-Roberts: On second thought x-macros are probably not needed. Instead I suspect something like this works
```
#include "cvode/cvode.h"
static constexpr auto NAME = "CVODE";
#include "test_logging_cvode_shared.h"
```
and the shared header can use `NAME`.

src/arkode/arkode.c
```diff
@@ -874,23 +874,29 @@ int ARKodeEvolve(void* arkode_mem, sunrealtype tout, N_Vector yout,
...
       ark_mem->nst_attempts++;
 
-#if SUNDIALS_LOGGING_LEVEL >= SUNDIALS_LOGGING_DEBUG
-      SUNLogger_QueueMsg(ARK_LOGGER, SUN_LOGLEVEL_DEBUG, "ARKODE::ARKodeEvolve",
-                         "start-step",
-                         "step = %li, attempt = %i, h = %" RSYM
-                         ", tcur = %" RSYM,
-                         ark_mem->nst, attempts, ark_mem->h, ark_mem->tcur);
-#endif
+      SUNLogInfo(ARK_LOGGER, __func__, "begin-step-attempt",
```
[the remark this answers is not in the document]
gardner48: Yes, this can be incorporated into the macro.

src/sundials/sundials_logger_impl.h
```diff
@@ -32,6 +32,125 @@
 #define SUNDIALS_LOGGING_EXTRA_DEBUG
 #endif
 
+#if SUNDIALS_LOGGING_LEVEL >= SUNDIALS_LOGGING_ERROR
+#define SUNLogError(logger, scope, label, msg_txt, ...)                 \
+  SUNLogger_QueueMsg(logger, SUN_LOGLEVEL_ERROR, scope, label, msg_txt, \
```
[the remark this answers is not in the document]
gardner48: Is this portable or will some preprocessors produce `SUNLogger_QueueMsg(loger, SUN_LOGLEVER_ERROR, scope, label,)`?

test/unit_tests/logging/CMakeLists.txt
```diff
@@ -0,0 +1,143 @@
...
+
+  if("${test_args}" STREQUAL "")
+    set(test_name ${test_name})
+  else()
+    string(REPLACE " " "_" test_name "${test_name}_${test_args}")
+    string(REPLACE " " ";" test_args "${test_args}")
+  endif()
+
+  # For now, only diff when dev test are enabled
+  # TODO(DJG): Update Jenkins diff handling for testing single/extended
```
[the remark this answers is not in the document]
gardner48: This is a longer term `ToDo` as it will require changes to how diffs are enabled in CMake and the Jenkins/GitLab CI scripts.

test/unit_tests/logging/test_logging_cvode.cpp
gardner48: I'd like to keep them separate. The intent is expanded the CVODES version (and similarly for IDA) to cover the FSA and ASA cases in the future (along with improvements to the logging output with FSA and ASA).

tools/log_example.py
gardner48: The modules in `suntools` only require modules from the Python standard library so a requirements file is not needed to use the parsing tools. However, these examples that use the parsing capabilities and then make plots do require matplotlib, but I'm hesitant to include a requirements file here that's not need for `suntools`.

test/unit_tests/utilities/check_return.hpp
```diff
@@ -0,0 +1,36 @@
+/* -----------------------------------------------------------------------------
```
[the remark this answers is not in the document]
gardner48: Maybe, I think we pretty much want to halt on all cases in the unit tests so having a shared error handler function could do this. Let's add this enhancement to the backlog issues.

tools/log_example.py
Steven-Roberts: In that case, I think it's fine then.

src/sundials/sundials_logger_impl.h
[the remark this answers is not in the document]
Steven-Roberts: I'm not sure I understand what circumstances could cause the preprocessor to produce that. Apparently old MSVC versions had some buggy behavior with `__VA_ARGS__`, but Visual Studio 2015 handles it fine: https://godbolt.org/z/477aKTosE. I'm not aware of any clang or gcc issues.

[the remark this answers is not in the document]
gardner48: I was thinking of something like this https://godbolt.org/z/z48KWeEPT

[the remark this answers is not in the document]
Steven-Roberts: Every call to `SUNLogError` already has at least 4 arguments so this shouldn't be an issue. If for some reason a developer passed 3, it will raise an error since an argument is missing (which is a good and expected outcome).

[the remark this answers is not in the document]
Steven-Roberts: I guess with `scope` now removed as an argument, subtract those numbers by 1.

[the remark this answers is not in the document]
gardner48: I see, it's sort of hacking `__VA_ARGS__` to place the last argument. This gets nicer in C23 with the addition of `__VA_OPT__` to deal with the trailing comma problem.

doc/superbuild/source/developers/style_guide/SourceCode.rst
```diff
@@ -460,3 +442,204 @@ There are other scenarios (e.g., a function call with a lot of parameters) where
...
+   :param condition: a boolean expression that determines if the log message
+                     should be queued.
+   :param logger: the :c:type:`SUNLogger`
+   :param label: the ``const char*`` message label
+   :param vec: the ``N_Vector`` to print
+   :param msg_txt: the ``const char*`` message text, may contain format
+                   specifiers
+   :param ...: the arguments for format specifiers in ``msg_txt``
+
+.. c:macro:: SUNLogExtraDebugVecArray(logger, label, nvecs, vecs, msg_txt)
```
drreynolds, line 624: Do we also need `SUNLogExtraDebugVecArrayIf`?

Steven-Roberts (approved):

drreynolds (commented):
This looks great; I provided a few suggestions for minor typos, but otherwise this can be approved.

doc/superbuild/source/developers/style_guide/SourceCode.rst
```diff
@@ -355,24 +355,6 @@ not adhere to all of these rules.
    ``sunindextype`` for it. Instead use the appropriate integer type (e.g., ``uint64_t``) directly.
    Do not use ``sunindextype`` for counters either.
 
-#. ``SUNLogger`` statements must be in the format:
```
balos1, line 358: Add this bullet back but in brief (i.e. without the examples),  and reference the logging section for the details.

test/unit_tests/problems/kpr.hpp
```diff
@@ -0,0 +1,382 @@
...
+  const sunrealtype tmp2 = (-TWO + v * v - kpr_s(t)) / (TWO * v);
+
+  rdata[0] = (a * tmp1 + b * tmp2 + kpr_rdot(t) / (TWO * u)) - up;
+  rdata[1] = (c * tmp1 + d * tmp2 + kpr_sdot(t) / (TWO * v)) - vp;
+
+  return 0;
+}
+
+/* -----------------------------------------------------------------------------
+ * DAE RHS Jacobin:
```
balos1, line 352:
```suggestion
 * DAE RHS Jacobian:
```

balos1 (changes requested):
Looks great. Just a couple quick fixes needed.

balos1 (dismissed):

drreynolds (approved):

test/unit_tests/ida/CXX_serial/ida_test_kpr.cpp
```diff
@@ -193,7 +193,7 @@ int res(sunrealtype t, N_Vector y, N_Vector yp, N_Vector rr, void* user_data)
 }
 
 /* -----------------------------------------------------------------------------
- * Compute the ODE RHS Jacobin:
+ * Compute the DAE residual Jacobin:
```
balos1, line 196:
```suggestion
 * Compute the DAE residual Jacobian:
```

test/unit_tests/idas/CXX_serial/idas_test_kpr.cpp
```diff
@@ -193,7 +193,7 @@ int res(sunrealtype t, N_Vector y, N_Vector yp, N_Vector rr, void* user_data)
 }
 
 /* -----------------------------------------------------------------------------
- * Compute the ODE RHS Jacobin:
+ * Compute the DAE residual Jacobin:
```
balos1, line 196:
```suggestion
 * Compute the DAE residual Jacobian:
```

balos1 (changes requested):

src/arkode/arkode_lsrkstep.c
```diff
@@ -1611,28 +1673,28 @@ int lsrkStep_TakeStepSSP43(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
...
-                     "ARKODE::lsrkStep_TakeStepSSP43", "start-stage",
-                     "step = %li, stage = %i, h = %" RSYM ", tcur = %" RSYM,
-                     ark_mem->nst, 3, ark_mem->h, ark_mem->tcur + ark_mem->h);
-#endif
-
   retval = step_mem->fe(ark_mem->tcur + ark_mem->h, ark_mem->ycur,
                         ark_mem->tempv3, ark_mem->user_data);
   step_mem->nfe++;
+
+  SUNLogExtraDebugVec(ARK_LOGGER, "stage RHS", ark_mem->fn, "F_2(:) =");
```
drreynolds, line 1688: Printed the wrong vector.
```suggestion
  SUNLogExtraDebugVec(ARK_LOGGER, "stage RHS", ark_mem->tempv3, "F_2(:) =");
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1653,36 +1722,41 @@ int lsrkStep_TakeStepSSP43(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
...
-                     "step = %li, stage = %i, h = %" RSYM ", tcur = %" RSYM,
-                     ark_mem->nst, 4, ark_mem->h,
-                     ark_mem->tcur + ark_mem->h * p5);
-#endif
-
   retval = step_mem->fe(ark_mem->tcur + ark_mem->h * p5, ark_mem->ycur,
                         ark_mem->tempv3, ark_mem->user_data);
   step_mem->nfe++;
+
+  SUNLogExtraDebugVec(ARK_LOGGER, "stage RHS", ark_mem->fn, "F_3(:) =");
```
drreynolds, line 1737: Printed the wrong vector.
```suggestion
  SUNLogExtraDebugVec(ARK_LOGGER, "stage RHS", ark_mem->tempv3, "F_3(:) =");
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1790,37 +1848,35 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
...
+        return ARK_POSTPROCESS_STAGE_FAIL;
+      }
     }
 
     retval = step_mem->fe(ark_mem->tcur +
                             ((sunrealtype)j - ONE) * onesixth * ark_mem->h,
                           ark_mem->ycur, ark_mem->tempv3, ark_mem->user_data);
     step_mem->nfe++;
+
+    SUNLogExtraDebugVec(ARK_LOGGER, "stage RHS", ark_mem->fn, "F_%i(:) =", j - 1);
```
drreynolds, line 1870: Printed the wrong vector.
```suggestion
    SUNLogExtraDebugVec(ARK_LOGGER, "stage RHS", ark_mem->tempv3, "F_%i(:) =", j - 1);
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1841,33 +1902,31 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
...
-                       ark_mem->tcur +
-                         ((sunrealtype)j - FOUR) * onesixth * ark_mem->h);
-#endif
-
     retval = step_mem->fe(ark_mem->tcur +
                             ((sunrealtype)j - FOUR) * onesixth * ark_mem->h,
                           ark_mem->ycur, ark_mem->tempv3, ark_mem->user_data);
     step_mem->nfe++;
+
+    SUNLogExtraDebugVec(ARK_LOGGER, "stage RHS", ark_mem->fn, "F_%i(:) =", j);
```
drreynolds, line 1920: Printed the wrong vector.
```suggestion
    SUNLogExtraDebugVec(ARK_LOGGER, "stage RHS", ark_mem->tempv3, "F_%i(:) =", j);
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1889,30 +1948,26 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
...
-                     "ARKODE::lsrkStep_TakeStepSSP104", "start-stage",
-                     "step = %li, stage = %i, h = %" RSYM ", tcur = %" RSYM,
-                     ark_mem->nst, 10, ark_mem->h, ark_mem->tcur + ark_mem->h);
-#endif
-
   retval = step_mem->fe(ark_mem->tcur + ark_mem->h, ark_mem->ycur,
                         ark_mem->tempv3, ark_mem->user_data);
   step_mem->nfe++;
+
+  SUNLogExtraDebugVec(ARK_LOGGER, "stage RHS", ark_mem->fn, "F_10(:) =", 9);
```
drreynolds, line 1964: Printed the wrong vector.
```suggestion
  SUNLogExtraDebugVec(ARK_LOGGER, "stage RHS", ark_mem->tempv3, "F_10(:) =", 9);
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1921,6 +1976,17 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
...
+  if (retval != 0)
+  {
+    SUNLogInfo(ARK_LOGGER, "end-stage",
+               "status = failed vector op, retval = %i", retval);
+    return ARK_VECTOROP_ERR;
+  }
+
+  SUNLogInfo(ARK_LOGGER, "end-stage", "status = success");
+  SUNLogExtraDebugVec(ARK_LOGGER, "updated solution", ark_mem->ycur, "ycur(:) =");
+  SUNLogExtraDebugVec(ARK_LOGGER, "embedded solution", ark_mem->tempv1,
```
drreynolds, line 1988: I can't highlight enough lines to make the "suggestion" correctly, but this and the following line need to be moved just after the `N_VLinearSum` call in the following `if (!ark_mem->fixedstep)` block.

test/unit_tests/logging/test_logging_arkode_sprkstep.cpp
drreynolds, line 1: This test seems to be missing `.out` files in this PR (although the `CMakeLists.txt` file seems to include them).

tools/suntools/logs.py
```diff
@@ -0,0 +1,384 @@
...
+                    values.append(np.double(line))
+                kvp_dict[key.strip()] = values
+            else:
+                kvp_dict[key.strip()] = value.strip()
+    return kvp_dict
+
+
+def parse_logfile_line(line, line_number, all_lines):
+    """
+    Parse a line from a SUNDIALS log file it into a dictionary.
```
drreynolds, line 65:
```suggestion
    Parse a line from a SUNDIALS log file into a dictionary.
```

tools/suntools/logs.py
```diff
@@ -0,0 +1,384 @@
...
+            print(f"{spaces}{{")
+            print_log_dict(a_dict[key], indent=indent + 2)
+            print(f"{spaces}}}")
+        else:
+            print(f"{spaces}{key} : {a_dict[key]}")
+
+
+def print_log(log, indent=0):
+    """
+    Print the entries entries from a log file list of step attempts.
```
drreynolds, line 319:
```suggestion
    Print the entries from a log file list of step attempts.
```

drreynolds (changes requested):
This is looking great, but I found a few items to fix in the current implementation.

test/unit_tests/logging/test_logging_arkode_mristep.cpp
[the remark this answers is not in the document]
gardner48: I plan to add those was part of this PR as all the `TakeStep` functions are consistent. I just haven gotten to updating those functions yet after merging in the changes from develop. I'll ping for another review pass when I've finished updating everything.

src/arkode/arkode.c
```diff
@@ -912,13 +918,21 @@ int ARKodeEvolve(void* arkode_mem, sunrealtype tout, N_Vector yout,
       if (ark_mem->relax_enabled && (kflag == ARK_SUCCESS))
       {
         kflag = arkRelax(ark_mem, &relax_fails, &dsm);
+
+        SUNLogInfoIf(kflag != ARK_SUCCESS, ARK_LOGGER, "end-step-attempt",
+                     "status = failed relaxtion, kflag = %i", kflag);
+
         if (kflag < 0) { break; }
```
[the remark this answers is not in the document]
gardner48: Above, positive `kflag` values are handled by `arkCheckConvergece` so we only need to log negative return values right after the `step` call before the `break`. After that, both positive and negative return values need logged, so we use `SUNLogInfoIf` statements before the `if (kflag < 0)` to handle both cases. The `if (kflag < 0)` checks could be changed to an `if-else` but then we would have the same logging statement twice. This does add in extra `if` but only when info (or higher) logging is enabled but this should be off (in which case the `if` checks are removed) when performance is a concern.

src/arkode/arkode.c
```diff
@@ -912,13 +918,21 @@ int ARKodeEvolve(void* arkode_mem, sunrealtype tout, N_Vector yout,
...
 
       /* perform constraint-handling (if selected, and if solver check passed) */
       if (ark_mem->constraintsSet && (kflag == ARK_SUCCESS))
       {
         kflag = arkCheckConstraints(ark_mem, &constrfails, &nflag);
+
+        SUNLogInfoIf(kflag != ARK_SUCCESS, ARK_LOGGER, "end-step-attempt",
+                     "status = failed constraints, kflag = %i", kflag);
+
         if (kflag < 0) { break; }
```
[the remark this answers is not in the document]
gardner48: See response [here](https://github.com/LLNL/sundials/pull/501/files#r1876410765)

src/arkode/arkode.c
```diff
@@ -927,13 +941,19 @@ int ARKodeEvolve(void* arkode_mem, sunrealtype tout, N_Vector yout,
...
       /* check temporal error (if checks above passed) */
       if (kflag == ARK_SUCCESS)
       {
         kflag = arkCheckTemporalError(ark_mem, &nflag, &nef, dsm);
+
+        SUNLogInfoIf(kflag != ARK_SUCCESS, ARK_LOGGER, "end-step-attempt",
+                     "status = failed error test, dsm = %" RSYM ", kflag = %i",
+                     dsm, kflag);
+
         if (kflag < 0) { break; }
```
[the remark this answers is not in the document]
gardner48: See response [here](https://github.com/LLNL/sundials/pull/501/files#r1876410765)

src/arkode/arkode_lsrkstep.c
```diff
@@ -597,14 +605,22 @@ int lsrkStep_TakeStepRKC(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
...
   {
     retval = ark_mem->ProcessStage(ark_mem->tn + ark_mem->h * mus,
                                    ark_mem->tempv2, ark_mem->user_data);
-    if (retval != 0) { return ARK_POSTPROCESS_STAGE_FAIL; }
+    if (retval != 0)
+    {
+      SUNLogInfo(ARK_LOGGER, "end-stage",
+                 "status = failed postprocess stage, retval = %i", retval);
+      return ARK_POSTPROCESS_STAGE_FAIL;
+    }
```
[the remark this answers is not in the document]
gardner48: Because of the number of inputs needed I'm not sure this would end up being shorter and might be less clear to read.

test/unit_tests/logging/test_logging_arkode_sprkstep.cpp
gardner48: I've not added the `.out` file for this yet. I'll ping for another review pass when I've finished updating everything.

balos1 (dismissed):

gardner48:
The logging in all the ARKODE steppers have been updated now and CI tests added for each

src/arkode/arkode_forcingstep.c
```diff
@@ -234,37 +234,81 @@ static int forcingStep_TakeStep(ARKodeMem ark_mem, sunrealtype* dsmPtr,
...
   err = SUNStepper_SetStopTime(s0, tout);
-  if (err != SUN_SUCCESS) { return ARK_SUNSTEPPER_ERR; }
+  if (err != SUN_SUCCESS)
+  {
+    SUNLogInfo(ARK_LOGGER, "end-partition",
+               "status = failed set stop time, err = %i", err);
+    return ARK_SUNSTEPPER_ERR;
+  }
+
   err = SUNStepper_Evolve(s0, tout, ark_mem->ycur, &tret);
```
drreynolds, line 258: Perhaps we can save this for a future PR, but I'd think that it would be good to print `ark_mem->ycur` here when running with "extra debugging" enabled.

src/arkode/arkode_forcingstep.c
```diff
@@ -234,37 +234,81 @@ static int forcingStep_TakeStep(ARKodeMem ark_mem, sunrealtype* dsmPtr,
...
+  {
+    SUNLogInfo(ARK_LOGGER, "end-partition",
+               "status = failed set stop time, err = %i", err);
+    return ARK_SUNSTEPPER_ERR;
+  }
 
   /* Write tendency (ycur - yn)/h into stepper 1 forcing */
   sunrealtype hinv = SUN_RCONST(1.0) / ark_mem->h;
   N_VLinearSum(hinv, ark_mem->ycur, -hinv, ark_mem->yn, ark_mem->tempv1);
   err = SUNStepper_SetForcing(s1, ZERO, ZERO, &ark_mem->tempv1, 1);
```
drreynolds, line 284: Perhaps we can save this for a future PR, but I'd think that it would be good to print `ark_mem->tempv1` here when running with "extra debugging" enabled.

src/arkode/arkode_forcingstep.c
```diff
@@ -234,37 +234,81 @@ static int forcingStep_TakeStep(ARKodeMem ark_mem, sunrealtype* dsmPtr,
...
-  if (err != SUN_SUCCESS) { return ARK_SUNSTEPPER_ERR; }
+  if (err != SUN_SUCCESS)
+  {
+    SUNLogInfo(ARK_LOGGER, "end-partition",
+               "status = failed set forcing, err = %i", err);
+    return ARK_SUNSTEPPER_ERR;
+  }
 
   /* Evolve stepper 1 with the forcing */
   err = SUNStepper_Evolve(s1, tout, ark_mem->ycur, &tret);
```
drreynolds, line 293: Perhaps we can save this for a future PR, but I'd think that it would be good to print `ark_mem->ycur` here when running with "extra debugging" enabled.

src/arkode/arkode_splittingstep.c
```diff
@@ -253,20 +237,50 @@ static int splittingStep_SequentialMethod(ARKodeMem ark_mem,
...
+      {
+        SUNLogInfo(ARK_LOGGER, "end-partition",
+                   "status = failed set stop time, err = %i", err);
+        SUNLogInfo(ARK_LOGGER, "end-stage",
+                   "status = failed partition, err = %i", err);
+        return ARK_SUNSTEPPER_ERR;
+      }
 
       sunrealtype tret = ZERO;
       err              = SUNStepper_Evolve(stepper, t_end, y, &tret);
```
drreynolds, line 270: Perhaps we can save this for a future PR, but I'd think that it would be good to print `y` here when running with "extra debugging" enabled.

src/arkode/arkode_splittingstep.c
```diff
@@ -298,12 +321,21 @@ static int splittingStep_TakeStep(ARKodeMem ark_mem, sunrealtype* dsmPtr,
 
   for (int i = 1; i < coefficients->sequential_methods; i++)
   {
+    SUNLogInfo(ARK_LOGGER, "begin-sequential-method", "sequential method = %i",
+               i);
+
     N_VScale(ONE, ark_mem->yn, ark_mem->tempv1);
     retval = splittingStep_SequentialMethod(ark_mem, step_mem, i,
```
drreynolds, line 328: Perhaps we can save this for a future PR, but I'd think that it would be good to print `ark_mem->tempv1` here when running with "extra debugging" enabled.

src/arkode/arkode_splittingstep.c
```diff
@@ -298,12 +321,21 @@ static int splittingStep_TakeStep(ARKodeMem ark_mem, sunrealtype* dsmPtr,
...
                                             ark_mem->tempv1);
-    if (retval != ARK_SUCCESS) { return retval; }
+    if (retval != ARK_SUCCESS)
+    {
+      SUNLogInfo(ARK_LOGGER, "end-sequential-method",
+                 "status = failed sequential method, retval = %i", retval);
+      return retval;
+    }
     N_VLinearSum(ONE, ark_mem->ycur, coefficients->alpha[i], ark_mem->tempv1,
                  ark_mem->ycur);
```
drreynolds, line 337: Perhaps we can save this for a future PR, but I'd think that it would be good to print `ark_mem->ycur` here when running with "extra debugging" enabled.

drreynolds (approved):

src/arkode/arkode_forcingstep.c
gardner48: This is quick and easy to add, I'll put in the logging statements in a sec

drreynolds (dismissed):

drreynolds (approved):

balos1 (approved):

PR #502, opened 2024-06-10
Add SUNDIALS::core to CMake Docs

Steven-Roberts:


doc/shared/sundials/Install.rst
```diff
@@ -1780,7 +1780,7 @@ configuration file to build against SUNDIALS in their own CMake project.
   # Link to SUNDIALS libraries through the exported targets.
   # This is just an example, users should link to the targets appropriate
   # for their use case.
-  target_link_libraries(myexec PUBLIC SUNDIALS::cvode SUNDIALS::nvecpetsc)
+  target_link_libraries(myexec PUBLIC SUNDIALS::core SUNDIALS::cvode SUNDIALS::nvecpetsc)
```
gardner48, line 1783: The `core` library should get added automatically to the link line by the `SUNDIALS::cvode` target so it shouldn't be necessary to list it explicitly. Did you run into issues when not including it?

Steven-Roberts: I ran into issues with linking to the core, but it turned out to be unrelated to SUNDIALS. Looks like it isn't explicitly needed.

Issue #67, opened 2021-08-03
Trouble Compiling with Visual Studio and Intel Fortran


Bill-gitt:
Hello dear developers,

Summary: 
I am working for a University project where I have access to Windows 10, Visual studio 2013 and Intel Parallel studio 2017 or Visual Studio 2019 Community along Intel OneApi (includes Fortran Classic 2021).  

For about 1 month now I am trying to compile the CVode with Fortran Interface,  from Sundials 5.7 package. I hope it has thread-safely  even when the Fortran interface is used. Anyways after lots of effort I managed to compile and build CVode. Though when I run the Fortran examples in debug mode the program crashes. 

The steps I followed for compilation:

1.  I am using Cmake-gui version 3.21.1 (have tried this with version 3.20 too). I properly start Cmake and  point it to the source, build and install directories. Everything seems normal. Cmake locates my compiler C  properly. Here is some output from Cmake: 

```
Selecting Windows SDK version 10.0.19041.0 to target Windows 10.0.19042.
The C compiler identification is MSVC 19.29.30040.0
Detecting C compiler ABI info
Detecting C compiler ABI info - done
Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.29.30037/bin/Hostx64/x64/cl.exe - skipped
Detecting C compile features
Detecting C compile features - done
Looking for sys/types.h
Looking for sys/types.h - found
Looking for stdint.h
Looking for stdint.h - found
Looking for stddef.h
Looking for stddef.h - found
Check size of int64_t
Check size of int64_t - done
Using int64_t for indices
Looking for _POSIX_TIMERS
Looking for _POSIX_TIMERS - not found
Performing Test COMPILER_HAS_DEPRECATED_ATTR
Performing Test COMPILER_HAS_DEPRECATED_ATTR - Failed
Performing Test COMPILER_HAS_DEPRECATED
Performing Test COMPILER_HAS_DEPRECATED - Success
Added NVECTOR_SERIAL module
Added NVECTOR_MANYVECTOR module
Added SUNMATRIX_BAND module
.
.
.

```
2. The Cmake finds all the Sundial 5.7 source (Arcode, CVode....). I select only the CVode and Build_Fortran_Module_Interface and I deselect everything that has to do with testing and examples.  Cmake now locates the Intel compiler properly:

```Selecting Windows SDK version 10.0.19041.0 to target Windows 10.0.19042.
[...]

PR #506, opened 2024-06-14
Bugfix/windows fortran builds

balos1:


.github/workflows/windows-latest-mingw.yml
```diff
@@ -60,7 +62,7 @@ jobs:
 
     - name: Build
       # Build program
-      run: cmake --build ${GITHUB_WORKSPACE}/build
+      run: cmake --build ${GITHUB_WORKSPACE}/build --config ${{env.BUILD_TYPE}}
```
mmuetzel, line 65: MinGW doesn't use multi-config generators. Setting `--config` here isn't necessary. Instead the value of `CMAKE_BUILD_TYPE` in the configure step is decisive (like for most other platforms). 
(The `--config` flag in the build step might be a co-op.)

.github/workflows/windows-latest-mingw.yml
```diff
@@ -50,8 +50,10 @@ jobs:
       # Configure CMake in a 'build' subdirectory
       run: |
         cmake \
+          -G "MinGW Makefiles"
```
mmuetzel, line 53: MinGW Makefiles are sometimes a bit strange. "MSYS Makefiles" might be better.
But preferably you'd use the default generator for reset build environment, i.e., `-GNinja`.

mmuetzel:
To automatically get exports for the symbols in Fortran libraries on Windows, the following change might work:
``` diff
diff --git a/cmake/macros/SundialsAddLibrary.cmake b/cmake/macros/SundialsAddLibrary.cmake
index 84505e4f1..e80110ea0 100644
--- a/cmake/macros/SundialsAddLibrary.cmake
+++ b/cmake/macros/SundialsAddLibrary.cmake
@@ -464,6 +464,7 @@ macro(sundials_add_f2003_library target)
     VERSION ${sundials_add_f2003_library_VERSION}
     SOVERSION ${sundials_add_f2003_library_SOVERSION}
     ${sundials_add_f2003_library_UNPARSED_ARGUMENTS}
+    WINDOWS_EXPORT_ALL_SYMBOLS ON
   )
 
 endmacro()
```

balos1:
@mmuetzel Any thoughts on why the msys toolchain is [failing to link](https://github.com/LLNL/sundials/actions/runs/9569441560/job/26381984013?pr=506) with shared libraries still?  I do not have a system with msys setup to be able to inspect it easily and it seems like you are more familiar with the MSYS/Windows toolchain than myself.

balos1:
@gardner48 this is ready for review now

doc/shared/RecentChanges.rst
```diff
@@ -129,3 +129,5 @@ The unsupported implementations of ``N_VGetArrayPointer`` and
 ``N_VSetArrayPointer`` for the *hypre* and PETSc vectors are now deprecated.
 Users should access the underlying wrapped external library vector objects
 instead with ``N_VGetVector_ParHyp`` and ``N_VGetVector_Petsc``, respectively.
+
+Fixed several build errors with the Fortran interfaces on Windows systems.
```
gardner48, line 133: Move this to the "Bug Fixes" section above

cmake/SundialsSetupFortran.cmake
```diff
@@ -19,25 +19,6 @@
...
-
-# check if environment variables are used and CMAKE_Fortran_FLAGS is not
-if ((NOT "${ENV_FFLAGS}" STREQUAL "") AND (NOT "${ENV_FCFLAGS}" STREQUAL "")
-    AND ("${CMAKE_Fortran_FLAGS}" STREQUAL ""))
-
-  # check if environment variables are equal
-  if (NOT "${ENV_FFLAGS}" STREQUAL "${ENV_FCFLAGS}")
-    print_error("FFLAGS='${ENV_FFLAGS}' and FCFLAGS='${ENV_FCFLAGS}' are both set but are not equal.")
-  endif()
-endif()
```
gardner48, line 39: Did this cause problems with the Fortran interfaces? Checking for conflicting `FFLAGS` vs `FCFLAGS` used to be an xSDK requirement, is that still the case?

cmake/macros/SundialsAddLibrary.cmake
```diff
@@ -150,7 +150,7 @@ macro(sundials_add_library target)
     # --------------------------------------------------------------------------
 
     # create the target for the object library
-    add_library(${obj_target} OBJECT ${sources})
+    add_library(${obj_target} OBJECT ${sources} ${sundials_add_library_UNPARSED_ARGUMENTS})
```
gardner48, line 153: Commenting here in case it's not clear below, what are the extra inputs that are needed?

examples/kinsol/F2003_serial/kinLaplace_bnd_f2003.f90
```diff
@@ -131,7 +131,7 @@ program main
   ! local variables
   real(c_double)  :: fnormtol, fnorm(1)
   integer(c_int)  :: ierr
-  integer(c_long) :: mset, msubset
+  integer(c_int) :: mset, msubset
```
gardner48, line 134: I think these are inputs to functions expecting `long int`
```suggestion
  integer(c_long) :: mset, msubset
```

mmuetzel:
> @mmuetzel Any thoughts on why the msys toolchain is [failing to link](https://github.com/LLNL/sundials/actions/runs/9569441560/job/26381984013?pr=506) with shared libraries still? I do not have a system with msys setup to be able to inspect it easily and it seems like you are more familiar with the MSYS/Windows toolchain than myself.

It looks like `WINDOWS_EXPORT_ALL_SYMBOLS` doesn't work with the two-step creation of shared libraries in SUNDIALS. (I wasn't aware of that.) 
IIUC, an object library is created from the source files. And in a second step a shared library is created from that object library.
It looks like `WINDOWS_EXPORT_ALL_SYMBOLS` only exports symbols from the object files corresponding to a *shared* library (not from an object library).

Alternatively, symbols could be exported from a Fortran library using ATTRIBUTES.
I tried to add `dllexport` attributes manually to some of the functions and subroutines in `libsundials_fcore_mod` like this:
```
diff --git "a/src/sundials/fmod_int64/fsundials_core_mod.f90" "b/src/sundials/fmod_int64/fsundials_core_mod.f90"
index 7096d0c6c..9361fa509 100644
--- "a/src/sundials/fmod_int64/fsundials_core_mod.f90"
+++ "b/src/sundials/fmod_int64/fsundials_core_mod.f90"
@@ -2311,6 +2311,8 @@ type(C_PTR), target, intent(inout) :: ctx
 integer(C_INT) :: fresult 
 type(C_PTR) :: farg1 
 
+!GCC$ ATTRIBUTES DLLEXPORT :: FSUNContext_Free
+
 farg1 = c_loc(ctx)
 fresult = swigc_FSUNContext_Free(farg1)
 swig_result = fresult
@@ -2838,6 +2840,8 @@ type(N_Vector), target, intent(inout) :: z
 real(C_DOUBLE) :: farg1 
 type(C_PTR) :: farg2 
 
+!GCC$ ATTRIBUTES DLLEXPORT :: FN_VConst
+
 farg1 = c
 farg2 = c_loc(z)
 call swigc_FN_VConst(farg1, farg2)
@@ -4656,6 +4660,8 @@ type(SUNAdaptController), target, intent(inout) :: c
 integer(C_INT) :: fresult 
 type(C_PTR) :: farg1 
 
+!GCC$ ATTRIBUTES DLLEXPORT :: FSUNAdaptController_Destroy
+
 farg1 = c_loc(c)
 fresult = swigc_FSUNAdaptController_Destroy(farg1)
 swig_result = fresult
```

With these added, I can see that they are exported from the .dll with that change.

But iiuc, these attributes are compiler-specific. (I was using `gfortran` for the test.)
Intel compilers would require, e.g., `!DEC$ ATTRIBUTES DLLEXPORT :: FSUNAdaptController_Destroy` (`DEC` instead of `GCC`).

What is the motivation for the two step process to create shared libraries?

gardner48 (changes requested):

cmake/macros/SundialsAddLibrary.cmake
gardner48: I might have missed it, but I didn't see where this was needed.

balos1: I ended up not using it, but if you wanted to pass extra inputs to go to `add_library` from `sundials_add_library` this would be needed.

cmake/SundialsSetupFortran.cmake
balos1: It has not been an xSDK requirement for awhile (since xSDK dropped the CMake requirements for Spack).

src/sunadaptcontroller/imexgus/CMakeLists.txt
```diff
@@ -22,7 +22,6 @@ sundials_add_library(sundials_sunadaptcontrollerimexgus
     PUBLIC sundials_core
   INCLUDE_SUBDIR
     sunadaptcontroller
-  OBJECT_LIB_ONLY
```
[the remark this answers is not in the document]
balos1: They are only used by ARKODE right now, but is there any reason to not install them like other modules (since in theory they could be used without ARKODE)?

balos1:
@mmuetzel 
> It looks like WINDOWS_EXPORT_ALL_SYMBOLS doesn't work with the two-step creation of shared libraries in SUNDIALS. (I wasn't aware of that.)
It seemed to work with the Intel oneAPI compilers, but not gfortran under MSYS.

@gardner48 I have opened #507 to document this problem, and we can address it later on. For now I have disabled to Fortran interfaces in the CI for MSYS+gfortran+shared. I think we should go ahead and merge this PR since it resolves several other problems including #67  and the Intel compiler toolchain on Windows.

src/sunadaptcontroller/imexgus/fmod_int32/CMakeLists.txt
```diff
@@ -17,10 +17,8 @@ sundials_add_f2003_library(sundials_fsunadaptcontrollerimexgus_mod
     fsunadaptcontroller_imexgus_mod.f90 fsunadaptcontroller_imexgus_mod.c
   LINK_LIBRARIES
     PUBLIC sundials_fcore_mod
-  OBJECT_LIBRARIES
```
gardner48, line 20: I think we should keep the controllers as object libraries only since they're already part of ARKODE and not very useful as a standalone library.

gardner48 (changes requested):
Other than keeping the controllers as only object libraries the changes look good.

src/sunadaptcontroller/imexgus/CMakeLists.txt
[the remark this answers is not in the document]
balos1: This would not make the user have to link them into ARKODE.  It just makes the libraries available to link to. It would be just like how we build in the iterative linear solvers, some matrices, serial vector, and the newton and fixed point, but also provide the standalone libraries.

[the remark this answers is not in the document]
gardner48: Unlike the other objects I don't think the controllers are as useful to have installed for use as a separate library.

cmake/macros/SundialsAddLibrary.cmake
```diff
@@ -239,25 +247,22 @@ macro(sundials_add_library target)
...
-      # add any object library dependencies
+      set(_object_sources $<TARGET_OBJECTS:${obj_target}>)
       if(sundials_add_library_OBJECT_LIBRARIES)
         if(${_libtype} MATCHES "STATIC")
           append_static_suffix(sundials_add_library_OBJECT_LIBRARIES _all_objs)
         else()
           set(_all_objs ${sundials_add_library_OBJECT_LIBRARIES})
         endif()
         foreach(_tmp ${_all_objs})
-          # We use target_sources since target_link_libraries does not work
```
balos1, line 254: Ill note that we cannot actually use `target_link_libraries` even though we have newer CMake now because we dont actually export these object targets.

gardner48 (approved):

PR #508, opened 2024-06-19
Update CMake version in docs

Steven-Roberts:
These also might need updating:

https://github.com/LLNL/sundials/blob/b685654c6958699821fc68a822b66076fec8cae7/cmake/macros/SundialsAddLibrary.cmake#L254-L257

https://github.com/LLNL/sundials/blob/f64d994640792801b089962beb526924cc57420d/.gitlab/spack_packages/sundials/package.py#L210-L211

### DIFF
--- a/doc/shared/sundials/Install.rst
+++ b/doc/shared/sundials/Install.rst
@@ -101,7 +101,7 @@ generate Unix and Linux Makefiles, as well as KDevelop, Visual Studio, and
 CMake also provides a GUI front end and which allows an interactive build and
 installation process.
 
-The SUNDIALS build process requires CMake version 3.12.0 or higher and a working
+The SUNDIALS build process requires CMake version 3.18.0 or higher and a working
 C compiler.  On Unix-like operating systems, it also requires Make (and
 ``curses``, including its development libraries, for the GUI front end to CMake,
 ``ccmake`` or ``cmake-gui``), while on Windows it requires Visual Studio.  While


gardner48 (approved):

gardner48:
> These also might need updating:
> 
> https://github.com/LLNL/sundials/blob/b685654c6958699821fc68a822b66076fec8cae7/cmake/macros/SundialsAddLibrary.cmake#L254-L257

This has been addressed in #506

> https://github.com/LLNL/sundials/blob/f64d994640792801b089962beb526924cc57420d/.gitlab/spack_packages/sundials/package.py#L210-L211

The spack package does need to be updated. Spack will prefer new CMake versions which is why this hasn't been an issue.

Steven-Roberts:
| The spack package does need to be updated. Spack will prefer new CMake versions which is why this hasn't been an issue.

Can we fix that in this PR? I assume we want
```
depends_on("cmake@3.18:", type="build") 
```

balos1:
It will need to be something like
```
depends_on("cmake@3.12:", type="build", when=":7.0.0") 
depends_on("cmake@3.18:", type="build", when="7.0.0:") 
```
FWIW, the Spack package in this repo is only used internally for our CI so it doesnt really matter. Additionally, the package needs to actually be synced with the [one](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/sundials/package.py) in Spack proper (and this change needs to be made there too).

balos1 (approved):

PR #509, opened 2024-06-20
Feature/format fortran

balos1:
Add fortran formatting and apply it. In the release pR, we will want to add the merge commit to [`.git-blame-ignore-revs`](https://github.com/LLNL/sundials/blob/main/.git-blame-ignore-revs).


balos1:
> Changes look good but the formatting CI job failed

Looks like I had missed a file.

gardner48 (approved):

PR #520, opened 2024-06-21
CMake: Formatting with cmake-format

gardner48:
Add CMake formatter


gardner48:
Also see alternative formatter in #520

balos1:
I have a slight preference to the formatting in this PR over #521.  Also, cmake-format seems to be much more popular so I think its probably the better choice for that reason too.

gardner48:
> I have a slight preference to the formatting in this PR over #521. Also, cmake-format seems to be much more popular so I think its probably the better choice for that reason too.

My main concern with `cmake-format` was that it doesn't seem to have any activity in the past three years. Which is why I looked at `gersemi` as well. One thing I like about `gersemi` over `cmake-format` is it can format custom functions/macros without having to supply additional information so it does a better join with `sundials_add_library`.

gardner48:
I updated the `cmake-format.py` file so it knows about our custom macros so those are formatted better now

balos1:
> > I have a slight preference to the formatting in this PR over #521. Also, cmake-format seems to be much more popular so I think its probably the better choice for that reason too.
> 
> My main concern with `cmake-format` was that it doesn't seem to have any activity in the past three years. Which is why I looked at `gersemi` as well. One thing I like about `gersemi` over `cmake-format` is it can format custom functions/macros without having to supply additional information so it does a better job with things like with `sundials_add_library`.

Hmm, I did not realize `cmake-format` was not maintained anymore. If it starts causing problems for us later on, I suppose worst case scenario we disable it in the formatting check.

balos1 (approved):

PR #532, opened 2024-06-26
Maintenance: Start recent changes for next release

gardner48:


drreynolds (approved):
It would be great if these tasks could be automated.

gardner48:
> It would be great if these tasks could be automated.

That would be nice. There's already a script to make the changes so we'd just need a bot to run it and post the PR

Issue #350, opened 2023-10-06
use oneMKL to provide CPU LAPACK functions


stuart-nolan:
_Originally posted by @gardner48 in https://github.com/LLNL/sundials/issues/173#issuecomment-1176270242_

Naively following the first suggestion in #173 by doing something like `cmake ... -DBLA_VENDOR=Intel10_64ilp ...` results in
```
CMake Warning:
  Manually-specified variables were not used by the project:

    BLA_VENDOR
```
FindBLAS exists on my system (ubuntu 22.04) but apparently is not implemented/used by sundials.  FWIW FindBLAS does work with MKL for at least one other lib (suitesparse) on my system.

The second suggestion, specify the lapack libraries along with appropriate (global) cflags, works; however results in all built dynamic libraries having links to MKL.  There does not seem to be cmake options for linking and c flags specific to lapack.  One can do two builds  e.g. one for all non lapack libraries, a second for only the two sundials lapack libraries, and then use the two lib sets together (this is what I've done and so far works) but seems like a kludge.

Can the sundials cmake build system better handle linking to MKL?  Have I missed some cmake option or should I label this issuse as a feature request?

balos1:
Did you set `ENABLE_LAPACK` to `ON`? `BLA_VENDOR` should be used. Since SUNDIALS doesn't ever use BLAS without also needing LAPACK, we only call FindLAPACK.

stuart-nolan:
Thank you for responding.  I tried multiple ways, most using the cmake command line and nearly all with -DENABLE_LAPACK=ON.  That said, I also tried the ccmake curses gui (with a simpler config than I give below) which gave the output from the first post in this issue.  

For the cmake cli, I use a script to generate the cmake command, they are long but here are examples:  

<details>

<summary>cmake cli examples</summary>
With -DBLA_VENDOR=Intel10_64ilp
```
/usr/bin/cmake -DBUILD_STATIC_LIBS=OFF -DENABLE_OPENMP=ON -DCMAKE_INSTALL_PREFIX:PATH=/home/ul/.local \
[...]

PR #542, opened 2024-07-15
CMake: Use LAPACK imported target

gardner48:
1. Use `LAPACK::LAPACK` rather than `LAPACK_LIBRARIES`
2. Simplify `try_compile` test
3. Don't overwrite `LAPACK_WORKS` (test every configuration pass)
4. Use `CHECK_START`, `CHECK_PASSED`, and `CHECK_FAILED` messages in compile test
5. Replace manual include guard with `include_guard(GLOBAL)` 
6. Update `SundialsTPL.cmake.template` with above changes


gardner48:
This may help address #350.

gardner48:
@balos1 If you're good with the updates to `SundialsTPL.cmake.template` and `LAPACK_WORKS`, I'll see about propagating these changes to the other TPL files. 

With `LAPACK_WORKS` this is now only used to bypass not the compile test and not to indicate if the compile test passed. So `_WORKS` might not be the best name, maybe `LAPACK_TEST` or `SUNDIALS_LAPACK_TEST` or `SUN_LAPACK_TEST` (for better namespacing).

balos1:
This looks great!

balos1 (dismissed):

gardner48:
Looks like there is a problem with the compile test that might be specific to CMake 3.18, I'm looking into it.

cmake/tpl/SundialsLapack.cmake
```diff
@@ -26,41 +26,34 @@
...
+# LAPACK_LIBRARIES contains multiple items only use the path of the first entry
+list(LENGTH LAPACK_LIBRARIES len)
+if(len EQUAL 1)
+  get_filename_component(LAPACK_LIBRARY_DIR ${LAPACK_LIBRARIES} PATH)
 else()
-  message(STATUS "Looking for LAPACK libraries... FAILED")
-  set(LAPACK_FOUND FALSE)
+  list(GET LAPACK_LIBRARIES 0 TMP_LAPACK_LIBRARIES)
+  get_filename_component(LAPACK_LIBRARY_DIR ${TMP_LAPACK_LIBRARIES} PATH)
 endif()
```
Steven-Roberts, line 55: Untested suggestion:
```suggestion
list(GET LAPACK_LIBRARIES 0 TMP_LAPACK_LIBRARIES)
get_filename_component(LAPACK_LIBRARY_DIR ${TMP_LAPACK_LIBRARIES} PATH)
```

cmake/tpl/SundialsLapack.cmake
```diff
@@ -340,49 +317,39 @@ if(LAPACK_LIBRARIES AND (NOT LAPACK_WORKS))
...
-  # Attempt to build and link the "ltest" executable
+  # Workaround bug in older versions of CMake where the BLAS::BLAS target, which
+  # LAPACK::LAPACK depends on, is not defined in the file
+  # ${LAPACK_TEST_DIR}/CMakeFiles/CMakeTmp/<random_name>Targets.cmake created by
+  # try_compile
+  if(CMAKE_VERSION VERSION_LESS 3.20)
+    set(_lapack_targets LAPACK::LAPACK BLAS::BLAS)
+  else()
+    set(_lapack_targets LAPACK::LAPACK)
+  endif()
```
Steven-Roberts, line 331: Also untested

```suggestion
  set(_lapack_targets LAPACK::LAPACK)
  if(CMAKE_VERSION VERSION_LESS 3.20)
    list(APPEND _lapack_targets BLAS::BLAS)
  endif()
```

src/sunlinsol/lapackdense/CMakeLists.txt
```diff
@@ -16,6 +16,14 @@
 
 install(CODE "MESSAGE(\"\nInstall SUNLINSOL_LAPACKDENSE\n\")")
 
+# Workaround bug in older versions of CMake where the BLAS::BLAS target, which
+# LAPACK::LAPACK depends on, is not picked up by LINK_LIBRARIES
+if(CMAKE_VERSION VERSION_LESS 3.20)
+  set(_lapack_targets LAPACK::LAPACK BLAS::BLAS)
+else()
+  set(_lapack_targets LAPACK::LAPACK)
+endif()
```
Steven-Roberts, line 25: Same as in `SundialsLapack.cmake`

src/sunlinsol/lapackband/CMakeLists.txt
```diff
@@ -16,6 +16,14 @@
 
 install(CODE "MESSAGE(\"\nInstall SUNLINSOL_LAPACKBAND\n\")")
 
+# Workaround bug in older versions of CMake where the BLAS::BLAS target, which
+# LAPACK::LAPACK depends on, is not picked up by LINK_LIBRARIES
+if(CMAKE_VERSION VERSION_LESS 3.20)
+  set(_lapack_targets LAPACK::LAPACK BLAS::BLAS)
+else()
+  set(_lapack_targets LAPACK::LAPACK)
+endif()
```
Steven-Roberts, line 25: Same as in `SundialsLapack.cmake`

Steven-Roberts (commented):
This fixes the missing linking flags issue I had. When I have two lapack versions (system and spack) it finds both in cmake and only uses the system version for compilation. It's unclear how I would use the spack version for compilation.

gardner48:
> It's unclear how I would use the spack version for compilation.

Unfortunately `FindLAPACK` does not seem to support the usual `_ROOT` or `_DIR` options to help the set search path so you'll probably have to set `LAPACK_LIBRARIES` and `LAPACK_LINKER_FLAGS` manually. I've not tried it but setting `BLA_VENDOR` might also work e.g., `-DBLA_VENDOR="OpenBLAS"`

Steven-Roberts:
Got it working with David's help. For posterity,  `-DBLA_VENDOR="OpenBLAS"` did work as well as manually setting `LAPACK_LIBRARIES` and `LAPACK_LINKER_FLAGS`. Just had to make sure the compiler version was consistent with the OpenBLAS build from spack.

Steven-Roberts:
@gardner48 The latest version still works for me, both with a system LAPACK and spack one.

cmake/tpl/SundialsTPL.cmake.template
```diff
@@ -46,38 +42,36 @@ find_package(<TPL> REQUIRED)
...
 
-  # Create the <TPL>_TEST directory
-  set(<TPL>_TEST_DIR ${PROJECT_BINARY_DIR}/<TPL>_TEST)
-  file(MAKE_DIRECTORY ${<TPL>_TEST_DIR})
+# Try building a simple test
+if(NOT <TPL>_WORKS)
 
-  # Create a CMakeLists.txt file
-  file(WRITE ${<TPL>_TEST_DIR}/CMakeLists.txt "")
+  message(CHECK_START "Testing <TPL>")
```
balos1, line 50: I did not know about the `CHECK_START` message option... cool.

balos1 (approved):

cmake/tpl/SundialsTPL.cmake.template
gardner48: This got added in 3.17 so with the update to 3.18 as the minimum we can make some things a little nicer

PR #546, opened 2024-07-26
Fix ARKODE_VERNER_8_5_6 reference

Steven-Roberts:


drreynolds (approved):

doc/arkode/guide/source/Butcher.rst
drreynolds, line 1: Given the misattributed source, perhaps we should add a new constant for this (e.g., `ARKODE_HEJ_8_5_6`) and then deprecate the current name?

Steven-Roberts: It was Verner who derived the method. To my knowledge, the new citation is the first publication providing the method. Verner is not an author but the paper cites personal communication with Verner to get the coefficients. I think the current name is ok.

PR #547, opened 2024-07-26
Update default ARKODE methods

Steven-Roberts:
Replaces https://github.com/LLNL/sundials/pull/345

In draft status until we settle on the defaults. Then I'll update the docs and output files.

Justification:
- ARK order >2: Follows the recommendations in the conclusion of https://doi.org/10.1016/j.apnum.2018.10.007
- ERK order 4: Improved principal error and embedded quality metrics
- ERK order 5: Improved principal error and embedded quality metrics
- ERK order 8: Switch to a method designed for local extrapolation mode. Likely better principal error but need to confirm
- IRK order 2: Improved principal error, L-stability, and stage order 2
- IRK order >2: Follows the recommendations in the conclusion of https://doi.org/10.1016/j.apnum.2019.07.008


doc/arkode/guide/source/Butcher.rst
```diff
@@ -477,6 +513,7 @@ or :c:func:`ARKodeButcherTable_LoadERK`.
 Accessible via the string ``"ARKODE_SOFRONIOU_SPALETTA_5_3_4"`` to
 :c:func:`ARKStepSetTableName`, :c:func:`ERKStepSetTableName` or
 :c:func:`ARKodeButcherTable_LoadERKByName`.
+This is the default 4th order explicit method
```
drreynolds, line 516: I'm not sure why periods were added in-between the text and all previous references for where the methods came from, but to be consistent a period should either be added here (or removed above).

doc/arkode/guide/source/Constants.rst
```diff
@@ -78,6 +78,8 @@ contains the ARKODE output constants.
    +-----------------------------------------------+------------------------------------------------------------+
    | :index:`ARKODE_FORWARD_EULER_1_1`             | Use the Forward-Euler-1-1 ERK method.                      |
    +-----------------------------------------------+------------------------------------------------------------+
+   | :index:`ARKODE_RALSTON_3_1_2`                 | Use the Ralston-3-1-2 ERK method.                          |
+   +-----------------------------------------------+------------------------------------------------------------+
```
drreynolds, line 82: These methods are generally listed in increasing order of accuracy, so I'd move `ARKODE_RALSTON_3_1_2` further down the list.  We might also want to adjust the order of the ones at the end of the section, and add the method/embedding orders of accuracy to the table text.

Steven-Roberts: It's after the first order method and with the other second order methods. I put it first among the second order since it's the new default. We should settle on an unambiguous ordering so, then I can rearrange the table and the Butcher table docs to be consistent.

doc/arkode/guide/source/Butcher.rst
Steven-Roberts, line 516:
```suggestion
This is the default 4th order explicit method.
```

doc/arkode/guide/source/Butcher.rst
```diff
@@ -671,7 +707,50 @@ Accessible via the string ``"ARKODE_SAYFY_ABURUB_6_3_4"`` to
    region is outlined in blue; the embedding's region is in red.
 
 
+.. _Butcher.Tsitouras:
+
+Tsitouras-6-4-5
+^^^^^^^^^^^^^^^
+
+.. index:: Tsitouras-6-4-5 ERK method
```
gardner48, line 715:
```suggestion
Tsitouras-7-4-5
^^^^^^^^^^^^^^^

.. index:: Tsitouras-7-4-5 ERK method
```

doc/arkode/guide/source/Butcher.rst
```diff
@@ -671,7 +707,50 @@ Accessible via the string ``"ARKODE_SAYFY_ABURUB_6_3_4"`` to
...
+         0.1610000000000000 & 0.1610000000000000 & 0 & 0 & 0 & 0 & 0 & 0 \\
+         0.3270000000000000 & -0.008480655492356989 & 0.3354806554923570 & 0 & 0 & 0 & 0 & 0 \\
+         0.9000000000000000 & 2.897153057105493 & -6.359448489975075 & 4.362295432869581 & 0 & 0 & 0 & 0 \\
+         0.9800255409045097 & 5.325864828439257 & -11.74888356406283 & 7.495539342889836 & -0.09249506636175525 & 0 & 0 & 0 \\
+         1.000000000000000 & 5.861455442946420 & -12.92096931784711 & 8.159367898576159 & -0.07158497328140100 & -0.02826905039406838 & 0 & 0 \\
+         1.000000000000000 & 0.09646076681806523 & 0.01000000000000000 & 0.4798896504144996 & 1.379008574103742 & -3.290069515436081 & 2.324710524099774 & 0 \\
+         \hline
+         5 & 0.09646076681806523 & 0.01000000000000000 & 0.4798896504144996 & 1.379008574103742 & -3.290069515436081 & 2.324710524099774 & 0 \\
+         4 & 0.09352374858189271 & 0.008652883141566368 & 0.4928930991314319 & 1.140235412267858 & -2.329180192439365 & 1.568875049316616 & 0.02500000000000000
+      \end{array}
```
gardner48, line 741: Could the trailing zeros be removed i.e., `0.01000000000000000` to `0.01`?

gardner48 (changes requested):

doc/arkode/guide/source/Constants.rst
Steven-Roberts: I like that idea, but unless we want anything more detailed that name and order, I think we should save that for later.

gardner48: Agreed, I was thinking this could be the start of such a table

doc/arkode/guide/source/Butcher.rst
Steven-Roberts: Done

Steven-Roberts:
@gardner48 @drreynolds can you confirm this is the table format you're looking for:
https://sundials--547.org.readthedocs.build/en/547/arkode/Butcher_link.html#id61

If that looks good, I'll do the same for DIRK and SPRK (MRI already exists). I'll also remove all the corresponding constants from `Constants.rst`.

gardner48:
> @gardner48 @drreynolds can you confirm this is the table format you're looking for: https://sundials--547.org.readthedocs.build/en/547/arkode/Butcher_link.html#id61
> 
> If that looks good, I'll do the same for DIRK and SPRK (MRI already exists). I'll also remove all the corresponding constants from `Constants.rst`.

Yes, that's what I was thinking

doc/arkode/guide/source/Constants.rst
```diff
@@ -74,249 +74,6 @@ contains the ARKODE output constants.
...
-   |                                               | (ARKODE_FORWARD_EULER_1_1).                                |
-   +-----------------------------------------------+------------------------------------------------------------+
-   | :index:`ARKSTEP_DEFAULT_ERK_2`                | Use ARKStep's default second-order ERK method              |
-   |                                               | (ARKODE_HEUN_EULER_2_1_2).                                 |
-   +-----------------------------------------------+------------------------------------------------------------+
-   | :index:`ARKSTEP_DEFAULT_ERK_3`                | Use ARKStep's default third-order ERK method               |
-   |                                               | (ARKODE_BOGACKI_SHAMPINE_4_2_3).                           |
-   +-----------------------------------------------+------------------------------------------------------------+
-   | :index:`ARKSTEP_DEFAULT_ERK_4`                | Use ARKStep's default fourth-order ERK method              |
-   |                                               | (ARKODE_ZONNEVELD_5_3_4).                                  |
```
drreynolds, line 139: While I really like having a table with method properties in `Butcher.rst`, the documentation now omits any reference to the constants for "default" methods, e.g., `ARKSTEP_DEFAULT_ERK_2` and `ARKSTEP_DEFAULT_ARK_ETABLE_5`.  I wonder if those constants could be retained in the tables on this page?

gardner48: The new table denotes the default method with an asterisk in the order column. Maybe this could be made clearer by putting "(default)" after the name in the new table.

drreynolds: Sure, but in the current form of this PR, users are never informed of the "names" `ARKSTEP_DEFAULT_ERK_2`, etc.  Perhaps we don't care for users to know about those names since they're already the defaults?  But if we could anticipate a scenario where a user would want to specify those names/constants, then we should let them know what they are.

gardner48: Ah, I see what you mean. Yes, it would be good to keep the `ARKSTEP_DEFAULT_*` entries in this table or have a separate table with just the defaults.

Steven-Roberts: I added them back to the table in `Constants.rst`

Steven-Roberts:
I think this is ready to merge unless someone would like to re-review

gardner48:
> I think this is ready to merge unless someone would like to re-review

It's been a while since I looked over this so I'll take another pass

drreynolds (dismissed):

gardner48 (approved):

PR #548, opened 2024-07-30
Simplify/Fix Adaptivity for Small Errors or no CFL

Steven-Roberts:
The `TINY` parameter used in the controllers could cause an error overestimation, particularly for the first step.

### DIFF
--- a/src/sunadaptcontroller/imexgus/sunadaptcontroller_imexgus.c
+++ b/src/sunadaptcontroller/imexgus/sunadaptcontroller_imexgus.c
@@ -48,7 +48,6 @@
 #define DEFAULT_K1I  SUN_RCONST(0.95)
 #define DEFAULT_K2I  SUN_RCONST(0.95)
 #define DEFAULT_BIAS SUN_RCONST(1.5)
-#define TINY         SUN_RCONST(1.0e-10)
 
 /* -----------------------------------------------------------------
  * exported functions
@@ -135,15 +134,15 @@ SUNErrCode SUNAdaptController_EstimateStep_ImExGus(SUNAdaptController C,
 
   /* set usable time-step adaptivity parameters -- first step */
   const sunrealtype k = -SUN_RCONST(1.0) / ord;
-  const sunrealtype e = SUNMAX(SACIMEXGUS_BIAS(C) * dsm, TINY);
+  const sunrealtype e = SACIMEXGUS_BIAS(C) * dsm;
 
   /* set usable time-step adaptivity parameters -- subsequent steps */
   const sunrealtype k1e  = -SACIMEXGUS_K1E(C) / ord;
   const sunrealtype k2e  = -SACIMEXGUS_K2E(C) / ord;
   const sunrealtype k1i  = -SACIMEXGUS_K1I(C) / ord;
   const sunrealtype k2i  = -SACIMEXGUS_K2I(C) / ord;
-  const sunrealtype e1   = SUNMAX(SACIMEXGUS_BIAS(C) * dsm, TINY);
-  const sunrealtype e2   = e1 / SUNMAX(SACIMEXGUS_EP(C), TINY);
+  const sunrealtype e1   = SACIMEXGUS_BIAS(C) * dsm;
+  const sunrealtype e2   = e1 / SACIMEXGUS_EP(C);
   const sunrealtype hrat = h / SACIMEXGUS_HP(C);
 
   /* compute estimated time step size, modifying the first step formula */
@@ -152,6 +151,10 @@ SUNErrCode SUNAdaptController_EstimateStep_ImExGus(SUNAdaptController C,
   {
     *hnew = h * SUNMIN(hrat * SUNRpowerR(e1, k1i) * SUNRpowerR(e2, k2i),
                        SUNRpowerR(e1, k1e) * SUNRpowerR(e2, k2e));
+  }
+
+  if (isnan(*hnew)) {
+    *hnew = INFINITY;
   }
 
   /* return with success */
--- a/src/sunadaptcontroller/soderlind/sunadaptcontroller_soderlind.c
+++ b/src/sunadaptcontroller/soderlind/sunadaptcontroller_soderlind.c
@@ -62,7 +62,6 @@
 #define DEFAULT_IMPGUS_K1 SUN_RCONST(0.98) /* Implicit Gustafsson parameters */
 #define DEFAULT_IMPGUS_K2 SUN_RCONST(0.95)
 #define DEFAULT_BIAS      SUN_RCONST(1.5)
-#define TINY              SUN_RCONST(1.0e-10)
 
 /* -----------------------------------------------------------------
  * exported functions
@@ -322,9 +321,9 @@ SUNErrCode SUNAdaptController_EstimateStep_Soderlind(SUNAdaptController C,
   const sunrealtype k3    = -SODERLIND_K3(C) / ord;
   const sunrealtype k4    = SODERLIND_K4(C);
   const sunrealtype k5    = SODERLIND_K5(C);
-  const sunrealtype e1    = SUNMAX(SODERLIND_BIAS(C) * dsm, TINY);
-  const sunrealtype e2    = SUNMAX(SODERLIND_EP(C), TINY);
-  const sunrealtype e3    = SUNMAX(SODERLIND_EPP(C), TINY);
+  const sunrealtype e1    = SODERLIND_BIAS(C) * dsm;
+  const sunrealtype e2    = SODERLIND_EP(C);
+  const sunrealtype e3    = SODERLIND_EPP(C);
   const sunrealtype hrat  = h / SODERLIND_HP(C);
   const sunrealtype hrat2 = SODERLIND_HP(C) / SODERLIND_HPP(C);
 
@@ -338,6 +337,10 @@ SUNErrCode SUNAdaptController_EstimateStep_Soderlind(SUNAdaptController C,
   {
     *hnew = h * SUNRpowerR(e1, k1) * SUNRpowerR(e2, k2) * SUNRpowerR(e3, k3) *
             SUNRpowerR(hrat, k4) * SUNRpowerR(hrat2, k5);
+  }
+
+  if (isnan(*hnew)) {
+    *hnew = INFINITY;
   }
 
   /* return with success */
--- a/src/sundials/sundials_math.c
+++ b/src/sundials/sundials_math.c
@@ -22,6 +22,8 @@
 #include <sundials/sundials_math.h>
 #include <sundials/sundials_types.h>
 
+#include <sundials/priv/sundials_errors_impl.h>
+
 sunrealtype SUNRpowerI(sunrealtype base, int exponent)
 {
   int i, expt;
@@ -36,7 +38,9 @@ sunrealtype SUNRpowerI(sunrealtype base, int exponent)
 
 sunrealtype SUNRpowerR(sunrealtype base, sunrealtype exponent)
 {
-  if (base <= SUN_RCONST(0.0)) { return (SUN_RCONST(0.0)); }
+  // TODO(SBR): cleanup this and header
+  // if (base <= SUN_RCONST(0.0)) { return (SUN_RCONST(0.0)); }
+  SUNAssert(base >= 0.0, "Base should be positive");
 
 #if defined(SUNDIALS_DOUBLE_PRECISION)
   return (pow(base, exponent));
--- a/src/sundials/sundials_math.c
+++ b/src/sundials/sundials_math.c
@@ -22,7 +22,7 @@
 #include <sundials/sundials_math.h>
 #include <sundials/sundials_types.h>
 
-#include <sundials/priv/sundials_errors_impl.h>
+#include <assert.h>
 
 sunrealtype SUNRpowerI(sunrealtype base, int exponent)
 {
@@ -40,7 +40,7 @@ sunrealtype SUNRpowerR(sunrealtype base, sunrealtype exponent)
 {
   // TODO(SBR): cleanup this and header
   // if (base <= SUN_RCONST(0.0)) { return (SUN_RCONST(0.0)); }
-  SUNAssert(base > 0.0, "Base should be positive");
+  assert(base > 0.0);
 
 #if defined(SUNDIALS_DOUBLE_PRECISION)
   return (pow(base, exponent));
--- a/src/sundials/sundials_math.c
+++ b/src/sundials/sundials_math.c
@@ -40,7 +40,7 @@ sunrealtype SUNRpowerR(sunrealtype base, sunrealtype exponent)
 {
   // TODO(SBR): cleanup this and header
   // if (base <= SUN_RCONST(0.0)) { return (SUN_RCONST(0.0)); }
-  SUNAssert(base >= 0.0, "Base should be positive");
+  SUNAssert(base > 0.0, "Base should be positive");
 
 #if defined(SUNDIALS_DOUBLE_PRECISION)
   return (pow(base, exponent));
--- a/src/arkode/arkode_adapt.c
+++ b/src/arkode/arkode_adapt.c
@@ -131,7 +131,7 @@ int arkAdapt(ARKodeMem ark_mem, ARKodeHAdaptMem hadapt_mem, N_Vector ycur,
                     "Error in explicit stability function.");
     return (ARK_ILL_INPUT);
   }
-  if (h_cfl <= ZERO) { h_cfl = SUN_RCONST(1.0e30) * SUNRabs(hcur); }
+  if (h_cfl <= ZERO) { h_cfl = INFINITY; }
 
 #if SUNDIALS_LOGGING_LEVEL >= SUNDIALS_LOGGING_INFO
   SUNLogger_QueueMsg(ARK_LOGGER, SUN_LOGLEVEL_INFO, "ARKODE::arkAdapt",
--- a/src/arkode/arkode_adapt.c
+++ b/src/arkode/arkode_adapt.c
@@ -156,7 +156,7 @@ int arkAdapt(ARKodeMem ark_mem, ARKodeHAdaptMem hadapt_mem, N_Vector ycur,
 #endif
 
   /* increment the relevant step counter, set desired step */
-  if (SUNRabs(h_acc) < SUNRabs(h_cfl)) { hadapt_mem->nst_acc++; }
+  if (SUNRabs(h_acc) <= SUNRabs(h_cfl)) { hadapt_mem->nst_acc++; }
   else { hadapt_mem->nst_exp++; }
   h_acc = int_dir * SUNMIN(SUNRabs(h_acc), SUNRabs(h_cfl));
 


drreynolds (commented):
I am not expert on how robustly various architectures/compilers work with infinity.  I'd like to see this PR pass all possible regression tests on all architectures before approving, but if infinity is handled correctly everywhere, then I would approve of these changes.

As another option, we could use the largest/smallest representable floating point numbers in the given precision to avoid having to do arithmetic with infinity.

Steven-Roberts:
The behavior of pow, abs, min, max, and other functions for inf/nan is specified by the C standard, so I'm not too worried about that. I think a unit test where the error is 0 is warranted to double check. The standard only specifies that `pow(0, 0)` may cause a domain error (and doesn't even specify what the value could be), and `pow(0, y)` can cause floating point exceptions for certain `y`. I think I'll need to handle that more carefully in the controller.

Steven-Roberts:
After considering the edge cases around `pow(0,0)`, i.e., when there's no local error and a controller exponent is 0, it turns out to be benign. The C standard specifies `pow(x,0)=1` for all `x`, and there's no undefined behavior. The test I added seems to support this for some limited cases.

src/sundials/sundials_math.c
```diff
@@ -36,8 +36,6 @@ sunrealtype SUNRpowerI(sunrealtype base, int exponent)
 
 sunrealtype SUNRpowerR(sunrealtype base, sunrealtype exponent)
 {
-  if (base <= SUN_RCONST(0.0)) { return (SUN_RCONST(0.0)); }
```
Steven-Roberts, line 39: I went through every reference to this function across SUNDIALS and did not see a case where `base` could be negative. There are a few cases where `base` could be 0, but `exponent` is finite so the behavior is unchanged.

src/sunadaptcontroller/imexgus/sunadaptcontroller_imexgus.c
```diff
@@ -133,27 +132,35 @@ SUNErrCode SUNAdaptController_EstimateStep_ImExGus(SUNAdaptController C,
...
     *hnew = h * SUNMIN(hrat * SUNRpowerR(e1, k1i) * SUNRpowerR(e2, k2i),
                        SUNRpowerR(e1, k1e) * SUNRpowerR(e2, k2e));
   }
 
+  if (!isfinite(*hnew))
+  {
+    /* hnew can be INFINITY or NAN if multiple e's are 0 or there are overflows.
+     * In that case, make hnew INFINITY with the same sign as h */
+    *hnew = SUNRcopysign(INFINITY, h);
+  }
```
gardner48, line 162: If `hnew` is `NAN`, should that be an error?

test/unit_tests/arkode/C_serial/ark_test_adapt.c
```diff
@@ -0,0 +1,153 @@
...
+  /* The ERK method should be able to take the maximum possible timestep without
+     any rejected steps or local error */
+  ARKodeEvolve(arkode_mem, tout, y, &tret, ARK_ONE_STEP);
+
+  long int local_err_fails;
+  ARKodeGetNumErrTestFails(arkode_mem, &local_err_fails);
+  if (local_err_fails != 0)
+  {
+    fprintf(stderr, "Expected 0 local error failures at step %i but is %li\n",
+            step, local_err_fails);
```
gardner48, line 57:
```suggestion
    fprintf(stderr, "Expected 0 local error failures at step %i but is %li\n",
            step, local_err_fails);
    return 1;
```

test/unit_tests/arkode/C_serial/ark_test_adapt.c
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  const N_Vector err = N_VClone(y);
+  ARKodeGetEstLocalErrors(arkode_mem, err);
+  const sunrealtype err_norm = N_VMaxNorm(err);
+  N_VDestroy(err);
+
+  if (err_norm != 0)
```
gardner48, line 65: Should this be
```suggestion
  if (SUNRCompare(err_norm, SUN_RCONST(0.0))
```

test/unit_tests/arkode/C_serial/ark_test_adapt.c
```diff
@@ -0,0 +1,153 @@
...
+                       SUN_RCONST(2.3));
+  if (retval != 0) { return retval; }
+
+  /* Forward integration from a negative time */
+  ERKStepReInit(arkode_mem, f, SUN_RCONST(-999.0), y);
+  retval = check_steps(arkode_mem, y, SUN_RCONST(20.0), SUN_RCONST(1.0e5),
+                       SUN_RCONST(1.1e3));
+  if (retval != 0) { return retval; }
+
+  /* TODO(SBR): test non-default controllers */
```
gardner48, line 144: Does this `TODO` need to be addressed?

gardner48 (changes requested):

src/sunadaptcontroller/imexgus/sunadaptcontroller_imexgus.c
Steven-Roberts: I don't see why it should be an error. As I state in the comment, `hnew` can be `NAN` from normal situations such as when e's are 0. In that case, we set `hnew` to the maximum possible value.

test/unit_tests/arkode/C_serial/ark_test_adapt.c
Steven-Roberts: Yes, I'll try to add a couple more tests

test/unit_tests/arkode/C_serial/ark_test_adapt.c
Steven-Roberts: The error should be exactly 0 in floating point arithmetic, and I don't see any reason it should ever be some epsilon off. I'll add the `SUN_RCONST`, but I don't think we should allow any tolerance on this.

src/sunadaptcontroller/imexgus/sunadaptcontroller_imexgus.c
gardner48: How do we distinguish between norm situations and errors we should catch? This is probably implementation specific but `NAN` can occur with a negative base and finite exponent or `NAN` as the exponent.

test/unit_tests/arkode/C_serial/ark_test_adapt.c
```diff
@@ -0,0 +1,154 @@
...
+            step, local_err_fails);
+    return 1;
+  }
+
+  const N_Vector err = N_VClone(y);
+  ARKodeGetEstLocalErrors(arkode_mem, err);
+  const sunrealtype err_norm = N_VMaxNorm(err);
+  N_VDestroy(err);
+
+  if (err_norm != SUN_RCONST(0))
```
gardner48, line 66: Need the `.0` otherwise this will cause compilation errors 
```suggestion
  if (err_norm != SUN_RCONST(0.0))
```

src/sunadaptcontroller/imexgus/sunadaptcontroller_imexgus.c
Steven-Roberts: Which error should we catch? The base is the result of a norm, so it should never be negative. The exponent could only be `NAN` if the provided method order was not positive or a controller parameters is `NAN`, both of which shouldn't happen in normal usage within SUNDIALS. Maybe some assert macros could be reasonable here.

gardner48: Sounds good, some asserts could catch any unexpected behavior

gardner48 (commented):
The recent updates look good, once the CI is passing and the `TODO` is addressed I think this will be good to merge

gardner48 (changes requested):

doc/shared/RecentChanges.rst
```diff
@@ -37,6 +37,13 @@ Improved the efficiency of default ARKODE methods with the following changes:
 
 **Bug Fixes**
 
+Removed error floors from the ```SUNAdaptController`` implementations
```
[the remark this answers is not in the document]
Steven-Roberts: In the CVODE docs, `SUNAdaptController` is not defined which causes a build error. I'll remove the extra backtick, though

examples/arkode/F2003_serial/ark_bruss_f2003.f90
```diff
@@ -305,6 +305,13 @@ program main
     stop 1
   end if
 
+  ! TODO(SBR): temporary fix until https://github.com/LLNL/sundials/pull/565 merged
```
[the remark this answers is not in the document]
Steven-Roberts: > Does the test fail with the current adaptivity settings?
Yes. When #565 is merged, this fix can be removed.

> Why is this fix needed?
With the current adaptivity adjustment value, the controller tend to overestimate the timestep growth for `dsm<1`. This causes too many rejected steps and the test to fail. For example, if `dsm=0.01` and we have $q=p-1=4$, the timestep will be scaled by 4.6 currently. In #565 it will be 3.2 due to the updated adjustment value.

test/unit_tests/arkode/C_serial/ark_test_mass.c
```diff
@@ -137,7 +137,7 @@ static int solve(const char* im, const char* ex, int steps,
 
   /* Specify a time step so no extra mass evaluations occur from initial step
    * size procedure */
-  retval = ARKodeSetInitStep(arkode_mem, 0.01);
+  retval = ARKodeSetInitStep(arkode_mem, SUN_RCONST(1.0e-5));
```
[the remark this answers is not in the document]
Steven-Roberts: Same issue described in https://github.com/LLNL/sundials/pull/548#discussion_r1954918899. When #565 is merged we can probably change it back

doc/shared/RecentChanges.rst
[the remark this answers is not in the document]
gardner48: In the other guides this will become a link to the online documentation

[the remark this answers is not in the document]
gardner48: When building the docs locally, if Sphinx can't pull down the cross reference index from readthedocs, you can build the "superbuild" first and it will use that.

test/unit_tests/logging/test_logging_arkode_arkstep_1_1_0.out
```diff
@@ -6,35 +6,35 @@ Using GMRES iterative linear solver
...
- 5.069740248736325e-03    1.224742248556792e+00    1.730567522072307e+00    3.889519817334985e-10    2.440982971307903e-10
- 7.964681773481579e-03    1.224738397861497e+00    1.728392193104400e+00    8.783298532932804e-10    1.803091009833224e-10
+ 1.437681224052325e-02    1.224723773776146e+00    1.720158561069919e+00    2.308683466623052e-09    2.193037085262972e-09
+ 2.865063845543699e-02    1.224661090922819e+00    1.685310465644560e+00    4.754813653917722e-09    1.297400853417230e-09
 ------------------------------------------------------------------------------------------------------------------------------
-Current time                  = 0.00796468177348158
+Current time                  = 0.028650638455437
 Steps                         = 3
-Step attempts                 = 3
+Step attempts                 = 5
```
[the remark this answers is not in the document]
Steven-Roberts: The default controller hasn't changed, so it should still be PID here I think. The aggressive growth is partially from removing the `dsm` floor and partially the issue described in https://github.com/LLNL/sundials/pull/548#discussion_r1954918899. Again, #565 is likely to resolve this. I can try a quick test locally to confirm this won't be an issue when the adjustment factor is updated...

src/sunadaptcontroller/imexgus/sunadaptcontroller_imexgus.c
```diff
@@ -128,28 +127,27 @@ SUNErrCode SUNAdaptController_EstimateStep_ImExGus(SUNAdaptController C,
 {
   SUNFunctionBegin(C->sunctx);
 
-  SUNAssert(hnew, SUN_ERR_ARG_CORRUPT);
-
```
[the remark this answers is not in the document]
Steven-Roberts: It's already checked in the "superclass":
https://github.com/LLNL/sundials/blob/d6630f3f90363418f2a0b3978eb18f896caa9927/src/sundials/sundials_adaptcontroller.c#L140

src/sunadaptcontroller/soderlind/sunadaptcontroller_soderlind.c
```diff
@@ -311,33 +299,33 @@ SUNErrCode SUNAdaptController_EstimateStep_Soderlind(SUNAdaptController C,
                                                      sunrealtype* hnew)
 {
   SUNFunctionBegin(C->sunctx);
-  SUNAssert(hnew, SUN_ERR_ARG_CORRUPT);
```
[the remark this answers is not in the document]
Steven-Roberts: Same as https://github.com/LLNL/sundials/pull/548#discussion_r1954962170

test/unit_tests/logging/test_logging_arkode_arkstep_1_1_0.out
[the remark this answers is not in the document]
Steven-Roberts: Ah but it's only taking a few steps so the first couple are I. I see now.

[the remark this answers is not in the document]
gardner48: Has the behavior of the PID controller changed for the first two steps though or are the updates there equivalent to what was done before?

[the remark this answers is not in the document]
Steven-Roberts: The behavior has changed for the first two steps. This was intentional as the old style could, in the worst case, cause the timestep to grow rather than shrink or vise versa.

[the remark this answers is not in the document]
gardner48: Ok, I suspect this change to what happens in the first two steps is what's causing most of the differences in the outputs. Several example have one or two more failed steps than before and I'd guess if we looked at the logs this would probably in the first two steps. Overall though this does seem improve several cases but a few do get worse. Do you think the changes in https://github.com/LLNL/sundials/pull/565 will improve the results here?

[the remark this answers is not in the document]
Steven-Roberts: I agree with that assessment. Yes #565 ought to help. Locally, I tried setting `ARKodeSetAdaptivityAdjustment(arkode_mem, 0);` which is a key change in #565, and this brings the error test failures back to 0.

examples/arkode/C_serial/ark_analytic_lsrk_varjac.out
```diff
@@ -1,3 +1,7 @@
+[ERROR][rank 0][/var/lib/jenkins/workspace/SUNDIALS_PR-548/test/tarballs/sundials-7.2.1/src/arkode/arkode.c:824][ARKodeEvolve] At t = 0.808058307775222, mxstep steps taken before reaching tout.
+
+SUNDIALS_ERROR: ARKodeEvolve() failed with flag = -1
```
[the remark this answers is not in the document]
Steven-Roberts: I can't replicate this locally, so I increased the max step limit so it will at least finish the Jenkins test and we can examine the full out file stats. I think it's just a case of the limit being very close to the actual number of steps taken, and these changes pushed it slightly over.

test/testRunner
```diff
@@ -351,10 +352,8 @@ def isFloat(value):
 
 def floatCompare(answer, value, precision):
     # Check if two floating point numbers are the same to a given number of digits
-    if precision == 0:
-        if answer == value:
-            return True
-    elif round(answer - value, precision) == 0:
+    tol = 10**-precision
+    if math.isclose(answer, value, rel_tol=tol):
```
gardner48, line 356: We still want this check for exact equality when `precision == 0`

examples/arkode/CXX_parallel/ark_brusselator1D_task_local_nls.USE_SERIAL_NVEC_--monitor_--explicit_--tf_1.out
```diff
@@ -25,47 +25,47 @@ Output directory: .
...
-       1.000000     1.617967     2.789700     3.499977
+       0.950000     1.581005     2.848117     3.499966
+       0.975000     1.599823     2.818833     3.499973
+       1.000000     1.617967     2.789700     3.499975
 
 Final Solver Statistics (for processor 0):
-   Internal solver steps = 79612 (attempted = 79835)
-   Total RHS evals:  Fe = 239508   Total number of error test failures = 223
+   Internal solver steps = 78730 (attempted = 110699)
+   Total RHS evals:  Fe = 332100   Total number of error test failures = 31969
```
gardner48, line 71: Steven confirmed this increase in the number of error test failures is fixed by the updated controller defaults in PR #565

examples/arkode/C_parallel/ark_brusselator1D_task_local_nls_--monitor_--explicit_--tf_3.out
```diff
@@ -24,47 +24,47 @@ Output directory: .
...
+       2.775000     0.878688     2.777156     3.499974
+       2.850000     0.831306     2.841860     3.499986
+       2.925000     0.785951     2.907162     3.499997
+       3.000000     0.742699     2.972704     3.499994
 
 Final Solver Statistics (for processor 0):
-   Internal solver steps = 238744 (attempted = 239406)
-   Total RHS evals:  Fe = 718221   Total number of error test failures = 662
+   Internal solver steps = 236112 (attempted = 332018)
+   Total RHS evals:  Fe = 996057   Total number of error test failures = 95906
```
gardner48, line 70: Steven confirmed this increase in the number of error test failures is fixed by the updated controller defaults in PR #565

examples/arkode/F2003_parallel/ark_brusselator1D_task_local_nls_f2003_--monitor_--explicit_--tf_3.out
```diff
@@ -23,50 +23,50 @@ Integrator Options:
...
  
 Final Solver Statistics (for processor 0):
-  Steps            = 238744
-  Step attempts    = 239406
-  Error test fails = 662
-  RHS evals        = 718221
+  Steps            = 236112
+  Step attempts    = 332018
+  Error test fails = 95906
+  RHS evals        = 996057
```
gardner48, line 72: Steven confirmed this increase in the number of error test failures is fixed by the updated controller defaults in PR #565

src/sunadaptcontroller/soderlind/sunadaptcontroller_soderlind.c
```diff
@@ -117,6 +118,16 @@ SUNErrCode SUNAdaptController_SetParams_Soderlind(SUNAdaptController C,
   SODERLIND_K3(C) = k3;
   SODERLIND_K4(C) = k4;
   SODERLIND_K5(C) = k5;
+
+  if (k5 != SUN_RCONST(0.0) || k3 != SUN_RCONST(0.0))
+  {
+    SODERLIND_HISTORYSIZE(C) = 2;
+  }
+  else if (k4 != SUN_RCONST(0.0) || k5 != SUN_RCONST(0.0))
```
gardner48, line 126: Should this check `k2` also?
```suggestion
  else if (k4 != SUN_RCONST(0.0) || k5 != SUN_RCONST(0.0) || k2 != SUN_RCONST(0.0))
```

gardner48 (changes requested):
On minor item otherwise this looks good

src/sunadaptcontroller/soderlind/sunadaptcontroller_soderlind.c
Steven-Roberts: Good catch. It should check `k2` instead of `k5`

gardner48: Should this still check `k5` i.e., `k2` and `k4` are are zero (not that such a setup really makes sense)?

Steven-Roberts: There's no need since it would have entered the previous if statement if k5 was nonzero

gardner48 (approved):

examples/arkode/CXX_parallel/ark_brusselator1D_task_local_nls.USE_SERIAL_NVEC_--monitor_--explicit_--tf_1.out
gardner48: Fixed by commit 40dd2bb

examples/arkode/C_parallel/ark_brusselator1D_task_local_nls_--monitor_--explicit_--tf_3.out
gardner48: Fixed by commit 40dd2bb

examples/arkode/F2003_parallel/ark_brusselator1D_task_local_nls_f2003_--monitor_--explicit_--tf_3.out
gardner48: Fixed by commit 40dd2bb

gardner48 (approved):

PR #462, opened 2024-04-25
Fix #461 - truncated error messages

balos1:

Issue #461, opened 2024-04-25
ARKODE error messages are being cutoff 


Steven-Roberts:
I noticed that some ARKODE error messages are getting cut off, .e.g, `At t = 0.001, mxstep steps taken before reaching tou`. When SUNDIALS is built in double precision, some format specifiers seem to be for long double, which I think is the issue:

https://github.com/LLNL/sundials/blob/2abd63bd6cbc354fb4861bba8e98d0b95d65e24a/src/arkode/arkode_impl.h#L1026-L1032

Looks like the other `_impl.h` files have the same thing.

balos1:
Lower case `l` is not for long double, that would be upper case `L`.  `l` actually [has no effect in this case](https://stackoverflow.com/a/10368369).

balos1:
Is it possible that your terminal is cutting off the output? Maybe try setting `SUNLOGGER_ERROR_FILENAME ` to a file.

Steven-Roberts:
Whoops, looks like it isn't the format specifier then. I saw it could cause undefined behavior, but apparently that was only before C99. After a closer look, the message length here is somehow wrong when I step through in the debugger

https://github.com/LLNL/sundials/blob/2abd63bd6cbc354fb4861bba8e98d0b95d65e24a/src/arkode/arkode.c#L3331

`msgfmt` has a length of 52, but `msgfmt` gets set to 51.

I'm seeing this with gcc 8.5 and 13.1. I might need to boil things down to a minimal example to make sure it's not a secondary effect.

Steven-Roberts:
It could be that `ap` is used twice without `va_copy` or `va_end`: https://stackoverflow.com/q/55274350

balos1:
Can you try this? 
```
diff --git a/src/arkode/arkode.c b/src/arkode/arkode.c
index eac2af422..5cdff8575 100644
--- a/src/arkode/arkode.c
+++ b/src/arkode/arkode.c
@@ -3328,12 +3328,17 @@ void arkProcessError(ARKodeMem ark_mem, int error_code, int line,
   /* Initialize the argument pointer variable
      (msgfmt is the last required argument to arkProcessError) */
   va_list ap;
-  va_start(ap, msgfmt);
 
   /* Compose the message */
+  va_start(ap, msgfmt);
   size_t msglen = vsnprintf(NULL, 0, msgfmt, ap) + 1;
+  va_end(ap);
+
   char* msg     = (char*)malloc(msglen);
+
[...]

PR #549, opened 2024-07-30
Fix variable arguments start/end

Steven-Roberts:
The `test_sundials_errors` test was failing for me which I traced down to the handling of variable arguments. It's the same issue discussed in https://github.com/LLNL/sundials/issues/461 which was partially fixed in https://github.com/LLNL/sundials/pull/462. This PR should finish it off.


balos1:
I fixed this in another branch, lost it in the rebase, and just redid ths fix also! Thanks for opening the PR.

balos1 (approved):

PR #560, opened 2024-08-28
ARKODE Stepper SetDefaults memory leak fix

drreynolds:
Updated stepper `SetDefaults` routines to deallocate pre-existing structures before nullifying, and moved initial creation of the `SUNAdaptController` into stepper `SetDefaults` routines instead of `arkCreate`.  This allows non-adaptive steppers to leave the SUNAdaptController object unallocated.


src/arkode/arkode_erkstep_io.c
```diff
@@ -259,23 +259,42 @@ int erkStep_SetRelaxFn(ARKodeMem ark_mem, ARKRelaxFn rfn, ARKRelaxJacFn rjac)
...
-    ark_mem->liw -= leniw;
-    ark_mem->lrw -= lenrw;
+    ARKodeButcherTable_Space(step_mem->B, &Bliw, &Blrw);
+    ark_mem->liw -= Bliw;
+    ark_mem->lrw -= Blrw;
+    ARKodeButcherTable_Free(step_mem->B);
   }
+  step_mem->B = NULL;
+
+  /* Remove current SUNAdaptController object, and replace with "PI" */
```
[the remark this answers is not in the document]
Steven-Roberts: I see, perfect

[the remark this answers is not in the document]
drreynolds: FWIW, I think that this new structure is more amenable to what Sylvia's currently doing to assess controller performance, in that if we determine controller X works better for nonstiff problems, while controller Y works better for stiff problems, each stepper could make different decisions about their default controller without having to remove any defaults specified by the underlying ARKODE infrastructure.

src/arkode/arkode.c
```diff
@@ -1485,21 +1484,6 @@ ARKodeMem arkCreate(SUNContext sunctx)
   ark_mem->lrw += ARK_ADAPT_LRW;
   ark_mem->liw += ARK_ADAPT_LIW;
 
-  /* Allocate default step controller (PID) and note storage */
-  ark_mem->hadapt_mem->hcontroller = SUNAdaptController_PID(sunctx);
```
gardner48, line 1489: Are there any `Set` functions that could be called after `Create` that assume a controller exists?

src/arkode/arkode_arkstep_io.c
```diff
@@ -677,12 +679,66 @@ int arkStep_SetDefaults(ARKodeMem ark_mem)
   step_mem->msbp     = MSBP;  /* max steps between updates to J or P */
   step_mem->stages   = 0;     /* no stages */
   step_mem->istage   = 0;     /* current stage */
-  step_mem->Be       = NULL;  /* no Butcher tables */
-  step_mem->Bi       = NULL;
-  step_mem->NLS      = NULL; /* no nonlinear solver object */
   step_mem->jcur     = SUNFALSE;
   step_mem->convfail = ARK_NO_FAILURES;
   step_mem->stage_predict = NULL; /* no user-supplied stage predictor */
+
```
gardner48, line 685: Is it assumed that `ReInit` is must be called after `SetDefautls`? Otherwise I think this leaves the integrator in an invalid state i.e., `NULL` tables and nonlinear solver. What should be done with the linear solver?

If `step_mem->explicit` and `step_mem->implicit` above are reset based on `fe` and `fi`, then the default tables and a nonlinear solver could be attached below (like the controller). However, this could still leave the integrator in a bad state e.g., if a fixed-point solver was attached previously and this attaches the default Newton solver it would not have a linear solver.

Maybe this should leave the current methods and solvers in place and only reset integrator parameters. In which case some values above should be retained e.g., `step_mem->explicit`, `step_mem->implicit`, `step_mem->stages`, etc.

src/arkode/arkode_arkstep_io.c
```diff
@@ -677,12 +679,66 @@ int arkStep_SetDefaults(ARKodeMem ark_mem)
...
+  /* Remove pre-existing nonlinear solver object */
+  if (step_mem->NLS && step_mem->ownNLS) { SUNNonlinSolFree(step_mem->NLS); }
+  step_mem->NLS = NULL;
+
+  /* Remove pre-existing SUNAdaptController object, and replace with "PID" */
+  if (ark_mem->hadapt_mem->owncontroller)
+  {
+    retval = SUNAdaptController_Space(ark_mem->hadapt_mem->hcontroller, &lenrw,
+                                      &leniw);
+    if (retval == SUN_SUCCESS)
```
gardner48, line 713: Does `SUNAdaptController` return `0` by default? If so, I think this check can be removed.

src/arkode/arkode_arkstep_io.c
```diff
@@ -677,12 +679,66 @@ int arkStep_SetDefaults(ARKodeMem ark_mem)
...
+  if (ark_mem->hadapt_mem->hcontroller == NULL)
+  {
+    arkProcessError(ark_mem, ARK_MEM_FAIL, __LINE__, __func__, __FILE__,
+                    "SUNAdaptControllerPID allocation failure");
+    return (ARK_MEM_FAIL);
+  }
+  ark_mem->hadapt_mem->owncontroller = SUNTRUE;
+  retval = SUNAdaptController_Space(ark_mem->hadapt_mem->hcontroller, &lenrw,
+                                    &leniw);
+  if (retval == SUN_SUCCESS)
```
gardner48, line 737: Does `SUNAdaptController` return `0` by default? If so, I think this check can be removed.

src/arkode/arkode.c
gardner48: I see, this a little different from what's done with the interpolation modules where attaching the default is done in `arkInitialSetup`. In this case, `SetDefaults` will (always?) create a controller so one will exist after `Create` is called.

gardner48 (changes requested):

src/arkode/arkode.c
drreynolds: Almost.

It is the responsibility of a time-stepper module's `SetDefaults` routine to allocate a controller if time adaptivity should be enabled.  Steppers that do not support adaptivity can leave it `NULL` (e.g., SPRKStep), and ARKODE will now refrain from calling the [`NULL`-valued] controller.

src/arkode/arkode_arkstep_io.c
drreynolds: The modifications to this file change nothing about any assumed order of operations regarding `ReInit` and `SetDefaults` -- the previous version merely nullified the pointer *without ever checking whether memory should be freed*.  This was @Steven-Roberts's main point in issuing the bug report, and that behavior has now been fixed.

If the previous implementation (and now current implementation) assume a specific order of operations, and if you would like to revisit those assumptions, I'd prefer that this question be handled in a separate PR.

src/arkode/arkode.c
gardner48: Sounds good. I think there are some `Set` functions that need to be updated to also check for a `NULL` controller e.g., `ARKodeSetInitStep`, `ARKodeSetErrorBias`, etc.

src/arkode/arkode_arkstep_io.c
drreynolds: The base `SUNAdaptController` class returns whatever integer return value is sent by the implementation.  Both of the provided `SUNAdaptController` implementations (Soderlind and ImExGus) indeed return `0` no matter what, but users could do differently (the API only states that they should return a `SUNErrCode` indicating success or failure).

src/arkode/arkode.c
drreynolds: Good catch.  I just updated the PR to add guards to protect against `NULL`-valued SUNAdaptController objects.

src/arkode/arkode_arkstep_io.c
gardner48: Agreed, this is a separate problem from the memory leak (I've opened a new issue) and can be addressed in a separate PR.

src/arkode/arkode_io.c
```diff
@@ -1,3 +1,4 @@
+
```
gardner48, line 1:
```suggestion
```

gardner48 (changes requested):
Other than minor suggestion to remove an erroneous new line, do `arkPrintAdaptMem` and `arkAdapt` also need check for a `NULL` controller?

drreynolds:
> Other than the minor suggestion to remove an erroneous new line, do `arkPrintAdaptMem` and `arkAdapt` also need checks for a `NULL` controller?

The `NULL` check was indeed necessary in `arkPrintAdaptMem`.  I added it to `arkAdapt` as well just in case, but since any non-adaptive stepper (who leaves a `NULL`-valued SUNAdaptController) must also set `ark_mem->fixedstep` to `SUNTRUE`, that check should be redundant.

PR #563, opened 2024-09-04
Add message about switching to default SPRK order

Steven-Roberts:
This makes SPRKStep's handling of an invalid order consistent with the rest of ARKODE (following #439). Currently an invalid order is an error, but it doesn't leave the integrator in an invalid state or prevent the integration from proceeding. For this reason I wonder if `ARK_WARNING` would be more appropriate for the `error_code`.


balos1 (approved):

balos1 (commented):
ARK_WARNING would probably be better, because then the output would go through the warning handler instead of the error handler. See https://github.com/LLNL/sundials/blob/main/src/arkode/arkode.c#L3555. However, we want the behavior to be consistent across all of the steppers and they seem to use `ARK_ILL_INPUT`.

Steven-Roberts:
I'll go ahead and make them all warnings across ARKODE in this PR then.

gardner48 (approved):

PR #567, opened 2024-09-10
Bugfix: Default IMEX-MRI method

gardner48:
Fix loading the default IMEX-MRI method if `ARKodeSetOrder` is used to specify a third or fourth order method. Previously, the default second order method was loaded in both cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ architecture for their system.
 
 Fixed the loading of ARKStep's default first order explicit method.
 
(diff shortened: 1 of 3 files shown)


drreynolds (commented):
These changes are obsolete, since #564 includes these updates (in addition to new MRI methods).

gardner48:
> These changes are obsolete, since #564 includes these updates (in addition to new MRI methods).

I'll close this PR then but it looks like the change log needs to updated in #564 to note the bug fix.

Issue #569, opened 2024-09-11
[BUG] ARKodeSetDefaults assumes ReInit


gardner48:
### Current Behavior:

As mentioned in #560, when restoring the default settings with `ARKodeSetDefaults` the integrator is put in a state where a call to `ARKReInit` is necessary before calling `Evolve`.   

### Expected Behavior:

This behavior needs to documented or updated such that `Evolve` may be called after `ARKodeSetDefaults`.

PR #517, opened 2024-06-21
Add macro for consistent sunrealtype formatting

Steven-Roberts:
Replaces https://github.com/LLNL/sundials/pull/498 after v7.1.0 update

PR #572, opened 2024-09-13
Feature/operator splitting

Steven-Roberts:


doc/arkode/guide/source/Usage/User_callable.rst
```diff
@@ -1094,6 +1099,30 @@ Set max number of constraint failures             :c:func:`ARKodeSetMaxNumConstr
    .. versionadded:: 6.1.0
 
 
+.. c:function:: int ARKodeSetStepDirection(void* arkode_mem, sunrealtype stepdir)
```
drreynolds, line 1102: I'm surprised that this uses `sunrealtype` instead of `int`.

doc/arkode/guide/source/Usage/User_callable.rst
```diff
@@ -3253,6 +3283,22 @@ Retrieve a pointer for user data                       :c:func:`ARKodeGetUserDat
    .. versionadded:: 6.1.0
 
 
+.. c:function:: int ARKodeGetStepDirection(void *arkode_mem, sunrealtype *stepdir)
```
drreynolds, line 3286: Similar comment as with `ARKodeSetStepDirection`

doc/shared/RecentChanges.rst
```diff
@@ -1,7 +1,14 @@
 **Major Features**
 
+Added an operator splitting module,
+:ref:`SplittingStep <ARKODE.Usage.SplittingStep>`, and forcing method module,
+:ref:`ForcingStep <ARKODE.Usage.ForcingStep>`, to ARKODE.
```
drreynolds, line 5: Same comment as for `CHANGELOG.md`

drreynolds (commented):
I've finished going through the documentation.  I'll start in on the code soon.

doc/arkode/guide/source/Usage/SplittingStep/SplittingStepCoefficients.rst
```diff
@@ -0,0 +1,470 @@
...
+
+      The number :math:`s` of stages
+
+   .. c:member:: int partitions
+
+      The number :math:`P` of partitions in the IVP
+
+   .. c:member:: int order
+
+      The method order of accuracy
```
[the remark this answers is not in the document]
Steven-Roberts: It's only used for determining the order of interpolation now.

doc/arkode/guide/source/Usage/User_callable.rst
Steven-Roberts: Usually the direction comes from something like `tf - t0` which is a `sunrealtype`. A cast to int seems unnecessary and a source of sneaky bugs from converting double to int (undefined behavior if greater than INT_MAX)

CHANGELOG.md
```diff
@@ -4,8 +4,14 @@
 
 ### Major Features
 
+Added an operator splitting module, SplittingStep, and forcing method module,
+ForcingStep. to ARKODE.
```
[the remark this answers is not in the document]
Steven-Roberts: Done

doc/arkode/guide/source/Usage/SplittingStep/SplittingStepCoefficients.rst
```diff
@@ -0,0 +1,470 @@
...
+   **Return value:**
+      * The name associated with *method*.
+      * ``NULL`` pointer if *method* was invalid.
+   
+   .. versionadded:: x.y.z
+
+
+.. c:function:: SplittingStepCoefficients SplittingStepCoefficients_LieTrotter(int partitions)
+
+   Create the coefficients for the first order Lie-Trotter splitting
```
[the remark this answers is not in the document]
Steven-Roberts: I left off the comma because the sentence continues through the equation

doc/arkode/guide/source/Usage/SplittingStep/SplittingStepCoefficients.rst
```diff
@@ -0,0 +1,470 @@
...
+.. c:function:: SplittingStepCoefficients SplittingStepCoefficients_SymmetricParallel(int partitions)
+
+   Create the coefficients for the second order, symmetrized Lie-Trotter
+   splitting
+
+   .. math::
+      y_n = \frac{1}{2} \left( L_h(y_{n-1}) + L^*_h(y_{n-1}) \right),
+
+   where :math:`L_h` is the Lie-Trotter splitting and
+   :math:`L^*_h = L^{-1}_{-h}` is its adjoint.
```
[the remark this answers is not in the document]
Steven-Roberts: I don't see anything wrong with this. The only thing that's identical is the definitions of $L_h, L^*_h$ which is intentional

src/arkode/arkode_splittingstep_coefficients.c
```diff
@@ -0,0 +1,474 @@
...
+  }
+
+  sunrealtype* const* beta_cur = beta;
+  sunrealtype start_cur        = start;
+  /* This is essentially the gamma coefficient from Geometric Numerical
+   * Integration (https://doi.org/10.1007/3-540-30666-8) pg 44-45 scaled by the
+   * current interval */
+  const sunrealtype gamma =
+    diff / (composition_stages - 1 -
+            SUNRpowerR(composition_stages - 1, SUN_RCONST(1.0) / (order - 1)));
```
drreynolds, line 372: Does this pass the CI without a "conversion error"?
```suggestion
            SUNRpowerR((sunrealtype) (composition_stages - 1), SUN_RCONST(1.0) / (order - 1)));
```

src/arkode/arkode_splittingstep_coefficients.c
```diff
@@ -0,0 +1,474 @@
...
+  return SplittingStepCoefficients_ComposeStrang(partitions, order, 5);
+}
+
+/*---------------------------------------------------------------
+  Routine to print a splitting coefficient structure
+  ---------------------------------------------------------------*/
+void SplittingStepCoefficients_Write(const SplittingStepCoefficients coefficients,
+                                     FILE* const outfile)
+{
+  // TODO(SBR): update when https://github.com/LLNL/sundials/pull/517 merged
```
drreynolds, line 440: Comment to remind you of this `TODO(SBR)`

src/arkode/execution_policy/arkode_execution_policy_serial.c
```diff
@@ -0,0 +1,85 @@
...
+  This routine evaluates the policy function fn in serial over
+  the range of sequential methods
+  ---------------------------------------------------------------*/
+static int execute_serial(
+  SUNDIALS_MAYBE_UNUSED const ARKodeSplittingExecutionPolicy policy,
+  const ARKExecutionPolicyFn fn, const N_Vector yn, const N_Vector ycur,
+  const N_Vector tmp, sunrealtype* const alpha, const int sequential_methods,
+  void* const user_data)
+{
+  N_VScale(1, yn, ycur);
```
drreynolds, line 46:
```suggestion
  N_VScale(SUN_RCONST(1.0), yn, ycur);
```

test/unit_tests/arkode/C_serial/ark_test_forcingstep.c
```diff
@@ -0,0 +1,135 @@
...
+static int f_forward_2(const sunrealtype t, const N_Vector y,
+                       const N_Vector ydot, void* const user_data)
+{
+  N_VInv(y, ydot);
+  return 0;
+}
+
+/* Integrates the ODE
+ * 
+ * y' = [t / y] - [1 / y]
```
drreynolds, line 47: I don't understand how this is correct, since shouldn't the full RHS just be the sum of the two RHS functions above?
```suggestion
 * y' = [t / y] + [1 / y]
```

test/unit_tests/arkode/C_serial/ark_test_forcingstep.c
```diff
@@ -0,0 +1,135 @@
...
+
+  err = test_forward(ctx);
+
+  SUNContext_Free(&ctx);
+
+  if (err == 0) { printf("Success\n"); }
+  else { printf("%d Test Failures\n", err); }
+
+  return 0;
+}
```
drreynolds, line 135: Looks like this is missing a newline.
```suggestion
}

```

examples/arkode/C_serial/ark_analytic_partitioned.c
```diff
@@ -0,0 +1,255 @@
...
+
+  /* Print the numerical error and statistics */
+  const N_Vector y_err = N_VClone(y);
+  if (check_flag(y_err, "N_VClone", 0)) { return 1; }
+  N_VLinearSum(SUN_RCONST(1.0), y, -SUN_RCONST(1.0), y_exact, y_err);
+  printf("\nError: %" GSYM "\n", N_VMaxNorm(y_err));
+
+  printf("Final Solver Statistics:\n");
+  flag = ARKodePrintAllStats(arkode_mem, stdout, SUN_OUTPUTFORMAT_TABLE);
+  if (check_flag(&flag, "ARKodePrintAllStats", 1)) { return 1; }
```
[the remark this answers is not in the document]
Steven-Roberts: Just to keep the output simple, but I don't feel strongly about that. I'll add the prints

src/arkode/arkode_forcingstep.c
```diff
@@ -0,0 +1,414 @@
...
+                          (tcur, ycur) or the start of the subsequent step i.e.,
+                          at (tn, yn) = (tcur, ycur) from the end of the last
+                          step
+
+     ARK_FULLRHS_OTHER -> called elsewhere (e.g. for dense output)
+
+  In ForcingStep, we accumulate the RHS functions in ARK_FULLRHS_OTHER mode.
+  Generally, inner steppers will not have the correct yn when this function is
+  called and will not be able to reuse a function evaluation since their state
+  resets at the next ARKodeEvolve call.
```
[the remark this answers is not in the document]
Steven-Roberts: `SUNStepper_Evolve` is not a function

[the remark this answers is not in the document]
drreynolds: Actually, it is: https://github.com/LLNL/sundials/blob/feature/operator-splitting/include/sundials/sundials_stepper.h#L114

But my point is really that a user can attach any valid SUNStepper to be used as an inner stepper for this module, nost just ARKStep, and so the inner stepper may not actually call `ARKodeEvolve` after it is reset.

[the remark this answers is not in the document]
Steven-Roberts: Whoops, I misread that as `SplittingStep_Evolve`. I'll make that change

src/arkode/arkode_splittingstep_coefficients.c
Steven-Roberts: Yes, this passes as is, so I don't thing the cast is necessary.

drreynolds: In that case, you can feel free to ignore some of my other related comments as well.

src/arkode/arkode_splittingstep.c
```diff
@@ -0,0 +1,690 @@
...
+  if (retval != ARK_SUCCESS) { return (retval); }
+
+  retval = splittingStep_SetOrder(ark_mem, 0);
+  if (retval != ARK_SUCCESS) { return retval; }
+
+  if (step_mem->own_policy) { free(step_mem->policy); }
+  step_mem->own_policy = SUNFALSE;
+
+  /* TODO(SBR): This may cause issues if a user calls ARKodeSetDefaults. This
+   * issues affects other ARKODE steppers as well */
```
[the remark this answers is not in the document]
Steven-Roberts: This is tied to https://github.com/LLNL/sundials/issues/569 which will need to be resolved separately of this PR

src/arkode/arkode_splittingstep_coefficients.c
Steven-Roberts: #517 will be merged after this PR

src/arkode/arkode_splittingstep.c
```diff
@@ -0,0 +1,690 @@
...
+
+/*---------------------------------------------------------------
+  This routine checks if all required vector operations are
+  present.  If any of them is missing it returns SUNFALSE.
+  ---------------------------------------------------------------*/
+static sunbooleantype splittingStep_CheckNVector(const N_Vector y)
+{
+  // TODO(SBR): check all ops are correct
+  return y->ops->nvdestroy != NULL && y->ops->nvlinearsum != NULL &&
+         y->ops->nvscale != NULL;
```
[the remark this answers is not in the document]
Steven-Roberts: Yes, `nvdestroy` is unnecessary. The upcoming commit will resolve this.

src/arkode/arkode_forcingstep.c
```diff
@@ -0,0 +1,414 @@
...
+  SUNErrCode err = SUNStepper_Reset(s0, ark_mem->tn, ark_mem->yn);
+  if (err != SUN_SUCCESS) { return err; }
+  err = SUNStepper_SetStepDirection(s0, ark_mem->h);
+  if (err != SUN_SUCCESS) { return err; }
+  err = SUNStepper_SetStopTime(s0, tout);
+  if (err != SUN_SUCCESS) { return err; }
+  err = SUNStepper_Evolve(s0, ark_mem->tn, tout, ark_mem->ycur, &tret,
+                          &stop_reason);
+  if (err != SUN_SUCCESS) { return err; }
+  if (stop_reason < 0) { return stop_reason; }
```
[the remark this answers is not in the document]
Steven-Roberts: I'm not sure I understand why `stop_reason` was added and what values it should take. At least in the ARKStep implementation, it's just returning an error code:
https://github.com/LLNL/sundials/blob/d3ca3b8ad9034f71e2658303aaf990c2363e645c/src/arkode/arkode_arkstep.c#L3200-L3201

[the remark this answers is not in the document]
gardner48: I believe `stop_reason` was needed for the adjoint work as a factor in determining if a checkpoint should be written. We discussed having this be the return value instead of an output but that was problematic for some reason. I don't recall what the issue was. Maybe @balos1 remembers.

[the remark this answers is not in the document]
gardner48: > it's just returning an error code:

This is probably part of the reason. `ARKodeEvolve` returns an `int` value that's specific to ARKODE and not a `SUNErrCode`. Having it return a `SUNErrCode` would require changing return values for similar functions across the integrators to make them consistent which was undesirable.

src/arkode/arkode_arkstep.c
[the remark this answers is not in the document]
Steven-Roberts: ERKStep works now (along with any other ARKode stepper)

src/arkode/arkode_forcingstep.c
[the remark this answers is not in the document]
Steven-Roberts: This has been addressed by mapping any non `SUN_SUCCESS` `SUNErrCode` to the ARKODE int error `ARK_SUNSTEPPER_ERR`. Then users can check `stepper->last_flag` to see the stepper-specific reason for the failure. I'll mark as resolved, but reopen or create a new comment if needed.

doc/arkode/guide/source/Usage/General.rst
```diff
@@ -51,11 +51,13 @@ to the SUNDIALS core header file.
...
-  #include <arkode/arkode_arkstep.h>  // ARKStep provides explicit, implicit, IMEX additive RK methods.
-  #include <arkode/arkode_mristep.h>  // MRIStep provides mutlirate RK methods.
-  #include <arkode/arkode_sprkstep.h> // SPRKStep provides symplectic partition RK methods.
+  #include <sundials/sundials_core.h>      // Provides core SUNDIALS types
+  #include <arkode/arkode_arkstep.h>       // ARKStep provides explicit, implicit, IMEX additive RK methods.
+  #include <arkode/arkode_erkstep.h>       // ERKStep provides explicit RK methods.
+  #include <arkode/arkode_sprkstep.h>      // SPRKStep provides symplectic partitioned RK methods.
+  #include <arkode/arkode_splittingstep.h> // SplittingStep provides operator splitting methods.
+  #include <arkode/arkode_forcingstep.h>   // ForcingStep provides a forcing method.
+  #include <arkode/arkode_mristep.h>       // MRIStep provides multirate RK methods.
```
drreynolds, line 60: Is there some logic behind moving MRIStep to the end of this list?  In the absence of some real reason for doing this, I think that this should retain the chronological ordering, with the new steppers put at the end of the list.

doc/arkode/guide/source/Usage/SplittingStep/SplittingStepCoefficients.rst
```diff
@@ -0,0 +1,470 @@
...
+
+
+   **Arguments:**
+      * ``method`` -- the splitting coefficients identifier.
+
+   **Return value:**
+      * A :c:type:`SplittingStepCoefficients` structure if successful.
+      * A ``NULL`` pointer if *method* was invalid or an allocation error occurred.
+   
+   .. versionadded:: x.y.z
```
drreynolds, line 129: Is there a reason this does not use the updated documentation style (e.g., `:param method:`?

doc/arkode/guide/source/Usage/User_callable.rst
```diff
@@ -47,8 +47,12 @@ E. functions that apply for time-stepping modules that support relaxation Runge-
...
 In the function descriptions below, we identify those that have any of the restrictions B-E above.
 Then in the introduction for each of the stepper-specific documentation sections
-(:numref:`ARKODE.Usage.ARKStep.UserCallable`, :numref:`ARKODE.Usage.ERKStep.UserCallable`,
-:numref:`ARKODE.Usage.MRIStep.UserCallable`, and :numref:`ARKODE.Usage.SPRKStep.UserCallable`)
+(:numref:`ARKODE.Usage.ARKStep.UserCallable`,
+:numref:`ARKODE.Usage.ERKStep.UserCallable`,
+:numref:`ARKODE.Usage.SplittingStep.UserCallable`,
+:numref:`ARKODE.Usage.ForcingStep.UserCallable`,
+:numref:`ARKODE.Usage.MRIStep.UserCallable`,
+and :numref:`ARKODE.Usage.SPRKStep.UserCallable`)
```
drreynolds, line 55: Is there some logic behind inserting the new steppers in the middle of this list?  In the absence of some clear justification, I think that this should retain the chronological ordering, with the new steppers put at the end of the list.

doc/arkode/guide/source/Usage/index.rst
```diff
@@ -37,8 +37,10 @@ ARKODE's time stepping modules, including "relaxation" methods and
 preconitioners.  Following our discussion of these commonalities, we
 separately discuss the usage details that that are specific to each of ARKODE's
 time stepping modules: :ref:`ARKStep <ARKODE.Usage.ARKStep>`,
-:ref:`ERKStep <ARKODE.Usage.ERKStep>`, :ref:`SPRKStep <ARKODE.Usage.SPRKStep>`
-and :ref:`MRIStep <ARKODE.Usage.MRIStep>`.
+:ref:`ERKStep <ARKODE.Usage.ERKStep>`, :ref:`SPRKStep <ARKODE.Usage.SPRKStep>`,
+:ref:`SplittingStep <ARKODE.Usage.SplittingStep>`,
+:ref:`ForcingStep <ARKODE.Usage.ForcingStep>`, and
+:ref:`MRIStep <ARKODE.Usage.MRIStep>`.
```
drreynolds, line 43: Is there some logic behind inserting the new steppers in the middle of this list?  In the absence of some clear justification, I think that this should retain the chronological ordering, with the new steppers put at the end of the list, and MRIStep before SPRKStep.

doc/arkode/guide/source/Usage/index.rst
```diff
@@ -78,4 +80,6 @@ ARKBBDPRE can only be used with NVECTOR_PARALLEL.
    ARKStep/index.rst
    ERKStep/index.rst
    SPRKStep/index.rst
+   SplittingStep/index.rst
+   ForcingStep/index.rst
    MRIStep/index.rst
```
drreynolds, line 85: Is there some logic behind inserting the new steppers in the middle of this list?  In the absence of some clear justification, I think that this should retain the chronological ordering, with the new steppers put at the end of the list, and MRIStep before SPRKStep.

examples/arkode/C_serial/ark_analytic_partitioned.c
```diff
@@ -0,0 +1,263 @@
...
+
+/* Compute the exact analytic solution */
+static N_Vector exact_sol(const N_Vector y0, const sunrealtype tf,
+                          const UserData* const user_data)
+{
+  N_Vector sol             = N_VClone(y0);
+  const sunrealtype y0_val = NV_Ith_S(y0, 0);
+  const sunrealtype lambda = user_data->lambda;
+  NV_Ith_S(sol, 0)         = lambda * y0_val /
+                     (y0_val - (y0_val - lambda) * SUNRexp(lambda * tf));
```
drreynolds, line 85: I thought that in the arkode-sts PR #541, you requested that @maggul's new examples avoid all uses of `NV_Ith_S`?

drreynolds (changes requested):

examples/arkode/C_serial/ark_analytic_partitioned.c
cswoodward: We want to remove the use of NV_Ith in all examples.

doc/arkode/guide/source/Mathematics.rst
```diff
@@ -572,6 +574,110 @@ using a fixed time-step size.
 .. The `ark_kepler.c` example demonstrates an implementation of such controller.
 
 
+.. _ARKODE.Mathematics.SplittingStep:
+
+SplittingStep -- Operator splitting methods
+================================================
+
```
drreynolds, line 581: Similar question as elsewhere -- why were the new steppers inserted before MRIStep?  As everywhere else, I see no reason to jumble the order.  These should be listed chronologically: ARKStep, ERKStep, MRIStep, SPRKStep, SplittingStep/ForcingStep/LSRKStep, ...

Steven-Roberts: It made sense to group MRI, splitting, and forcing in the docs and I put the new methods first since they are simpler. I can switch to chronological, but I'll note many place put SPRKStep before MRIStep. Should I reorder SPRKStep while I'm at it?

examples/arkode/C_serial/ark_analytic_partitioned.c
Steven-Roberts: Hmm, I think that was David who requested that. Anyway, `NV_Ith_S` has been removed.

doc/arkode/guide/source/Mathematics.rst
drreynolds: We can discuss this as a group, but I think that ordering them either chronologically or most-to-least-used would be optimal.  In the latter case, we have multiple applications who use ARKStep, and a few who use ERKStep and MRIStep (with more coming).  I see no reason to insert new and currently-unused steppers in front, forcing users to scroll down to get to the documentation that they need.

Steven-Roberts:
In preparation of #587 being merged, I'm wondering how to best count RHS evals. Currently, it can't be accessed through the `SUNStepper` API, and for some implementations, e.g., exact solution, no RHS may even be used. I see two options:

- Add new `SUNStepper_SetGetNumRhsEvalsFn` and `SUNStepper_GetNumRhsEvals` functions and use these in `ARKodeGetNumRhsEvals`
- Have `ARKodeGetNumRhsEvals` return an error or 0 evals.

I suppose I favor the first option. Anyone else have ideas or preferences?

Steven-Roberts:
> I think that you could leave it out entirely, and ARKODE will then return an error that the function is unsupported.

Ok, that sounds reasonable. On further consideration, my original idea of getting RHS evals from the SUNStepper becomes very complicated if nested partitioning is used. Probably best to avoid that.

src/arkode/arkode_io.c
```diff
@@ -1286,6 +1286,58 @@ int ARKodeSetFixedStep(void* arkode_mem, sunrealtype hfixed)
...
+  ark_mem = (ARKodeMem)arkode_mem;
+
+  if (stepdir == ZERO) { return ARK_SUCCESS; }
+
+  retval = ARKodeGetStepDirection(arkode_mem, &h);
+  if (retval != SUN_SUCCESS) { return retval; }
+
+  // TODO(SBR): use SUNRcopysign once merged from other PR
+  // if (SUNRcopysign(h, stepdir) == h) {
+  if (h == ZERO || ((h > 0) == (stepdir > 0))) { return ARK_SUCCESS; }
```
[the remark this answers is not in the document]
Steven-Roberts: I'm going to mark as resolved since the PR with `SUNRcopysign` will be merged after this one. The current implementation is fine for now.

doc/arkode/guide/source/Constants.rst
```diff
@@ -317,6 +317,30 @@ contains the ARKODE output constants.
...
+   +-----------------------------------------------+------------------------------------------------------------+
+   | :index:`ARKODE_SPLITTING_SUZUKI_3_3_2`        | 3rd order Suzuki splitting for problems with two           |
+   |                                               | partitions.                                                |
+   +-----------------------------------------------+------------------------------------------------------------+
+   | :index:`ARKODE_SPLITTING_RUTH_3_3_2`          | 3rd order Ruth splitting for problems with two partitions. |
+   +-----------------------------------------------+------------------------------------------------------------+
+   | :index:`ARKODE_SPLITTING_YOSHIDA_4_4_2`       | 4th order Yoshida splitting for problems with two          |
+   |                                               | partitions.                                                |
+   +-----------------------------------------------+------------------------------------------------------------+
+   | :index:`ARKODE_SPLITTING_YOSHIDA_8_6_2`       | 6th order Yoshida splitting for problems with two          |
```
balos1, line 339: It would be interesting to compare the operator split implementation of the methods in this table that are the same as ones in SPRKStep.  The comparison is not needed for the PR though.

doc/arkode/guide/source/Usage/User_callable.rst
balos1: Note to self: might be able to use this function in the adjoint stepping.

doc/shared/sunstepper/SUNStepper_Description.rst
```diff
@@ -158,6 +158,17 @@ Stepping Functions
    :return: A :c:type:`SUNErrCode` indicating success or failure.
 
 
+.. c:function:: SUNErrCode SUNStepper_SetStepDirection(SUNStepper stepper, sunrealtype stepdir)
```
balos1, line 161: Note to self: might be able to leverage this function in the Adjoint stepping.

examples/arkode/C_serial/ark_analytic_partitioned.c
```diff
@@ -0,0 +1,263 @@
...
+  const char* const coefficients_name = (argc > 2) ? argv[2] : NULL;
+
+  /* Problem parameters */
+  const sunrealtype t0           = SUN_RCONST(0.0);  /* initial time */
+  const sunrealtype tf           = SUN_RCONST(1.0);  /* final time */
+  const sunrealtype dt           = SUN_RCONST(0.01); /* outer time step */
+  const sunrealtype dt_linear    = dt / 5;  /* linear integrator time step */
+  const sunrealtype dt_nonlinear = dt / 10; /* nonlinear integrator time step */
+
+  UserData user_data = {.lambda = SUN_RCONST(2.0)};
```
balos1, line 152: Since its just a C example, its fine, but beware that designated initializers like this one won't work with a C++ compiler unless using C++20 or later.

src/arkode/arkode_forcingstep_impl.h
```diff
@@ -0,0 +1,30 @@
...
+ *---------------------------------------------------------------
+ * This header defines the step memory for ForcingStep.
+ *--------------------------------------------------------------*/
+
+#ifndef ARKODE_FORCINGSTEP_IMPL_H_
+#define ARKODE_FORCINGSTEP_IMPL_H_
+
+#include <sundials/sundials_stepper.h>
+
+#define PARTITIONS 2
```
balos1, line 22:
```suggestion
#define NUM_PARTITIONS 2
```

src/arkode/arkode_splittingstep_coefficients.def
```diff
@@ -0,0 +1,68 @@
...
+    return coefficients;
+ })
+
+ARK_SPLITTING_COEFFICIENTS(ARKODE_SPLITTING_YOSHIDA_4_4_2, {
+    return SplittingStepCoefficients_TripleJump(2, 4);
+ })
+
+ARK_SPLITTING_COEFFICIENTS(ARKODE_SPLITTING_YOSHIDA_8_6_2, {
+    return SplittingStepCoefficients_TripleJump(2, 6);
+ })
```
balos1, line 68: add newline

src/arkode/arkode_splittingstep.c
```diff
@@ -0,0 +1,630 @@
...
+                    "steppers = NULL illegal.");
+    return NULL;
+  }
+
+  for (int i = 0; i < partitions; i++)
+  {
+    if (steppers[i] == NULL)
+    {
+      arkProcessError(NULL, ARK_ILL_INPUT, __LINE__, __func__, __FILE__,
+                      "steppers[i] = NULL illegal.", i);
```
balos1, line 459:
```suggestion
                      "steppers[%d] = NULL illegal.", i);
```

balos1 (changes requested):
A few minor items to address, but this looks great!

src/arkode/arkode_splittingstep_coefficients.def
Steven-Roberts: Not sure what you mean. The file already ends with an empty line, although the GitHub file view hides it.

balos1 (dismissed):

balos1:
LGTM. Ill leave it to others to chime in on anything else.

doc/arkode/guide/source/Usage/SplittingStep/Skeleton.rst
```diff
@@ -0,0 +1,115 @@
...
+#. *Set vector of initial values*
+
+#. Create a stepper object for each problem partition
+
+   * If using ARKStep as an inner integrator, create the ARKStep object with
+     :c:func:`ARKStepCreate` and configure the integrator as desired for
+     evolving the partition. See sections :numref:`ARKODE.Usage.Skeleton`,
+     :numref:`ARKODE.Usage.OptionalInputs`, and
+     :numref:`ARKODE.Usage.ARKStep.OptionalInputs` for details on configuring
+     ARKStep.
```
drreynolds, line 45: Should this paragraph be so specific to ARKStep?  Perhaps a final sentence could be added with something like "We note that other ARKODE steppers may also be used as an inner integrator, using similar steps."

drreynolds (commented):
One minor recommendation.  Otherwise, this looks good.

test/unit_tests/arkode/CXX_serial/ark_test_splittingstep.cpp
```diff
@@ -0,0 +1,441 @@
...
+  const auto coefficients = SplittingStepCoefficients_SuzukiFractal(2, 6);
+  SplittingStep_SetCoefficients(arkode_mem, coefficients);
+  SplittingStepCoefficients_Free(coefficients);
+
+  auto tret = t0;
+  ARKodeEvolve(arkode_mem, tf, y, &tret, ARK_NORMAL);
+
+  /* Check that the solution matches the exact solution */
+  const auto exact_solution     = std::exp(t0 - tf);
+  const auto numerical_solution = NV_Ith_S(y, 0);
```
gardner48, line 397: Remove `NV_Ith_S`
```suggestion
  const auto numerical_solution = N_VGetArrayPointer(y)[0];
```

gardner48 (changes requested):
Mostly minor comments. One general point that came up is we should be careful not to over use `const` when it does provide meaningful protection.

src/arkode/arkode_forcingstep.c
```diff
@@ -0,0 +1,415 @@
...
+
+     ARK_FULLRHS_OTHER -> called elsewhere (e.g. for dense output)
+
+  In ForcingStep, we accumulate the RHS functions in ARK_FULLRHS_OTHER mode.
+  Generally, inner steppers will not have the correct yn when this function is
+  called and will not be able to reuse a function evaluation since their state
+  resets at the next SUNStepper_Evolve call.
+  ----------------------------------------------------------------------------*/
+static int forcingStep_FullRHS(const ARKodeMem ark_mem, const sunrealtype t,
+                               const N_Vector y, const N_Vector f,
```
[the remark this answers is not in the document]
Steven-Roberts: > Since const does not have much effect when applied to an `N_Vector`

It has the benefit of preventing reassignment, e.g., `y = NULL`. This is following the standard and rules for const; C's const syntax for typedef'd pointers is a bit strange, though. I'm fine removing it from the example code but think it is somewhat useful in `src`.

src/arkode/arkode_splittingstep_coefficients.c
```diff
@@ -0,0 +1,474 @@
...
+  return coefficients;
+}
+
+/*---------------------------------------------------------------
+  Routine to create splitting coefficients which performs a copy
+  of the alpha and beta parameters
+  ---------------------------------------------------------------*/
+SplittingStepCoefficients SplittingStepCoefficients_Create(
+  const int sequential_methods, const int stages, const int partitions,
+  const int order, sunrealtype* const alpha, sunrealtype* const beta)
```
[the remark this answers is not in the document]
Steven-Roberts: Correct. `sunrealtype const * const alpha` would be ideal, but that would be inconsistent with what's used in Butcher tables and MRI coupling coefficients.

examples/arkode/C_serial/ark_advection_diffusion_reaction_splitting.c
```diff
@@ -0,0 +1,351 @@
...
+            funcname);
+    return 1;
+  }
+
+  return 0;
+}
+
+/* f routine to compute the advection RHS function. */
+static int f_advection(const sunrealtype t, const N_Vector y,
+                       const N_Vector ydot, void* const user_data)
```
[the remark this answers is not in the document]
Steven-Roberts: I agree for the examples

src/arkode/arkode_splittingstep_coefficients.c
```diff
@@ -0,0 +1,474 @@
...
+    free(coefficients->alpha);
+    if (coefficients->beta != NULL)
+    {
+      free(coefficients->beta[0][0]);
+      free(coefficients->beta[0]);
+      free(coefficients->beta);
+    }
+  }
+
+  free(coefficients);
```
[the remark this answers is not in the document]
Steven-Roberts: It's generally considered good practice *not* to check for `NULL` when calling free. It's unnecessary and adds clutter.

doc/arkode/guide/source/Usage/SplittingStep/SplittingStepCoefficients.rst
```diff
@@ -0,0 +1,432 @@
...
+   :param order: The method order of accuracy.
+   :param alpha: An array of length ``sequential_methods`` containing the
+      weight of each sequential method used to produce the overall operator
+      splitting solution.
+   :param beta: An array of length
+      ``sequential_methods * (stages+1) * partitions`` containing the time nodes
+      of the inner integrations in the order
+
+      .. math::
+         \beta_{1,0,1}, \dots, \beta_{1,0,P}, \beta_{1,1,1}, \dots, \beta_{1,1,P}, \dots, \beta_{2,0,1}, \dots, \beta_{r,s,P}.
```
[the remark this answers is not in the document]
Steven-Roberts: It's either this or have `j = 1, ..., s+1` which also somewhat odd with respect to the end index. I like the current because an index of `j=0` indicates the start time, similar to how we use $t_0$ as the start of the time span.

doc/arkode/guide/source/Usage/User_callable.rst
Steven-Roberts: `signbit` splits by negative and non-negative but doesn't indicate when 0. For that reason, I don't think we could use it directly.

src/arkode/arkode_splittingstep_coefficients.c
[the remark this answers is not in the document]
gardner48: This is the style used in all the other `Free` and `Delete` functions and we have encountered compilers or memory checking tools in the past that complain about freeing `NULL` pointers (even though it's allowed)

doc/arkode/guide/source/Usage/SplittingStep/SplittingStepCoefficients.rst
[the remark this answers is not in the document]
gardner48: I think having all the indexing starting from 1 is a bit more clear even if it does not align as nicely with `t_0`

doc/arkode/guide/source/Mathematics.rst
```diff
@@ -729,6 +652,191 @@ may supply their own method by defining and attaching a coupling table, see
...
+:math:`\phi^k_{h}` is the flow map for partition :math:`k`:
+
+.. math::
+   \phi^k_{h}(y_{n_1}) = v(t_n),
+   \quad \begin{cases}
+      v(t_{n-1}) = y_{n-1}, \\ \dot{v} = f_k(t, v).
+   \end{cases}
+
+SplittingStep provides standard operator splitting methods such as Lie-Trotter
+and Strang splitting, as well as schemes of arbitrarily high order.
```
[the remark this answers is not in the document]
Steven-Roberts: I've added the examples and citations: https://sundials--572.org.readthedocs.build/en/572/arkode/Mathematics_link.html#splittingstep-operator-splitting-methods. Marking as resolved, but you can add more review comments to new docs if needed.

doc/arkode/guide/source/Usage/SplittingStep/SplittingStepCoefficients.rst
```diff
@@ -0,0 +1,432 @@
...
+   Deallocate the splitting coefficients memory.
+
+   :param coefficients: The splitting coefficients.
+   
+   .. versionadded:: x.y.z
+
+
+.. c:function:: void SplittingStepCoefficients_Write(SplittingStepCoefficients coefficients, FILE* outfile)
+
+   Write the splitting coefficients to the provided file pointer.
```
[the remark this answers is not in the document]
Steven-Roberts: I agree it can be moved out of the note, but I wonder if it belongs in the `:param outfile:` description rather than the overall function description.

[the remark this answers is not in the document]
gardner48: Agreed, I think that would be better

src/arkode/arkode_forcingstep.c
```diff
@@ -0,0 +1,415 @@
...
+
+  const SUNStepper s0    = step_mem->stepper[0];
+  const sunrealtype tout = ark_mem->tn + ark_mem->h;
+  sunrealtype tret       = 0;
+
+  /* Evolve stepper 0 on its own */
+  SUNErrCode err = SUNStepper_Reset(s0, ark_mem->tn, ark_mem->yn);
+  if (err != SUN_SUCCESS) { return ARK_SUNSTEPPER_ERR; }
+  err = SUNStepper_SetStepDirection(s0, ark_mem->h);
+  if (err != SUN_SUCCESS) { return ARK_SUNSTEPPER_ERR; }
```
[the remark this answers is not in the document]
Steven-Roberts: Suppose forcingstep is used to integrate forward for some time then `ARKodeReset` and `ARKodeSetStepDirection(..., -1)` are called on the forcingstep memory. Before evolving again (now backward in time), the partition integrators need to reverse direction. We can either have forcingstep do it or require the user to do it. Since splittingstep internally must call `SUNStepper_SetStepDirection` to handle direction changes, it made sense for forcingstep to do the same.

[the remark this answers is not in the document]
Steven-Roberts: If this makes sense, I'll add what I wrote as a comment in the code.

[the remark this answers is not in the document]
gardner48: The user should have to call `ARKodeSetStepDirection` if they want to change the integration direction but they shouldn't have to call it on each of the `SUNSteppers` individually just the "overall" `arkode_mem`. For this steppers could provide a `setstepdirection` function to ARKODE that `ARKodeSetStepDirection` will call (if it exists) to perform any additional actions needed. So the ForcingStep version of this function would call `SUNStepper_SetStepDirection` on each `SUNStepper` (and similarly for SpilttingStep and MRIStep) while ERKStep, ARKStep, etc. would not need to set this function pointer in the ARKODE memory.

This also makes me realize we that we should add a check in `ARKodeSteSetDirection` to ensure it's only called after `Reset` (or before the first `Evolve` call). This can be done by checking if `ark_mem->initialized` is true like in `ARKodeSetInterpolantType`.

src/arkode/arkode_splittingstep.c
```diff
@@ -0,0 +1,635 @@
...
+                          const SUNContext sunctx)
+{
+  if (steppers == NULL)
+  {
+    arkProcessError(NULL, ARK_ILL_INPUT, __LINE__, __func__, __FILE__,
+                    "steppers = NULL illegal.");
+    return NULL;
+  }
+
+  for (int i = 0; i < partitions; i++)
```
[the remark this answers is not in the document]
Steven-Roberts: I went with the extra validation route. Supporting 1 partition will require special cases and testing for an edge case that is only really useful for debugging. It can be added later if we find a strong need.

src/arkode/arkode_splittingstep.c
```diff
@@ -0,0 +1,635 @@
...
+  ARKodeSplittingStepMem step_mem = NULL;
+  int retval = splittingStep_AccessStepMem(ark_mem, __func__, &step_mem);
+  if (retval != ARK_SUCCESS) { return retval; }
+
+  *nflagPtr = ARK_SUCCESS; /* No algebraic solver */
+  *dsmPtr   = ZERO;        /* No error estimate */
+
+  const SplittingStepCoefficients coefficients = step_mem->coefficients;
+
+  N_VScale(ONE, ark_mem->yn, ark_mem->ycur);
```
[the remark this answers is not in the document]
Steven-Roberts: Closing, but I made a separate issue so this doesn't get lost.

test/unit_tests/arkode/CXX_serial/ark_test_splittingstep.cpp
```diff
@@ -0,0 +1,441 @@
...
+  ARKodeSStolerances(parititon_mem[0], local_tol, local_tol);
+  ARKodeSStolerances(parititon_mem[1], local_tol, local_tol);
+
+  SUNStepper steppers[] = {nullptr, nullptr};
+  ARKodeCreateSUNStepper(parititon_mem[0], &steppers[0]);
+  ARKodeCreateSUNStepper(parititon_mem[1], &steppers[1]);
+
+  auto arkode_mem = SplittingStepCreate(steppers, 2, t0, y, ctx);
+  ARKodeSetFixedStep(arkode_mem, dt);
+  const auto coefficients = SplittingStepCoefficients_SymmetricParallel(2);
```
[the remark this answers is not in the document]
Steven-Roberts: I generalized all the other test function to allow other methods, but I'd like to keep this one as is. I want to include some testing of parallel methods which can only be constructed with function call like this are are a bit awkward to generalize. Also, resizing shouldn't really depend on the particular coefficients.

src/arkode/arkode_io.c
```diff
@@ -1299,41 +1299,75 @@ int ARKodeSetFixedStep(void* arkode_mem, sunrealtype hfixed)
...
+    if (h != ZERO && ((h > ZERO) != (stepdir > ZERO)))
+    {
+      /* Reverse the sign of h. If adaptive, h will be overwritten anyway by the
+       * initial step estimation since ARKodeReset must be called before this.
+       * However, the sign of h will be used to check if the integration
+       * direction and stop time are consistent, e.g., in ARKodeSetStopTime, so
+       * we should not set h = 0. */
+      ark_mem->h = -h;
+      /* Clear previous initial step and force an initial step recomputation.
+       * Normally, this would not occur after a reset, but is is necessary here
```
gardner48, line 1344:
```suggestion
       * Normally, this would not occur after a reset, but it is necessary here
```

test/unit_tests/arkode/C_serial/ark_test_forcingstep.c
```diff
@@ -95,9 +102,174 @@ static int test_forward(SUNContext ctx)
...
+  ARKodeSStolerances(parititon_mem[0], local_tol, local_tol);
+  ARKodeSStolerances(parititon_mem[1], local_tol, local_tol);
+
+  SUNStepper steppers[] = {NULL, NULL};
+  ARKodeCreateSUNStepper(parititon_mem[0], &steppers[0]);
+  ARKodeCreateSUNStepper(parititon_mem[1], &steppers[1]);
+
+  void* arkode_mem = ForcingStepCreate(steppers[0], steppers[1], t0, y, ctx);
+  ARKodeSetFixedStep(arkode_mem, dt);
+  // ARKodeSetInterpolantType(arkode_mem, ARK_INTERP_HERMITE);
```
gardner48, line 159: Uncomment or remove?

gardner48 (changes requested):

src/arkode/arkode_splittingstep.c
```diff
@@ -403,88 +389,144 @@ static void splittingStep_PrintMem(const ARKodeMem ark_mem, FILE* const outfile)
...
+  if (partitions <= 1)
+  {
+    arkProcessError(ark_mem, ARK_ILL_INPUT, __LINE__, __func__, __FILE__,
+                    "The number of partitions must be greater than one");
+    return ARK_ILL_INPUT;
   }
 
   for (int i = 0; i < partitions; i++)
   {
     if (steppers[i] == NULL)
```
drreynolds, line 455: As with ForcingStep, the documentation for SplittingStep states that the provided steppers must implement all but one operation.  Those required operations should be checked here inside the loop.

test/unit_tests/arkode/C_serial/ark_test_forcingstep.c
```diff
@@ -95,9 +102,174 @@ static int test_forward(SUNContext ctx)
...
+}
+
+/* Integrates the ODE
+ * 
+ * y' = [t / y] + [1 / y]
+ * 
+ * with initial condition y(0) = 1 and partitioning specified by the square
+ * brackets. We integrate to t = 1 then reinitialize the forcing method by
+ * swapping the SUNSteppers and updating the initial condition to y(1) = -1.
+ * Next we integrator to t = 2 and check the error against the exact solution
```
drreynolds, line 208:
```suggestion
 * Next we integrate to t = 2 and check the error against the exact solution
```

drreynolds (commented):
I think that this is looking great!  I have only a few minor recommendations.

src/arkode/arkode_forcingstep.c
```diff
@@ -278,57 +322,79 @@ static void forcingStep_PrintMem(const ARKodeMem ark_mem, FILE* const outfile)
...
-                        sunrealtype t0, N_Vector y0, SUNContext sunctx)
+/*------------------------------------------------------------------------------
+  This routine validates arguments when (re)initializing a ForcingStep
+  integrator
+  ----------------------------------------------------------------------------*/
+static int forcingStep_CheckArgs(ARKodeMem ark_mem, SUNStepper stepper1,
+                                 SUNStepper stepper2, N_Vector y0)
 {
   if (stepper1 == NULL)
   {
```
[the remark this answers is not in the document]
Steven-Roberts: Yes, good point

drreynolds (dismissed):

src/arkode/arkode_splittingstep.c
```diff
@@ -429,6 +443,17 @@ static sunbooleantype splittingStep_CheckNVector(N_Vector y)
   return y->ops->nvlinearsum != NULL && y->ops->nvscale != NULL;
 }
 
+/*------------------------------------------------------------------------------
+  This routine checks if all required SUNStepper operations are present. If any
+  of them are missing it return SUNFALSE.
+  ----------------------------------------------------------------------------*/
+static sunbooleantype forcingStep_CheckSUNStepper(SUNStepper stepper)
```
gardner48, line 450:
```suggestion
static sunbooleantype splittingStep_CheckSUNStepper(SUNStepper stepper)
```

src/arkode/arkode_splittingstep.c
```diff
@@ -458,6 +483,14 @@ static int splittingStep_CheckArgs(ARKodeMem ark_mem, SUNStepper* steppers,
                       "steppers[%d] = NULL illegal.", i);
       return ARK_ILL_INPUT;
     }
+
+    if (!forcingStep_CheckSUNStepper(steppers[i]))
```
gardner48, line 487:
```suggestion
    if (!splittingStep_CheckSUNStepper(steppers[i]))
```

gardner48 (changes requested):
A couple minor items, otherwise the changes look good. The Jenkins CI is failing when building the PDF docs and the answers submodule will need to be updated.

gardner48 (approved):

gardner48 (approved):

PR #573, opened 2024-09-16
fix missing semicolon

balos1:
Fixes https://github.com/xsdk-project/xsdk-issues/issues/266.


gardner48 (approved):
Any idea why the CI didn't catch this (it's using Ginkgo 1.8)?

gardner48:
> Any idea why the CI didn't catch this (it's using Ginkgo 1.8)?

Nevermind, is see it's specific to HIP. So we need the GitLab CI to catch this.

PR #577, opened 2024-09-23
add magic /autofix ability for PRs

balos1:
This wont do anything until it is propagated into main, but I have tested it out in my sundials fork and it seems to work.  The idea is that you can comment `/autofix` on your PR and a GitHub action will fix the spelling, formatting, and swig errors.


.github/workflows/check-format.yml
```diff
@@ -56,9 +67,50 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      # Checkout the GitHub created reference for the PR.
+      # The only way to do this is by using the "issue" number
+      # but that is really the PR number in this context.
+      # This is due to using an `issue_comment` event which
+      # is due to the GitHub Actions API that does not have
+      # a `pull_request_comment` event or something similar.
+      # This leaves us in a detached head state which is corrected
+      # in `apply-style/checkout.sh`
```
Steven-Roberts, line 89: Perhaps you've already considered this, but here's a possible way to avoid the extra checkout script:
https://github.com/actions/checkout/issues/331#issuecomment-1120113003

balos1: Unfortunately, the [hub command is no longer available by default on the runners](https://github.com/actions/runner-images/issues/8362) so that wont work.

drreynolds (dismissed):
This looks fantastic.  I'm excited to try it out.

.github/workflows/check-format.yml
gardner48, line 1: Add new line at end of file

gardner48 (commented):
Will this work for PRs from forks as well?

balos1:
> Will this work for PRs from forks as well?

I believe so.

gardner48 (approved):

Issue #581, opened 2024-09-28
[BUG] out-of-bounds write in SUNMatScaleAddI_Sparse


dweindl:
### Current Behavior:

For the example below, I encounter out-of-bounds writes in `SUNMatScaleAddI_Sparse` at

https://github.com/LLNL/sundials/blob/c28eaa3764a03705d61decb6025b409360e9d53f/src/sunmatrix/sparse/sunmatrix_sparse.c#L753

### Expected Behavior:

No out-of-bounds writes.

### Steps To Reproduce:

* Download [test_SUNMatScaleAddI_Sparse.tar.gz](https://github.com/user-attachments/files/17173831/test_SUNMatScaleAddI_Sparse.tar.gz)
* Unpack: `tar -xzf test_SUNMatScaleAddI_Sparse.tar.gz && cd test_SUNMatScaleAddI_Sparse`
* Configure & build: `cmake -S. -Bbuild && cmake --build build` (this might need some adaptation to find the sundials configuration)
* `valgrind build/sparsetest`:
  
  ```
   Invalid write of size 8
      at 0x10C882: SUNMatScaleAddI_Sparse (sunmatrix_sparse.c:753)
      by 0x111B9C: SUNMatScaleAddI (sundials_matrix.c:201)
      by 0x109630: main (in build/sparsetest)
    Address 0x4a8f588 is 8 bytes before a block of size 312 alloc'd
      at 0x484D953: calloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
      by 0x109D07: SUNSparseMatrix (sunmatrix_sparse.c:130)
      by 0x10943F: main (in build/sparsetest)
   
   Invalid write of size 8
      at 0x10C8C4: SUNMatScaleAddI_Sparse (sunmatrix_sparse.c:754)
      by 0x111B9C: SUNMatScaleAddI (sundials_matrix.c:201)
      by 0x109630: main (in build/sparsetest)
    Address 0x4a8f408 is 8 bytes before a block of size 312 alloc'd
      at 0x484D953: calloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
      by 0x109C81: SUNSparseMatrix (sunmatrix_sparse.c:127)
      by 0x10943F: main (in build/sparsetest)
  ```

### Environment:

- SUNDIALS version: 7.1.1
- OS: Ubuntu 24.04
- Compiler: gcc 13.2.0

### Anything else:

[...]

PR #584, opened 2024-10-02
Bugfix/sparse add i

Steven-Roberts:
Fixes https://github.com/LLNL/sundials/issues/581

The new implementation eliminates all temporary storage/mallocs and only traverses the matrix entries once when diagonal elements don't need to be added. The sparse matrix example shows about a 1.5x speedup on 50000x50000 matrices.

### DIFF
--- a/doc/shared/RecentChanges.rst
+++ b/doc/shared/RecentChanges.rst
@@ -20,12 +20,13 @@ which should ease building SUNDIALS with LAPACK libraries that require setting
 specific linker flags e.g., MKL.
 
 **Bug Fixes**
-Fixed a `bug <https://github.com/LLNL/sundials/issues/581>`_ in 
-:c:func:`SUNMatScaleAddI_Sparse` which caused out of bounds writes unless
-``indexvals`` were in ascending order for each row/column.
 
-Fixed c:func:`ARKodeResize` not using the default ``hscale`` when an argument of
-``0`` was provided.
+Fixed a `bug <https://github.com/LLNL/sundials/issues/581>`__ in the sparse
+matrix implementation of :c:func:`SUNMatScaleAddI` which caused out of bounds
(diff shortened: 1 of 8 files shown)


drreynolds (commented):
I like the updated implementation.  

Did you test this with the `main` routine that @dweindl provided?  Did you do any tests to verify that this works correctly when the original matrix lacks storage for the diagonal (this is not currently tested in `Test_SUNMatScaleAddI` from `examples/sunmatrix/test_sunmatrix.c`).

dweindl:
That looks great, @Steven-Roberts, thanks!

I used your revised implementation to run our model simulation tests suite from over at https://github.com/AMICI-dev/AMICI/ - where I originally ran into #581 - and did not notice any issues.

drreynolds (dismissed):

doc/shared/RecentChanges.rst
```diff
@@ -20,6 +20,9 @@ which should ease building SUNDIALS with LAPACK libraries that require setting
 specific linker flags e.g., MKL.
 
 **Bug Fixes**
+Fixed a `bug <https://github.com/LLNL/sundials/issues/581>`_ in
```
balos1, line 23: Looks like sphinx is not happy about the `bug` target name for the link (I did not realize they had to be unique) and `SUNMatScaleAddI_Sparse` is not recognized as a function (I think the docs for it are in the old format where we do not fully write out the docs for the functions which implement the generic operations).

Steven-Roberts: Looks like I needed to use an anonymous link which doesn't make a target. It uses a double underscore instead of a single. I went ahead and switched to that for all external links in the changelog.

https://stackoverflow.com/a/27435441

balos1 (approved):

PR #593, opened 2024-10-17
Extending Precision for Some DIRK Methods

Steven-Roberts:
The Billington, Cash, and Kvaerno DIRK method coefficients only had ~10-14 digits of accuracy which can be insufficient for double and extended precision. This PR extends the coefficients to 40 digits.


drreynolds (changes requested):
These look great.  Can you add a comment to the `RecentChanges` and `CHANGELOG` files to state that the precision of these coefficients has been increased?

drreynolds (approved):

PR #608, opened 2024-12-05
Enable more warning flags which caught a couple small bugs

Steven-Roberts:
Tested new warning flags on gcc and clang. Not sure if this will work on Windows compilers (if that's something we intend to support)

### DIFF
--- a/cmake/SundialsSetupCompilers.cmake
+++ b/cmake/SundialsSetupCompilers.cmake
@@ -86,6 +86,8 @@ if(ENABLE_ALL_WARNINGS)
       "-Wno-unknown-warning-option -Wall -Wpedantic -Wextra -Wshadow \
 -Wwrite-strings -Wcast-align -Wdisabled-optimization -Wvla -Walloca \
 -Wduplicated-cond -Wduplicated-branches")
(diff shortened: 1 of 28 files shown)


cmake/SundialsSetupCompilers.cmake
```diff
@@ -79,26 +79,31 @@ endif()
 if(ENABLE_ALL_WARNINGS)
   message(STATUS "Enabling all compiler warnings")
 
+  # Some warning flags are not supported by all compilers so ignore unknown
+  # flags with -Wno-unknown-warning-option. Ironically, this is not supported by
+  # some compilers.
+  set(WARNING_FLAGS
+      "-Wno-unknown-warning-option -Wall -Wpedantic -Wextra -Wshadow \
+-Wwrite-strings -Wcast-align -Wdisabled-optimization -Wvla -Walloca \
+-Wduplicated-cond -Wduplicated-branches")
```
gardner48, line 88: We might be able to add back `-Wredundant-decls` when we update the SuperLU_DIST version in the GitHub CI tests. It would be good to add a TODO here or an issue to revisit this when we make that update.

gardner48 (approved):

PR #610, opened 2024-12-06
Bugfix: Memory leaks

gardner48:
* Fix memory leak when attaching an H-Tol controller to MRIStep
* Fix memory leak / out of bound array access in Lagrange interpolation module
* Fix memory leaks in unit tests

### DIFF
--- a/src/arkode/arkode_io.c
+++ b/src/arkode/arkode_io.c
@@ -3239,8 +3239,7 @@ int arkReplaceAdaptController(ARKodeMem ark_mem, SUNAdaptController C,
     }
     ark_mem->hadapt_mem->owncontroller = SUNTRUE;
   }
-  else if (take_ownership) { ark_mem->hadapt_mem->owncontroller = SUNTRUE; }
-  else { ark_mem->hadapt_mem->owncontroller = SUNFALSE; }
+  else { ark_mem->hadapt_mem->owncontroller = take_ownership; }
 
   /* Attach new SUNAdaptController object */
   retval = SUNAdaptController_Space(C, &lenrw, &leniw);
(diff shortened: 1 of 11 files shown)


drreynolds (commented):
These changes all look good to me, but it would be nice if there existed a unit test to exercise the new "ownership" change for the adaptivity controller.

src/arkode/arkode_io.c
```diff
@@ -3238,6 +3239,7 @@ int arkReplaceAdaptController(ARKodeMem ark_mem, SUNAdaptController C)
     }
     ark_mem->hadapt_mem->owncontroller = SUNTRUE;
   }
+  else if (take_ownership) { ark_mem->hadapt_mem->owncontroller = SUNTRUE; }
   else { ark_mem->hadapt_mem->owncontroller = SUNFALSE; }
```
Steven-Roberts, line 3243:
```suggestion
  }
  ark_mem->hadapt_mem->owncontroller = take_ownership;
```

gardner48: I think we should keep this as it is. The proposed logic would allow passing `NULL` to get the default controller and also `SUNFALSE` to tell ARKODE not to take ownership which would likely cause a memory leak.

gardner48:
> These changes all look good to me, but it would be nice if there existed a unit test to exercise the new "ownership" change for the adaptivity controller.

This is already covered by the current unit tests which fail when address sanitizers are enabled because the wrapper is leaked.

src/arkode/arkode_io.c
Steven-Roberts: Ah, I got the brace levels mixed up. I suppose it could still be shortened a bit, but I'll leave that up to you
```suggestion
    ark_mem->hadapt_mem->owncontroller = SUNTRUE;
  }
  else { ark_mem->hadapt_mem->owncontroller = take_ownership; }
```

Steven-Roberts (approved):

drreynolds (approved):

Issue #596, opened 2024-10-29
[BUG] Compilation failure for SYCL with Intel oneAPI 2025.0


baperry2:
### Current Behavior:

The PelePhysics CI is failing while compiling Sundials with SYCL support. The error is reproduced below - the `std::sqrt` function is not found. This is new behavior for Intel oneAPI 2025.0 (released 10/25/24) and does not occur for older versions.  

```
/home/runner/work/PelePhysics/PelePhysics/PelePhysics-sycl/Submodules/sundials/src/nvector/sycl/nvector_sycl.cpp:1149:10: error: no member named 'sqrt' in namespace 'std'; did you mean 'sycl::sqrt'?
 1149 |   return std::sqrt(sum / N);
      |          ^~~~~~~~~
      |          sycl::sqrt
/opt/intel/oneapi/compiler/2025.0/bin/compiler/../../include/sycl/detail/builtins/math_functions.inc:188:34: note: 'sycl::sqrt' declared here
  188 | BUILTIN_GENF_NATIVE_OPT(ONE_ARG, sqrt)
      |                                  ^
/home/runner/work/PelePhysics/PelePhysics/PelePhysics-sycl/Submodules/sundials/src/nvector/sycl/nvector_sycl.cpp:1195:10: error: no member named 'sqrt' in namespace 'std'; did you mean 'sycl::sqrt'?
 1195 |   return std::sqrt(sum / N);
      |          ^~~~~~~~~
      |          sycl::sqrt
/opt/intel/oneapi/compiler/2025.0/bin/compiler/../../include/sycl/detail/builtins/math_functions.inc:188:34: note: 'sycl::sqrt' declared here
  188 | BUILTIN_GENF_NATIVE_OPT(ONE_ARG, sqrt)
      |                                  ^
/home/runner/work/PelePhysics/PelePhysics/PelePhysics-sycl/Submodules/sundials/src/nvector/sycl/nvector_sycl.cpp:1236:10: error: no member named 'sqrt' in namespace 'std'; did you mean 'sycl::sqrt'?
 1236 |   return std::sqrt(N_VWSqrSumLocal_Sycl(x, w));
      |          ^~~~~~~~~
      |          sycl::sqrt
/opt/intel/oneapi/compiler/2025.0/bin/compiler/../../include/sycl/detail/builtins/math_functions.inc:188:34: note: 'sycl::sqrt' declared here
[...]

PR #617, opened 2024-12-10
Fix SYCL vector for oneapi 2025

balos1:
Fix for #596

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,9 @@ inner stepper object, `ARKodeCreateMRIStepInnerStepper`.
 
 ### Bug Fixes
 
+Fixed a build failure with the SYCL NVector when using Intel oneAPI 2025.0
+compilers. See GitHub Issue [#596](https://github.com/LLNL/sundials/issues/596).
+
 Fixed a bug where `CVodeSetProjFailEta` would ignore the `eta` parameter.
 
 Fixed a bug in the SPTFQMR linear solver where recoverable preconditioner errors
--- a/doc/shared/RecentChanges.rst
+++ b/doc/shared/RecentChanges.rst
@@ -75,6 +75,9 @@ inner stepper object, :c:func:`ARKodeCreateMRIStepInnerStepper`.
 
 **Bug Fixes**
 
+Fixed a build failure with the SYCL NVector when using Intel oneAPI 2025.0
+compilers. See GitHub Issue `#596 <https://github.com/LLNL/sundials/issues/596>`__.
+
 Fixed a bug where :c:func:`CVodeSetProjFailEta` would ignore the `eta`
 parameter.
 


src/nvector/sycl/nvector_sycl.cpp
```diff
@@ -2285,14 +2285,14 @@ static int FusedBuffer_CopyRealArray(N_Vector v, sunrealtype* rdata, int nval,
     return SUN_ERR_GENERIC;
   }
 
-  sunrealtype* h_buffer = (sunrealtype*)((char*)(vcp->fused_buffer_host->ptr) +
-                                         vcp->fused_buffer_offset);
+  sunrealtype* h_buffer = reinterpret_cast<sunrealtype*>(
```
Steven-Roberts, line 2288: It looks like the buffer is used to store `sunrealtype` and/or `sunrealtype*` which may have different alignments. I'm concerned that `shortcut` may have improper alignment and cause undefined behavior, but that depends on how the buffer is filled. Also, I think `static_cast` or placement new would be more idiomatic.

.github/workflows/ubuntu-latest-oneapi.yml
```diff
@@ -0,0 +1,45 @@
...
+  build:
+    strategy:
+      matrix:
+        ONEAPI_VERSION: [
+          '2024.2.1-0-devel-ubuntu22.04',
+          'latest' # 2025.0.0-0-devel-ubuntu24.04 at the time of commit
+        ]
+
+    runs-on: ubuntu-latest
+    container: intel/oneapi-basekit:${{ matrix.ONEAPI_VERSION }}
```
gardner48, line 26: Where are these containers hosted? I thought they'd be at https://github.com/orgs/intel/packages but I don't see them there.

balos1: They are hosted in the [default docker registry](https://hub.docker.com/r/intel/oneapi-basekit).

src/nvector/sycl/nvector_sycl.cpp
balos1: I believe the buffers are filled appropriately as we use the same code in all of the GPU vectors (so it has been tested extensively). The reason for the change to `reinterpret_cast` is because the dpcpp compiler emits a warning about possible realignment (due to the conversion from `char*` to `sunrealtype*`/`sunrealtype**`) when using the C style cast (I will try `static_cast`).   However, the first cast to `char*` is done purely to for the purpose of doing the pointer arithmetic with the offset which was already computed appropriately based on the underlying data type. So it should be safe.

Steven-Roberts: I think conversion to `char*` is fine, but the use of the buffer for two datatypes of potentially different sizes is my concern. Consider 32 bit `sunrealtype` and 64 bit pointers. If we write an odd number number of `sunrealtype` to the buffer, then get a `sunrealtype*` pointer at the current offset, it will not have 64 bit alignment. For many compilers, this may happen to work, but it is not guaranteed by the standard and could impact performance.

balos1: I see what you are saying, but I guess the first issue is that the calculation of the offset does not account for alignment. If it did, then we would have two different offsets potentially and would need two different buffers.

Steven-Roberts: Ok, that sounds reasonable.

gardner48 (changes requested):
Looks good to me, just need to note the SYCL fix in the change log

PR #618, opened 2024-12-10
Maintenance: Rename set function

gardner48:
Update set function name to be consistent with `SetNum` naming convention


gardner48:
Missed this in the initial LSRKStep PR review but noticed it when adding the logging tests for LSRKStep

drreynolds (approved):

PR #625, opened 2024-12-16
CI: Update Testing Options

gardner48:
* Always enable testing
* Consistent naming of CMake testing options
* Add warnings/errors for some testing option combinations
* Add documentation for testing options
* Move testing README into testing docs
* Cleanup using default testing options
* Enable running test in containers without enabling development tests
* Simplify setting pre and post executable run commands/options in sundials_add_test
* Remove redundant Jenkins tests
* Remove precision dependent diff option in unit tests


cmake/macros/SundialsAddBenchmark.cmake
```diff
@@ -29,62 +29,57 @@ macro(sundials_add_benchmark NAME EXECUTABLE BASE_BENCHMARK_NAME)
   # ENABLE_GPU        = indicate this benchmark should be run with GPUs
   set(options ENABLE_GPU)
 
-  cmake_parse_arguments(sundials_add_benchmark "${options}" "${oneValueArgs}"
-                        "${multiValueArgs}" ${ARGN})
+  cmake_parse_arguments(arg "${options}" "${oneValueArgs}" "${multiValueArgs}"
```
balos1, line 32: Is it possible that this could cause variable name conflict, e.g., if you call another macro from within this one that has an argument of the same name?

cmake/SundialsBuildOptionsPre.cmake
```diff
@@ -311,26 +311,85 @@ endif()
 # Options for SUNDIALS testing
 # ---------------------------------------------------------------
 
+sundials_option(SUNDIALS_TEST_ENABLE_DEV_TESTS BOOL "Include development tests"
+                OFF ADVANCED)
+
+sundials_option(SUNDIALS_TEST_ENABLE_UNIT_TESTS BOOL "Include unit tests" OFF
```
balos1, line 317: This would be a breaking option name change, yes?

gardner48: I've added the old options to `SundialsDeprecated.cmake`

cmake/macros/SundialsAddBenchmark.cmake
gardner48: Yes it could. I've switched this and `sundials_add_test` from a `macro` to a `function`. Side note, I got this idea to simplify the prefix looking at some BLT macros and I suspect they might run into the same scoping issues.

balos1 (approved):

balos1 (approved):

PR #635, opened 2025-01-08
guard zeroing of yerr with compensated sums check

balos1:


src/arkode/arkode_sprkstep.c
```diff
@@ -232,7 +232,9 @@ int SPRKStepReInit(void* arkode_mem, ARKRhsFn f1, ARKRhsFn f2, sunrealtype t0,
   step_mem->istage = 0;
 
   /* Zero yerr for compensated summation */
-  N_VConst(ZERO, step_mem->yerr);
+  if (ark_mem->use_compensated_sums) {
+    N_VConst(ZERO, step_mem->yerr);
+  }
```
gardner48, line 237: Looks like `sprkStep_Reset` needs this same check as well

gardner48 (changes requested):

gardner48:
Looks like the Swig test is failing because of `ubuntu-latest` switching from `ubuntu-22.04` to `ubuntu-24.04`. It seems `pcre` is not installed by default now.

CHANGELOG.md
```diff
@@ -8,13 +8,16 @@
 
 ### Bug Fixes
 
+Fixed bug in the ARKODE SPRKStep `SPRKStepReInit` function that could cause a
+segmentation fault when compensated summation is not used.
```
gardner48, line 12:
```suggestion
Fixed bug in the ARKODE SPRKStep `SPRKStepReInit` function and `ARKodeReset` function
with SPRKStep that could cause a segmentation fault when compensated summation is not
used.
```

doc/shared/RecentChanges.rst
```diff
@@ -4,4 +4,7 @@
 
 **Bug Fixes**
 
+Fixed bug in the ARKODE SPRKStep :c:func:`SPRKStepReInit` function that could cause a
+segmentation fault when compensated summation is not used.
```
gardner48, line 8:
```suggestion
Fixed bug in the ARKODE SPRKStep :c:func:`SPRKStepReInit` function and
:c:func:`ARKodeReset` function with SPRKStep that could cause a segmentation
fault when compensated summation is not used.
```

gardner48 (changes requested):
Minor update to the recent changes otherwise this looks good

PR #646, opened 2025-01-17
Fix /autofix command

balos1:
Currently, our `/autofix` command does not work because the jobs don't have the 'write' permission.  This will fix it, but not until the change ends up in `main`.

This permission can be set globally too, but this is more secure.

### DIFF
--- a/.github/workflows/check-format.yml
+++ b/.github/workflows/check-format.yml
@@ -6,6 +6,9 @@ on:
   issue_comment:
     types: [created]
 
+permissions:
+  contents: write
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
   cancel-in-progress: true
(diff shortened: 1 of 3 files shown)


gardner48 (dismissed):

.github/workflows/check-format.yml
```diff
@@ -16,6 +16,8 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/llnl/sundials_spack_cache:llvm-17.0.4-h4lflucc3v2vage45opbo2didtcuigsn.spack
+    permissions:
+      contents: write
```
gardner48, line 20: Does this need to be added to the `apply_format` job as well?

balos1: It only needs to be under `apply_format`. I pasted it into the wrong job.

gardner48 (approved):

PR #636, opened 2025-01-08
Update copyright year

Steven-Roberts:
I noticed a couple dozen files had copyright end data <2024. I guess those were from PRs that started in one year but were merged the next. There are probably a few situations like that now, e.g., ERK adjoint, where we'll need to run the update copyright script.

PR #647, opened 2025-01-17
Reorganized organization of checkNvector routines

drreynolds:
This PR cleans up how ARKODE and its steppers check for required N_Vector routines.  Previously, ARKODE required routines that were only actually needed when certain options were configured, and each stepper performed redundant checks on the supplied N_Vector, when in reality, no stepper utilized routines that were not already required by the main ARKODE infrastructure.

In this PR:
1. Stepper-specific N_Vector operation checks have been removed.
2. ARKODE performs an initial check on the supplied N_Vector for operations that are always required.
3. Individual ARKODE utilities perform their own N_Vector checks when they are configured.
4. ARKODE then performs a final check after user configuration is complete to check any remaining N_Vector operation requirements.


src/arkode/arkode.c
```diff
@@ -1851,23 +1859,71 @@ sunbooleantype arkCheckTimestepper(ARKodeMem ark_mem)
...
 
+/*---------------------------------------------------------------
+  arkCheckNvector2:
+
+  This routine perform conditional checks on required vector
+  operations are present (i.e., if the current ARKODE
+  configuration requires additional N_Vector routines).  If any
+  of them is missing it returns SUNFALSE.
+  ---------------------------------------------------------------*/
+sunbooleantype arkCheckNvector2(ARKodeMem ark_mem)
```
Steven-Roberts, line 1887: I'd recommend more descriptive name to distinguish it from `arkCheckNvector`. Maybe call them `arkCheckNvectorRequired` and `arkCheckNvectorOptional`

src/arkode/arkode.c
```diff
@@ -272,6 +272,14 @@ int ARKodeSStolerances(void* arkode_mem, sunrealtype reltol, sunrealtype abstol)
     return (ARK_ILL_INPUT);
   }
 
+  /* Ensure that vector supports N_VAddConst */
+  if (!ark_mem->tempv1->ops->nvaddconst)
```
Steven-Roberts, line 276: I assume it doesn't matter but just want to double check this doesn't need to be `ark_mem->yn`.

gardner48, line 276: This also needs to be checked in `arkCheckNvector2` in case the user is running with the default tolerances (in which case this function is not called)

gardner48: This same check is needed in `ARKodeResStolerance` and a similar check in `arkCheckNvector2`

src/arkode/arkode.c
```diff
@@ -1851,23 +1859,71 @@ sunbooleantype arkCheckTimestepper(ARKodeMem ark_mem)
...
-sunbooleantype arkCheckNvector(N_Vector tmpl) /* to be updated?? */
+sunbooleantype arkCheckNvector(N_Vector tmpl)
 {
   if ((tmpl->ops->nvclone == NULL) || (tmpl->ops->nvdestroy == NULL) ||
       (tmpl->ops->nvlinearsum == NULL) || (tmpl->ops->nvconst == NULL) ||
       (tmpl->ops->nvdiv == NULL) || (tmpl->ops->nvscale == NULL) ||
       (tmpl->ops->nvabs == NULL) || (tmpl->ops->nvinv == NULL) ||
-      (tmpl->ops->nvaddconst == NULL) || (tmpl->ops->nvmaxnorm == NULL) ||
-      (tmpl->ops->nvwrmsnorm == NULL))
+      (tmpl->ops->nvmaxnorm == NULL) || (tmpl->ops->nvwrmsnorm == NULL))
```
gardner48, line 1872: I think the `nvmaxnorm` check can also be removed here. It's only needed if we're estimating the initial step size and that check is covered in `arkCheckNvector2`.

src/arkode/arkode_arkstep.c
```diff
@@ -66,15 +65,6 @@ void* ARKStepCreate(ARKRhsFn fe, ARKRhsFn fi, sunrealtype t0, N_Vector y0,
...
   }
 
-  /* Test if all required vector operations are implemented */
-  nvectorOK = arkStep_CheckNVector(y0);
-  if (!nvectorOK)
-  {
-    arkProcessError(NULL, ARK_ILL_INPUT, __LINE__, __func__, __FILE__,
-                    MSG_ARK_BAD_NVECTOR);
-    return (NULL);
-  }
```
gardner48, line 76: [suggestion] To fail faster, maybe call `arkCheckNvector` here (and similarly in the other constructors) instead of waiting until `arkInit` is called at the end of this function

src/arkode/arkode.c
gardner48: I agree these names would be clearer

src/arkode/arkode_ls.c
```diff
@@ -114,7 +114,8 @@ int ARKodeSetLinearSolver(void* arkode_mem, SUNLinearSolver LS, SUNMatrix A)
   /* Check for compatible LS type, matrix and "atimes" support */
   if (iterative)
   {
-    if (ark_mem->tempv1->ops->nvgetlength == NULL)
+    if ((ark_mem->tempv1->ops->nvgetlength == NULL) ||
+        (ark_mem->tempv1->ops->nvdotprod == NULL))
```
gardner48, line 118: Move the `nvdotprod` check to `ARKodeSetLSNormFactor` and `ARKodeSetMassLSNormFactor` as it's only needed if the input to those functions is < 0.

gardner48 (changes requested):

src/arkode/arkode.c
drreynolds: Either is fine -- this just needs to be a vector that was allocated inside `arkAllocVectors` (both `yn` and `tempv1` are allocated there)

src/arkode/arkode_arkstep.c
gardner48: I agree it would be preferable to have this check in one place. Since `arkInit` is at the end `arkAllocVec` should check that `clone` is defined since the constructors call it before the operation checks happen

gardner48 (changes requested):
Other than the comment [here](https://github.com/LLNL/sundials/pull/647#discussion_r1923100985) about checking for `clone` this looks good

src/arkode/arkode.c
```diff
@@ -3397,6 +3397,9 @@ int arkCheckTemporalError(ARKodeMem ark_mem, int* nflagPtr, int* nefPtr,
   ---------------------------------------------------------------*/
 sunbooleantype arkAllocVec(ARKodeMem ark_mem, N_Vector tmpl, N_Vector* v)
 {
+  /* return failure if N_VClone is not implemented */
+  if (!tmpl->ops->nvclone) { return SUNFALSE; }
```
gardner48, line 3401: I should have thought of this when noting the check here but we should similarly check if `ops->nvdestroy` is defined in `arkFreeVec` in case `ARKodeFree` is called before `arkInit`

gardner48 (changes requested):
Minor change I should have noticed before

drreynolds:
I don't understand why the "spelling" check just failed.  The patch file that it produces "fixes" numerous variable names in completely unrelated files (`examples/cvodes/parallel/cvsAdvDiff_ASAp_non_p.c`, `examples/kinsol/CUDA_mpi/kin_em_mpicuda.cu`, `src/cvodes/cvodea.c`, and `src/idas/idaa.c`).

Did something change in the CI that could have caused this?

Steven-Roberts:
> I don't understand why the "spelling" check just failed.

codespell had an update today which caught a few more things. I have a fix in #636 so you can just wait for that to be merged. It might be good to select a fixed version of codespell and manually update it as need to avoid issues like this.

gardner48 (approved):

drreynolds:
> > I don't understand why the "spelling" check just failed.
> 
> codespell had an update today which caught a few more things. I have a fix in #636 so you can just wait for that to be merged. It might be good to select a fixed version of codespell and manually update it as need to avoid issues like this.

Actually, the PR had no typos to fix, it's just failing the CI due to teh codespell issue.  I don't see how this PR can be merged before #636, given that the "spelling" check is required before any PR can be merged, so it can just wait.

Steven-Roberts:
Yes, codespell recognizes `lins` and `indx` as typos now which are used in code totally unrelated to this PR.

Steven-Roberts (approved):

PR #657, opened 2025-01-29
Simplify addressing of CI errors

Steven-Roberts:
Uniquely names artifacts for a matrix of jobs to so there's no conflicts like https://github.com/LLNL/sundials/actions/runs/13020815131/job/36320839738#step:7:16
The changes will allow us to get out files across all logging levels and precisions with on run rather than repeated runs and updates. That's been quite tedious on a couple recent PRs.

Any reason we can't remove limits like this:
https://github.com/LLNL/sundials/blob/2de70893c6360b0bb5047e86f3b0b882324c03fa/.github/workflows/ubuntu-latest.yml#L25

### DIFF
--- a/.github/workflows/check-format.yml
+++ b/.github/workflows/check-format.yml
@@ -64,7 +64,7 @@ jobs:
           ./scripts/format.sh CMakeLists.txt benchmarks cmake examples include src test
 
       - name: Run git diff to see if anything changed
-        run: /usr/bin/git diff --name-only --exit-code
+        run: /usr/bin/git diff --exit-code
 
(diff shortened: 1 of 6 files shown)


gardner48 (approved):

gardner48:
> Any reason we can't remove limits like this:
> 
> https://github.com/LLNL/sundials/blob/2de70893c6360b0bb5047e86f3b0b882324c03fa/.github/workflows/ubuntu-latest.yml#L25

There are a limited number of runners for an organization and this helps prevent longer jobs like this from taking over all the resources.

PR #659, opened 2025-01-31
Fix return value in ARKodeSetFixedStep

Steven-Roberts:


gardner48 (changes requested):
Looks good, note the fix in the change log

drreynolds (dismissed):

gardner48 (approved):

PR #661, opened 2025-02-05
fixed a typo in README and in a function input argument

maggul:
Fixed a typo in README and in a function input argument

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ nonlinear algebraic systems:
 
 * ARKODE - for integrating stiff, nonstiff, and multirate ODEs of the form
 
-  $$M(t) \, y' = f_1(t,y) + f_2(t,y), \quad y(t_0) = y_0$$
+  $$M(t) y' = f_1(t,y) + f_2(t,y), \quad y(t_0) = y_0$$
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -19,7 +19,7 @@ nonlinear algebraic systems:
 
 * ARKODE - for integrating stiff, nonstiff, and multirate ODEs of the form
 
-  $$M(t) \\, y' = f_1(t,y) + f_2(t,y), \quad y(t_0) = y_0$$
+  $$M(t) \, y' = f_1(t,y) + f_2(t,y), \quad y(t_0) = y_0$$
```
balos1, line 22: We do not want the comma. I dont think we really need a space between `M(t)` and `y'` either so I recommend removing the backslashes completely. 
```suggestion
  $$M(t)y' = f_1(t,y) + f_2(t,y), \quad y(t_0) = y_0$$
```

balos1 (changes requested):

README.md
maggul: It seems to render differently on different systems. In the codespace of github, my change was doing what we really mean without comma. I will remove those as well to make sure it will render correctly on the one you pointed.

balos1 (approved):

PR #662, opened 2025-02-06
Command-line control over scalar-valued set routines

drreynolds:
I believe this PR is now complete, although we should hold off on reviews until the current release is out.


include/arkode/arkode.h
```diff
@@ -214,6 +214,10 @@ typedef enum
  * Shared API routines
  * -------------------------- */
 
+/* Command-line control over ARKODE options */
+SUNDIALS_EXPORT int ARKodeSetFromCommandLine(void* arkode_mem, int argc,
+                                             char* argv[]);
```
Steven-Roberts, line 219:
```suggestion
SUNDIALS_EXPORT int ARKodeSetFromCommandLine(void* arkode_mem, int argc,
                                             const char* argv[]);
```

src/arkode/arkode_io.c
```diff
@@ -34,6 +35,347 @@
...
+                           ARKodeSetLSetupFrequency,
+                           ARKodeSetPredictorMethod,
+                           ARKodeSetMaxNonlinIters,
+                           ARKodeSetMaxHnilWarns,
+                           ARKodeSetInterpolateStopTime,
+                           ARKodeSetMaxNumConstrFails,
+                           ARKodeSetAdaptivityAdjustment,
+                           ARKodeSetSmallNumEFails,
+                           ARKodeSetMaxErrTestFails,
+                           ARKodeSetMaxConvFails};
```
Steven-Roberts, line 169:
```suggestion
  static const char* int_keys[] = {"arkode.order",
                            "arkode.interpolant_degree",
                            "arkode.linear",
                            "arkode.autonomous",
                            "arkode.deduce_implicit_rhs",
                            "arkode.lsetup_frequency",
                            "arkode.predictor_method",
                            "arkode.max_nonlin_iters",
                            "arkode.max_hnil_warns",
                            "arkode.interpolate_stop_time",
                            "arkode.max_num_constr_fails",
                            "arkode.adaptivity_adjustment",
                            "arkode.small_num_efails",
                            "arkode.max_err_test_fails",
                            "arkode.max_conv_fails"};
  static arkIntSetFn int_set[] = {ARKodeSetOrder,
                           ARKodeSetInterpolantDegree,
                           ARKodeSetLinear,
                           ARKodeSetAutonomous,
                           ARKodeSetDeduceImplicitRhs,
                           ARKodeSetLSetupFrequency,
                           ARKodeSetPredictorMethod,
                           ARKodeSetMaxNonlinIters,
                           ARKodeSetMaxHnilWarns,
                           ARKodeSetInterpolateStopTime,
                           ARKodeSetMaxNumConstrFails,
                           ARKodeSetAdaptivityAdjustment,
                           ARKodeSetSmallNumEFails,
                           ARKodeSetMaxErrTestFails,
                           ARKodeSetMaxConvFails};
  static const int num_int_keys = sizeof(int_keys) / sizeof(*int_keys);
```
and similar changes for code below...
To avoid issues with maintaining consistent lengths of these arrays, a cleaner solution might be to have a single array of `struct { char* name; arkIntSetFn fun}`.
Alternatively, this seems like an ideal place to use SUNDIALS' hashmap.

src/arkode/arkode_io.c
```diff
@@ -34,6 +35,347 @@
...
+typedef int (*arkRealSetFn)(void*,sunrealtype);
+typedef int (*arkBoolSetFn)(void*);
+static int CheckAndSetIntArg(ARKodeMem ark_mem, int *i, char *argv[],
+                             const char* argtest, arkIntSetFn fname,
+                             sunbooleantype *arg_used)
+{
+  *arg_used = SUNFALSE;
+  if (strcmp(argv[*i], argtest) == 0)
+  {
+    int iarg = atoi(argv[++(*i)]);
```
Steven-Roberts, line 56: This usage of `++` is confusing. Can it be done on a separate line?

src/arkode/arkode_io.c
```diff
@@ -34,6 +35,347 @@
...
+  }
+  return ARK_SUCCESS;
+}
+
+static int CheckAndSetBoolArg(ARKodeMem ark_mem, int *i, char *argv[],
+                              const char* argtest, arkBoolSetFn fname,
+                              sunbooleantype *arg_used)
+{
+  *arg_used = SUNFALSE;
+  if (strcmp(argv[*i], argtest) == 0)
```
Steven-Roberts, line 114: So booleans are specified with just the flag? I think it would be preferable to have flag-value pair command line syntax for boolean options. Otherwise, a user will need to know the default and whether to override it.

src/arkode/arkode_io.c
```diff
@@ -34,6 +35,347 @@
...
+
+  int num_bool_keys = 4;
+  const char* bool_keys[] = {"arkode.nonlinear",
+                             "arkode.clear_stop_time",
+                             "arkode.no_inactive_root_warn",
+                             "arkode.reset_accumulated_error"};
+  arkBoolSetFn bool_set[] = {ARKodeSetNonlinear,
+                             ARKodeClearStopTime,
+                             ARKodeSetNoInactiveRootWarn,
+                             ARKodeResetAccumulatedError};
```
Steven-Roberts, line 221: It seems there's two type of boolean set functions to consider. Ones like `ARKodeResetAccumulatedError` don't take a boolean as input while ones like `ARKodeSetAutonomous` do. Those might need to be handled separately in this parsing.

src/arkode/arkode_io.c
```diff
@@ -34,6 +35,347 @@
...
+
+  int i, j, retval;
+  for (i = 1; i < argc; i++)
+  {
+    sunbooleantype arg_used = SUNFALSE;
+
+    /* skip command-line arguments that do not begin with "arkode." */
+    if (strncmp(argv[i], "arkode.", 7) != 0) { continue; }
```
Steven-Roberts, line 229: Since we filter by the `arkode.` prefix here, I think it's redundant to include it in all the array entries above. We would just need to shift the string pointer below.

Steven-Roberts, line 229: Removes "magic constant"
```suggestion
    static const char *prefix = "arkode.";
    if (strncmp(argv[i], prefix, strlen(prefix)) != 0) { continue; }
```

src/arkode/arkode_io.c
```diff
@@ -34,6 +35,347 @@
...
+      if (arg_used) break;
+    }
+    if (arg_used) continue;
+
+    /*** handle all remaining command-line options ***/
+
+    if (strcmp(argv[i], "arkode.interpolant_type") == 0)
+    {
+      i++;
+      retval = 1;
```
Steven-Roberts, line 284:
```suggestion
      retval = ARK_ILL_INPUT;
```

src/arkode/arkode_io.c
```diff
@@ -34,6 +35,347 @@
...
+      else if (strcmp(argv[i], "ARK_INTERP_NONE") == 0)
+      {
+        retval = ARKodeSetInterpolantType(arkode_mem, ARK_INTERP_NONE);
+      }
+      if (retval != ARK_SUCCESS)
+      {
+        arkProcessError(ark_mem, ARK_ILL_INPUT, __LINE__, __func__, __FILE__,
+                        "error setting command-line argument: %s %s",
+                        argv[i-1], argv[i]);
+        return ARK_ILL_INPUT;
```
Steven-Roberts, line 302: It seems this function always returns `ARK_ILL_INPUT` on a failure. If the input was ok, but a function like `ARKodeSetInterpolantType` failed for a different reason, e.g., some internal interpolation memory was corrupted, it would be more informative to pass along that error.

Steven-Roberts (commented):
I like how clean this is! I'm ok with the public API aside from the handling of boolean settings.

src/arkode/arkode_io.c
Steven-Roberts: Possibly out of scope for this initial PR, but in order to support full configuration over an MRI or operator splitting method where they may be several arkode instances, we may need to support user-specified prefixes. For example,  `arkode.order 2 partition1.order 2 partition2.linear 1`

src/arkode/arkode_io.c
Steven-Roberts: Nevermind, I overlooked that `ARKodeSetAutonomous` is already covered in `arkIntSetFn`

src/arkode/arkode_io.c
gardner48: I agree using the hashmap would be a good way to make sure these lists to not get out of sync somehow.

src/arkode/arkode_io.c
gardner48: I think it would be clearer to refer to these a "aciton" functions rather than boolean functions as their only input is the integrator memory structure e.g., `ARKodeSetNonlinear` and `ARKodeClearStopTime`. The functions that take a boolean input are group with the "int" functions.

src/arkode/arkode_io.c
gardner48: I think we could do this by requiring a certain format for the keys e.g., `user_label.arkode.option` or `arkode.user_label.option`. Then `ARKodeSetFromCommandLine` could accept an additional input for which `user_label` to filter for and if none is provided, only read `arkode.` keys.

balos1: We have a rough sketch of scheme for MRI and operator splitting in the design document.

src/arkode/arkode_io.c
```diff
@@ -34,6 +35,347 @@
...
+      }
+      arg_used = SUNTRUE;
+      continue;
+    }
+
+    if (strcmp(argv[i], "arkode.tolerances") == 0)
```
gardner48, line 308: Maybe this should be `arkode.scalar_tolerances` so we could have `arkode.vector_tolerances`. Setting the vector tolerance would be more complicated though e.g., if using a distributed vector, GPU vector, user-defined vector, etc. Do we plan on supporting vector-valued inputs from the command line?

balos1: We do not have to support vector-valued inputs in the first pass, but yes, we plan to support vector-valued. The design doc covers non-distributed vectors. For distributed vectors, we will have to think about it more. Initially, I would be inclined to take the first `local_size` vector-value inputs and copy them to all ranks.

balos1: It would be nice to have separate arguments for absolute and relative tolerances. However, since the tolerance 'set' functions take both as inputs it would be a little complicated to implement. Maybe we could call the tolerance function twice?

src/arkode/arkode_io.c
gardner48: Ah, I remember this now. Looks like we settled on a `pkg.scope.option` style.

If this functions is going to be at the  ARKODE level, I think it will need an additional input to filter with since this function won't know which MRI level or splitting partition the memory corresponds to. 

Will we also need stepper specific versions of this function to handle options like setting the method to use or should we have only stepper-specific versions that call this function internally?

src/arkode/arkode_io.c
Steven-Roberts: Unlike the current arrays, a hashmap would not be stored fully as a static global variable since it uses malloc. We could declare it as a static local variable that only gets initialized once, but then I worry about thread safety.

Now I'm thinking arrays might be safer. If we want to keep fast lookup times, we can ensure they are sorted and use `bsearch` rather than manually iterating over them.

gardner48: Good point, maybe using an array of structs would be a better way to go to keep the `name` and `fun` together but not need to `malloc` the array.

src/arkode/arkode_io.c
gardner48: > I dont think we need to discuss the distributed aspect here though, we should do it in the design document.

Agreed, we'll need to work through this for GPU and ManyVectors also. I think we should be able to leverage the buffer pack and unpack functions to make these work.

> Maybe we could call the tolerance set function twice?

I think this would be a little tricky since we need to save the values for the later calls

src/arkode/arkode_io.c
drreynolds: I like the plan for an array of structs.  I haven't seen bsearch before, so I'll need to read up a bit on that instead of just iterating through the list.

drreynolds:
I think I've finished revising the previous implementation according to the initial feedback.  It would be great if folks who looked at this previously would give it another look.  Once you're happy with the structure, I'll get started on the ARKODE stepper-specific functions, other integrators, linear solvers, nonlinear solvers, etc.

src/arkode/arkode_io.c
```diff
@@ -34,6 +35,360 @@
   ARKODE optional input functions
   ===============================================================*/
 
+/*---------------------------------------------------------------
```
Steven-Roberts, line 38: [Optional suggestion]: Move the new content in `arkode_io.c` to a separate file like `arkode_cli.c` since this file is already close to 4k lines

src/arkode/arkode_io.c
```diff
@@ -34,6 +35,360 @@
...
+     {"predictor_method", ARKodeSetPredictorMethod},
+     {"max_nonlin_iters", ARKodeSetMaxNonlinIters},
+     {"max_hnil_warns", ARKodeSetMaxHnilWarns},
+     {"interpolate_stop_time", ARKodeSetInterpolateStopTime},
+     {"max_num_constr_fails", ARKodeSetMaxNumConstrFails},
+     {"adaptivity_adjustment", ARKodeSetAdaptivityAdjustment},
+     {"small_num_efails", ARKodeSetSmallNumEFails},
+     {"max_err_test_fails", ARKodeSetMaxErrTestFails},
+     {"max_conv_fails", ARKodeSetMaxConvFails}};
+  int num_int_keys = 15;
```
Steven-Roberts, line 191:
```suggestion
  static const int num_int_keys = sizeof(int_pairs) / sizeof(*int_pairs);
```

src/arkode/arkode_io.c
```diff
@@ -34,6 +35,360 @@
...
+     {"step_direction", ARKodeSetStepDirection},
+     {"cfl_fraction", ARKodeSetCFLFraction},
+     {"safety_factor", ARKodeSetSafetyFactor},
+     {"error_bias", ARKodeSetErrorBias},
+     {"max_growth", ARKodeSetMaxGrowth},
+     {"min_reduction", ARKodeSetMinReduction},
+     {"max_first_growth", ARKodeSetMaxFirstGrowth},
+     {"max_efail_growth", ARKodeSetMaxEFailGrowth},
+     {"max_cfail_growth", ARKodeSetMaxCFailGrowth}};
+  int num_real_keys = 18;
```
Steven-Roberts, line 216:
```suggestion
  static const int num_real_keys = sizeof(real_pairs) / sizeof(*real_pairs);
```

src/arkode/arkode_io.c
```diff
@@ -34,6 +35,360 @@
...
+
+  static struct arkKeyActionPair action_pairs[] =
+    {{"nonlinear", ARKodeSetNonlinear},
+     {"clear_stop_time", ARKodeClearStopTime},
+     {"no_inactive_root_warn", ARKodeSetNoInactiveRootWarn},
+     {"reset_accumulated_error", ARKodeResetAccumulatedError}};
+  int num_action_keys = 4;
```
Steven-Roberts, line 223:
```suggestion
  static const int num_action_keys = sizeof(action_pairs) / sizeof(*action_pairs);
```

src/arkode/arkode_io.c
```diff
@@ -34,6 +35,360 @@
...
+     {"max_num_constr_fails", ARKodeSetMaxNumConstrFails},
+     {"adaptivity_adjustment", ARKodeSetAdaptivityAdjustment},
+     {"small_num_efails", ARKodeSetSmallNumEFails},
+     {"max_err_test_fails", ARKodeSetMaxErrTestFails},
+     {"max_conv_fails", ARKodeSetMaxConvFails}};
+  int num_int_keys = 15;
+
+  static struct arkKeyLongPair long_pairs[] = {
+    {"max_num_steps", ARKodeSetMaxNumSteps}};
+  int num_long_keys = 1;
```
Steven-Roberts, line 195:
```suggestion
  static const int num_long_keys = sizeof(long_pairs) / sizeof(*long_pairs);
```

src/arkode/arkode_io.c
```diff
@@ -34,6 +35,360 @@
...
+     {"reset_accumulated_error", ARKodeResetAccumulatedError}};
+  int num_action_keys = 4;
+
+  int i, j, retval;
+  for (i = 1; i < argc; i++)
+  {
+    sunbooleantype arg_used = SUNFALSE;
+
+    /* skip command-line arguments that do not begin with "arkode." */
+    static const char* prefix = "arkode.";
```
Steven-Roberts, line 231: Just repeating my message on Slack for posterity

What about this format: `<arkid>.<option>` where `arkid` defaults to `arkode.` if `NULL`. I think this has two benefits:
- Users can better control nesting and prefix of option
- If a user wants to swap integrators without rewriting each option, they can use an empty or consistent prefix among the command line parsing functions.

include/sundials/priv/sundials_cli.h
```diff
@@ -0,0 +1,148 @@
+/* -----------------------------------------------------------------
```
Steven-Roberts, line 1: Can this be moved to `src/sundials/sundials_cli.h` to limit scope? I don't think any of the declarations are used in `include`.

include/arkode/arkode.h
```diff
@@ -214,6 +214,10 @@ typedef enum
  * Shared API routines
  * -------------------------- */
 
+/* Command-line control over ARKODE options */
+SUNDIALS_EXPORT int ARKodeSetFromCommandLine(void* arkode_mem, const char* arkid,
+                                             int argc, char* argv[]);
```
Steven-Roberts, line 219:
```suggestion
SUNDIALS_EXPORT int ARKodeSetFromCommandLine(void* arkode_mem, const char* arkid,
                                             int argc, const char* argv[]);
```
I'm not sure if I suggested this already... Functions with this arguments will also need to be updated.

src/arkode/arkode_cli.c
```diff
@@ -0,0 +1,312 @@
...
+        arkProcessError(ark_mem, retval, __LINE__, __func__, __FILE__,
+                        "error setting command-line argument: %s %s",
+                        argv[i - 1], argv[i]);
+        return retval;
+      }
+      arg_used = SUNTRUE;
+      continue;
+    }
+
+    if (strcmp(argv[i] + offset, "writeparameters") == 0)
```
Steven-Roberts, line 270: Is the convention snake case? Either way, I like that this option is included.
```suggestion
    if (strcmp(argv[i] + offset, "write_parameters") == 0)
```

include/arkode/arkode.h
drreynolds: You did suggest this previously, but I could not get it to work.  If I recall the error messages correctly, the problem is that this it is essentially a `char**` so the compiler cannot ensure that functions will not interfere with its internal memory.  I ended up having to propagate the `const` all the way to the `main` routine for it to compile, and I believe that is overly restrictive for user code (who may want to remove command-line options after they are parsed).

include/sundials/priv/sundials_cli.h
gardner48: `include/sundials/priv` is installed because some header files include function implementations so the "private" error handling functions and macros need to be installed so they're accessible in these functions. If these implementations were made internal then `include/sundials/priv` could be removed.

src/sunadaptcontroller/imexgus/sunadaptcontroller_imexgus.c
```diff
@@ -94,6 +107,92 @@ SUNAdaptController SUNAdaptController_ImExGus(SUNContext sunctx)
...
+      if (retval != SUN_SUCCESS) { return retval; }
+      continue;
+    }
+
+    /* control over SetErrorBias function */
+    if (strcmp(argv[i] + offset, "error_bias") == 0)
+    {
+      i += 1;
+      sunrealtype rarg = atof(argv[i]);
+      retval           = SUNAdaptController_SetErrorBias_ImExGus(C, rarg);
```
Steven-Roberts, line 170: Since bias can be set to a generic `SUNAdaptController` should we parse the "error_bias" argument in the superclass rather than duplicating it in each subclass?

src/sunadaptcontroller/imexgus/sunadaptcontroller_imexgus.c
```diff
@@ -94,6 +107,92 @@ SUNAdaptController SUNAdaptController_ImExGus(SUNContext sunctx)
...
+    }
+  }
+
+  /* Call SUNAdaptController_Write_ImExGus (if requested) now that all
+     command-line options have been set -- WARNING: this knows
+     nothing about MPI, so it could be redundantly written by all
+     processes if requested. */
+  if (write_parameters)
+  {
+    retval = SUNAdaptController_Write_ImExGus(C, stdout);
```
Steven-Roberts, line 189: Same issue about moving this parsing up to superclass

src/sunadaptcontroller/soderlind/sunadaptcontroller_soderlind.c
```diff
@@ -103,6 +117,187 @@ SUNAdaptController SUNAdaptController_Soderlind(SUNContext sunctx)
...
+    size_t SOffset, PIDOffset, PIOffset, IOffset, ExpGusOffset, ImpGusOffset;
+    const char* SPrefix   = (strlen(Cid) > 0) ? Cid
+                                              : "sunadaptcontroller_soderlind.";
+    const char* PIDPrefix = (strlen(Cid) > 0) ? Cid : "sunadaptcontroller_pid.";
+    const char* PIPrefix  = (strlen(Cid) > 0) ? Cid : "sunadaptcontroller_pi.";
+    const char* IPrefix   = (strlen(Cid) > 0) ? Cid : "sunadaptcontroller_i.";
+    const char* ExpGusPrefix = (strlen(Cid) > 0) ? Cid
+                                                 : "sunadaptcontroller_expgus.";
+    const char* ImpGusPrefix = (strlen(Cid) > 0) ? Cid
+                                                 : "sunadaptcontroller_impgus.";
```
Steven-Roberts, line 146: I think a couple more controllers were added in 7.3 so this would need to be updated. The bigger issues are this function is difficult to maintain and it breaks what I think is a convention of associating one prefix per class. I'm struggling to think of a good solution, but moving bias, defaults, and write parameter parsing up to superclass will help.

Steven-Roberts (changes requested):
Not necessary for this PR, but some sort of help functionality would be very helpful for identifying available options.

src/sunadaptcontroller/imexgus/sunadaptcontroller_imexgus.c
```diff
@@ -94,6 +107,92 @@ SUNAdaptController SUNAdaptController_ImExGus(SUNContext sunctx)
...
+    if (strcmp(argv[i] + offset, "params") == 0)
+    {
+      i += 1;
+      sunrealtype rarg1 = atof(argv[i]);
+      i += 1;
+      sunrealtype rarg2 = atof(argv[i]);
+      i += 1;
+      sunrealtype rarg3 = atof(argv[i]);
+      i += 1;
+      sunrealtype rarg4 = atof(argv[i]);
```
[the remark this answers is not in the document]
gardner48: The function is `sunrealtype SUNStrToReal(const char* str)` in `sundials_math.h|c`

src/sundials/sundials_cli.c
```diff
@@ -0,0 +1,211 @@
...
+                             const size_t offset, const char* argtest,
+                             sunIntRealSetFn fname, sunbooleantype* arg_used)
+{
+  *arg_used = SUNFALSE;
+  if (strcmp(argv[*i] + offset, argtest) == 0)
+  {
+    (*i) += 1;
+    int iarg = atoi(argv[*i]);
+    (*i) += 1;
+    sunrealtype rarg = atof(argv[*i]);
```
balos1, line 88:
```suggestion
    sunrealtype rarg = SUNStrToReal(argv[*i]);
```

src/sundials/sundials_cli.c
```diff
@@ -0,0 +1,211 @@
...
+                                 sunIntRealRealSetFn fname,
+                                 sunbooleantype* arg_used)
+{
+  *arg_used = SUNFALSE;
+  if (strcmp(argv[*i] + offset, argtest) == 0)
+  {
+    (*i) += 1;
+    int iarg = atoi(argv[*i]);
+    (*i) += 1;
+    sunrealtype rarg1 = atof(argv[*i]);
```
balos1, line 107:
```suggestion
    sunrealtype rarg1 = SUNStrToReal(argv[*i]);
```

src/sundials/sundials_cli.c
```diff
@@ -0,0 +1,211 @@
...
+    (*i) += 1;
+    sunrealtype rarg2 = atof(argv[*i]);
```
balos1, line 109:
```suggestion
    sunrealtype rarg2 = SUNStrToReal(argv[*i]);
```

src/sundials/sundials_cli.c
```diff
@@ -0,0 +1,211 @@
...
+
+int sunCheckAndSetRealArg(void* mem, int* i, char* argv[], const size_t offset,
+                          const char* argtest, sunRealSetFn fname,
+                          sunbooleantype* arg_used)
+{
+  *arg_used = SUNFALSE;
+  if (strcmp(argv[*i] + offset, argtest) == 0)
+  {
+    (*i) += 1;
+    sunrealtype rarg = atof(argv[*i]);
```
balos1, line 143:
```suggestion
    sunrealtype rarg = SUNStrToReal(argv[*i]);
```

src/sundials/sundials_cli.c
```diff
@@ -0,0 +1,211 @@
...
+
+int sunCheckAndSetTwoRealArg(void* mem, int* i, char* argv[],
+                             const size_t offset, const char* argtest,
+                             sunTwoRealSetFn fname, sunbooleantype* arg_used)
+{
+  *arg_used = SUNFALSE;
+  if (strcmp(argv[*i] + offset, argtest) == 0)
+  {
+    (*i) += 1;
+    sunrealtype rarg1 = atof(argv[*i]);
```
balos1, line 159:
```suggestion
    sunrealtype rarg1 = SUNStrToReal(argv[*i]);
```

src/sundials/sundials_cli.c
```diff
@@ -0,0 +1,211 @@
...
+    (*i) += 1;
+    sunrealtype rarg2 = atof(argv[*i]);
```
balos1, line 161:
```suggestion
    sunrealtype rarg2 = SUNStrToReal(argv[*i]);
```

balos1 (changes requested):
The comments I have left are pretty minor. 

My two biggest concerns are 1) that developers will have to remember that the *_cli.c files need to be updated when a set function is added, 2) its hard to test that everything is wired up correctly. 

Unfortunately, I don't have a great way to address either of these concerns, but one thing that may help is a developer documentation section on "Setting SUNDIALS Options through the Command Line" where we somewhat briefly describe how this all works. Then, we can also add a note to the [developer checklist](https://sundials.readthedocs.io/en/latest/developers/getting_started/Checklist.html) about having to update the CLI files when a set function is added.

src/arkode/arkode_arkstep_io.c
```diff
@@ -628,6 +629,63 @@ int ARKStepGetTimestepperStats(void* arkode_mem, long int* expsteps,
...
+                                      twochar_pairs[j].set, arg_used);
+    if (retval != ARK_SUCCESS)
+    {
+      arkProcessError(ark_mem, retval, __LINE__, __func__, __FILE__,
+                      "error setting command-line argument: %s",
+                      twochar_pairs[j].key);
+      return retval;
+    }
+    if (*arg_used) { return ARK_SUCCESS; }
+  }
```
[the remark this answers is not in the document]
drreynolds: Given that each integrator still has its own mechanism for error handling, I did not create a SUNDIALS-wide helper for this.  I'll consider whether a SUNDIALS-wide helper function could just use SUNDIALS-based error handling, and then each intgrator could just test whether the overall helper function failed before calling the integrator-specific error hander.

src/sunlinsol/klu/sunlinsol_klu.c
```diff
@@ -168,6 +181,47 @@ SUNErrCode SUNLinSol_KLUReInit(SUNLinearSolver S, SUNMatrix A, sunindextype nnz,
...
+    }
+    else
+    {
+      static const char* prefix = "klu.";
+      if (strncmp(argv[i], prefix, strlen(prefix)) != 0) { continue; }
+      offset = strlen(prefix);
+    }
+
+    /* control over SetOrdering function */
+    if (strcmp(argv[i] + offset, "ordering") == 0)
```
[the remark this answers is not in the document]
drreynolds: I don't see how that would be possible.

The issue is that `sunCheckAndSetIntArg` expects that the "Set" routine provided must have type `int (*sunIntSetFn)(void*, int)`.  However, the first argument to each linear solver module set routine has type `SUNLinearSolver`.  This same pattern is present for our other object-oriented modules as well.  It would be a huge API change to convert all our object-oriented module set routines to require `void*` input (and then cast the argument to the appropriate type), just so that the command-line processing function could call the shared routine.

src/sunadaptcontroller/imexgus/sunadaptcontroller_imexgus.c
drreynolds: The reason I put it here is so that if a user wants to set both `error_bias` and `params`, those would use the same prefix.  If `error_bias` were moved to the superclass, then it seems to me that the user would need to differentiate the prefixes even though they're using the same controller, e.g.,
```
  a.out sunadaptcontroller.error_bias 1.2 sunadaptcontroller_imexgus.params 1.0 1.0 1.0 1.0
```

I'd rather have a small amount of duplicated code between the three adaptivity controllers than cause that level of confusion for users.

src/sunadaptcontroller/imexgus/sunadaptcontroller_imexgus.c
drreynolds: Same response as for `error_bias`

src/sunlinsol/klu/sunlinsol_klu.c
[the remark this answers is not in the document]
Steven-Roberts: Ah, I didn't realize the first argument type wasn't `void*`.

src/sunadaptcontroller/soderlind/sunadaptcontroller_soderlind.c
drreynolds: I disagree.  The new controllers that were added in 7.3 have fixed coefficients, and thus they do not even provide `SetParams` routines in the first place.  This piece had to be complicated to support our older controller API, where the controller "type" merely indicated the specific structure of the controller (one vs two vs three (h,dsm) history values, etc), but all exponents were user modifiable.

Since the new controllers have no `SetParams` routines, then their other parameters (e.g., bias) can already be specified via the `Cid` value `sunadaptcontroller_soderlind`.  I added this note to the documentation for each of those utility constructors.

drreynolds:
I believe I have addressed all PR comments. Please re-review as needed.

src/sunadaptcontroller/imexgus/sunadaptcontroller_imexgus.c
gardner48: Continuing the discussion in [this](https://github.com/LLNL/sundials/pull/662#discussion_r2072196366) thread

src/sunlinsol/klu/sunlinsol_klu.c
```diff
@@ -168,6 +181,47 @@ SUNErrCode SUNLinSol_KLUReInit(SUNLinearSolver S, SUNMatrix A, sunindextype nnz,
...
+  SUNErrCode retval;
+  for (idx = 1; idx < argc; idx++)
+  {
+    /* if LSid is supplied, skip command-line arguments that do not begin with LSid;
+       else, skip command-line arguments that do not begin with "spbcgs." */
+    size_t offset;
+    if (LSid != NULL)
+    {
+      if (strncmp(argv[idx], LSid, strlen(LSid)) != 0) { continue; }
+      offset = strlen(LSid) + 1;
```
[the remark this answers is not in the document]
gardner48: See related discussion [here](https://github.com/LLNL/sundials/pull/662#discussion_r2197782093) about supporting an empty string as the prefix.

src/arkode/arkode_sprkstep_io.c
```diff
@@ -194,6 +195,48 @@ int SPRKStepGetNumRhsEvals(void* arkode_mem, long int* nf1, long int* nf2)
...
+int sprkStep_SetFromCommandLine(ARKodeMem ark_mem, int* argidx, char* argv[],
+                                size_t offset, sunbooleantype* arg_used)
+{
+  /* Set lists of command-line arguments, and the corresponding set routines */
+  static const struct sunKeyCharPair char_pairs[] = {
+    {"method_name", SPRKStepSetMethodName}};
+  static const int num_char_keys = sizeof(char_pairs) / sizeof(*char_pairs);
+
+  static const struct sunKeyIntPair int_pairs[] = {
+    {"use_compensated_sums", SPRKStepSetUseCompensatedSums}};
```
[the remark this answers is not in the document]
gardner48: I don't think I noted this elsewhere but `ARKodeSetUseCompensatedSums` need to be added to `arkode_cli.c`

src/sunadaptcontroller/soderlind/sunadaptcontroller_soderlind.c
```diff
@@ -103,6 +117,184 @@ SUNAdaptController SUNAdaptController_Soderlind(SUNContext sunctx)
...
+  SUNFunctionBegin(C->sunctx);
+
+  int idx;
+  SUNErrCode retval;
+  sunbooleantype write_parameters = SUNFALSE;
+  for (idx = 1; idx < argc; idx++)
+  {
+    /* if Cid is supplied, skip command-line arguments that do not begin with Cid;
+       else, skip command-line arguments that do not begin with "sunadaptcontroller." */
+    size_t SOffset, PIDOffset, PIOffset, IOffset, ExpGusOffset, ImpGusOffset;
```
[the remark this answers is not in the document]
gardner48: See related discussion in the conversation thread [here](https://github.com/LLNL/sundials/pull/662#discussion_r2072196366)

gardner48:
@drreynolds I jumped around the files a little in reviewing, so you'll probably want to read through the less trivial comments before diving into updates.

doc/arkode/guide/source/Usage/User_callable.rst
```diff
@@ -906,6 +912,39 @@ Set the checkpointing step index (for adjoint)     :c:func:`ARKodeSetAdjointChec
 
 
 
+.. c:function:: int ARKodeSetFromCommandLine(void* arkode_mem, const char* arkid, int argc, char* argv[])
```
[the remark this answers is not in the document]
drreynolds: Yes, the goal is to eventually support input files.  I'll add the extra layer and update the current routines, but this will take some time since there are now many of these interspersed throughout the repository.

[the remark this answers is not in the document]
drreynolds: In thinking about this further, we'd need to give precedence to either command-line or file input options with this approach.  Perhaps if `file_name` is non-NULL then those are processed first, followed by command-line options?  For now we would leave the non-NULL `file_name` conditional as a "To-Do" item, and note in the documentation that it is not yet implemented.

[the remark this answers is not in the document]
gardner48: Yes, that's what I'm thinking and `SetOptions` could just be a wrapper function something like
```
int ARKodeSetOptions(void* arkode_mem, const char* arkid, const char* file_name, int argc, char* argv[])
{
  if (file_name != NULL && strlen(empty_string) != 0)
  {
    int retval = arkSetOptionsFromFile(arkode_mem, arkid, file_name);
    if (retval != ARK_SUCCESS) { return retval; }
  }

  if (argc > 0 && argv != NULL)
  {
    int retval = arkSetOptionsFromCommandLine(arkode_mem, arkid, argc, argv);
    if (retval != ARK_SUCCESS) { return retval; }
  }

  return ARK _SUCCESS;
}
```
For now, we would not call `arkSetOptionsFromFile` and instead return an error that this is not supported yet.

doc/shared/sunadaptcontroller/SUNAdaptController_Description.rst
```diff
@@ -261,13 +267,47 @@ note these requirements below. Additionally, we note the behavior of the base SU
...
 
    Sets the controller parameters to their default values.
 
    :param C:  the :c:type:`SUNAdaptController` object.
    :return: :c:type:`SUNErrCode` indicating success or failure.
 
+   .. note::
+
+      This routine will be called by :c:func:`SUNAdaptController_SetFromCommandLine`
+      when using the command-line option "Cid.defaults".
```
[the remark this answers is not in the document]
gardner48: I think I made this comment before getting into the source. I see now that the inputs are processed in the order they are provided (we should note this in the user guide if we don't already) so the user will just need to consider any option order dependencies when providing the inputs. 

Usually we don't require a particular order, so maybe this is generally not an issue, but it's something we'll need to watch out for when adding new set functions and, to make things easier for the user, we should try to avoid adding an order dependence when possible.

examples/arkode/CXX_serial/ark_heat2D_lsrk.cpp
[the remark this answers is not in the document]
drreynolds: No, I believe that I'd tested some options with this file (which required me to remove the extraneous error check whether too-many options were supplied), but once it passed I removed the extra code.

doc/arkode/guide/source/Usage/User_callable.rst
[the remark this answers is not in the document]
drreynolds: What is the role of the `&& strlen(empty_string) != 0` here?

src/sunadaptcontroller/imexgus/sunadaptcontroller_imexgus.c
drreynolds: I'm updating everything (SUNAdaptControllers, SUNLinearSolvers, SUNNonlinearSolvers) so that each shares a common prefix (e.g., "sunadaptcontroller" or "sunlinearsolver").  However, the new "feature" wherein we will support either command-line or file-based arguments using the same class function necessitates that the actual processing be handled in each implementation and not directly in the base class.  While it will be somewhat redundant, it will certainly clean up the API and allow us to be more "forward thinking" with respect to file-based inputs.

drreynolds: I am concerned with how the branching (file-based input vs command-line input) would occur at two separate levels, given the single `SetOptions` points of entry for both the base class and the implementations.  If we want the base class to handle some keys before calling the implementations, then it would need to first branch between file-based vs command-line input, call its private function for file-based input, call its private function for command-line input, and then call the implementation's `SetOptions` function pointer to repeat the process.  Since both the base class and implementations will share the same prefix (here, `sunadaptcontroller`), then I am concerned that doing (a) base class file-based, then (b) base class command-line, then (c) implementation file-based, then (d) implementation command-line, could open things up for odd behavior if the keys were not sufficiently distinct.

drreynolds:
I believe I've addressed all PR comments *except* for moving command-line processing to base classes from implementations.  As I noted in my responses, I have concerns with making that move, since I worry that it decreases robustness for the sake of saving a few lines of code from implementations (but adding around the same number of lines elsewhere). 

If others feel differently then I recommend we debate that later in a subsequent PR.  I see no reason such a minor implementation detail should hold up this release, since even if we decide to eventually re-engineer the internals to make the move, it will not change the functionality or API.

I'm still updating answer files due to changes in what is printed in some examples, but otherwise this PR should be good for re-review.

PR #673, opened 2025-03-17
spack v1.0 support: `%foo +bar` -> `+bar %foo`

haampie:
Spack v1.0 will parse `pkg %gcc +foo` as "pkg with direct dependency gcc
with variant foo enabled" instead of "pkg with variant foo enabled and compiler
gcc". Reorder as `pkg +foo %gcc` to be compatible with Spack v0.x and
v1.0.

- [ ] Please target the `develop` branch not `main`.
- [ ] Review our [Contributing Guide](https://github.com/LLNL/sundials/blob/main/CONTRIBUTING.md), and ensure that you sign your last commit (at minimum) as per the guide.
- [ ] Provide a concise description of what your pull request does, and why it is needed/benefical.
- [ ] Add a note about your change to the `CHANGELOG.md` and `docs/shared/RecentChanges.rst` files. Notice that the former is a markdown file and the latter is reStructuredText, so the formatting is slightly different.
- [ ] After your PR is opened, ensure that all of the tests are passing (a SUNDIALS developer will have to allow the testing to run).

### DIFF
--- a/docker/sundials-ci/e4s-quarterly/int32-double/spack.yaml
+++ b/docker/sundials-ci/e4s-quarterly/int32-double/spack.yaml
@@ -10,16 +10,16 @@ spack:
       compiler:
       - gcc@9.4.0
   specs:
-  - cmake %gcc@9.4.0 arch=x86_64
-  - hypre~int64~internal-superlu %gcc@9.4.0 arch=x86_64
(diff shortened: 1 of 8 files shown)


drreynolds (approved):
LGTM

PR #678, opened 2025-03-23
Feature: CVODE Resize History

gardner48:
Add support for resizing CVODE and CVODES when solving initial value problems where the number of equations and unknowns changes over time. Resizing requires a user supplied history of solution and right-hand side values at the new problem size.


test/unit_tests/cvode/CXX_serial/cv_test_resize_history.cpp
```diff
@@ -0,0 +1,423 @@
...
+  {
+    N_VScale(one, y_hist[i], y_hist[i + 1]);
+    N_VScale(one, f_hist[i], f_hist[i + 1]);
+  }
+  N_VScale(one, y_n, y_hist[0]);
+  int retval = ode_rhs(t_n, y_n, f_hist[0], user_data);
+  if (retval) { return 1; }
+
+  return 0;
+}
```
drreynolds, line 63: Would it be possible to provide a CVODE "Get" routine to retrieve a saved history from the current Nordsiek array (e.g., by calling `CVodeGetDky` appropriately), thereby streamlining things so that users aren't required to do this kind of thing themselves?

test/unit_tests/cvode/CXX_serial/cv_test_resize_history.cpp
```diff
@@ -0,0 +1,423 @@
...
+  for (int i = 0; i < hist_size; i++)
+  {
+    N_VDestroy(f_hist[i]);
+    f_hist[i]  = N_VClone(y_hist[i]);
+    int retval = ode_rhs(t_hist[i], y_hist[i], f_hist[i], user_data);
+    if (retval) { return 1; }
+  }
+
+  return 0;
+}
```
drreynolds, line 90: I can also imagine a CVODE utility routine that would handle something like this as well, by taking a user-supplied function to "project" a vector of the old size onto a vector of the new size, or perhaps two projection functions: one for `y` and the other for `f`?

src/cvodes/cvodes_resize.c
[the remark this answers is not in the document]
balos1: It is good to have all the features of CVODE for forward IVP integration also available in CVODES (and I believe this is our precedent). Many users utilize CVODES rather than CVODE for forward IVP integration so that they do not have to maintain an interface to both CVODE and CVODES.

test/unit_tests/cvode/CXX_serial/cv_test_resize_history.cpp
```diff
@@ -0,0 +1,423 @@
...
+    if (atoi(argv[2]) == 1) { method = CV_ADAMS; }
+    else if (atoi(argv[2]) == 2) { method = CV_BDF; }
+    else
+    {
+      cerr << "Invalid method option" << endl;
+      return 1;
+    }
+  }
+
+  // Use fixed-point (0) or Newton (1) nonlinear solver
```
Steven-Roberts, line 151: A bit inconsistent that the first arg is 1 or 2 while this is 0 or 1.

test/unit_tests/utilities/check_return.hpp
```diff
@@ -18,7 +18,7 @@
 #include <string>
 
 // Check function for non-zero return flag
-static int check_flag(int flag, const std::string funcname)
+inline int check_flag(int flag, const std::string funcname)
```
Steven-Roberts, line 21: While you're updating this
```suggestion
inline int check_flag(int flag, const std::string& funcname)
```

test/unit_tests/utilities/check_return.hpp
```diff
@@ -28,7 +28,7 @@ static int check_flag(int flag, const std::string funcname)
 }
 
 // Check if a function returned a NULL pointer
-static int check_ptr(void* ptr, const std::string funcname)
+inline int check_ptr(void* ptr, const std::string funcname)
```
Steven-Roberts, line 31:
```suggestion
inline int check_ptr(void* ptr, const std::string& funcname)
```

test/unit_tests/utilities/test_utilities.hpp
```diff
@@ -21,7 +21,7 @@
 #include <vector>
 
 // Check function return flag
-static int check_flag(const int flag, const std::string funcname)
+inline int check_flag(const int flag, const std::string funcname)
```
Steven-Roberts, line 24:
```suggestion
inline int check_flag(const int flag, const std::string& funcname)
```

test/unit_tests/utilities/test_utilities.hpp
```diff
@@ -30,7 +30,7 @@ static int check_flag(const int flag, const std::string funcname)
 }
 
 // Check if a function returned a NULL pointer
-static int check_ptr(const void* ptr, const std::string funcname)
+inline int check_ptr(const void* ptr, const std::string funcname)
```
Steven-Roberts, line 33:
```suggestion
inline int check_ptr(const void* ptr, const std::string& funcname)
```

doc/cvode/guide/source/Usage/index.rst
```diff
@@ -3274,6 +3274,83 @@ vector.
...
+      Any nonlinear or linear solvers attached to CVODE will also need to be
+      resized. At present, for SUNDIALS-provided algebraic solvers, this
+      requires be destroying, re-creating, and re-attaching the solvers
+      following each call to :c:func:`CVodeResizeHistory`. Similarly, any matrix
+      objects provided when attaching the linear solver will also need to be
+      resized.
+
+      If using a vector of absolute tolerances, the absolute tolerance vector
+      will be invalid after the call to :c:func:`CVodeResizeHistory`, so a new
+      absolute tolerance vector should be create and set following each call to
```
gardner48, line 3346:
```suggestion
      absolute tolerance vector should be created and set following each call to
```

doc/cvodes/guide/source/Usage/SIM.rst
```diff
@@ -3266,6 +3266,91 @@ vector.
...
+      Any nonlinear or linear solvers attached to CVODE will also need to be
+      resized. At present, for SUNDIALS-provided algebraic solvers, this
+      requires be destroying, re-creating, and re-attaching the solvers
+      following each call to :c:func:`CVodeResizeHistory`. Similarly, any matrix
+      objects provided when attaching the linear solver will also need to be
+      resized.
+
+      If using a vector of absolute tolerances, the absolute tolerance vector
+      will be invalid after the call to :c:func:`CVodeResizeHistory`, so a new
+      absolute tolerance vector should be create and set following each call to
```
gardner48, line 3344:
```suggestion
      absolute tolerance vector should be created and set following each call to
```

src/cvode/cvode.c
```diff
@@ -2456,7 +2502,11 @@ static void cvAdjustParams(CVodeMem cv_mem)
 {
   if (cv_mem->cv_qprime != cv_mem->cv_q)
   {
-    cvAdjustOrder(cv_mem, cv_mem->cv_qprime - cv_mem->cv_q);
+    /* History adjustments for an order change were applied when resizing */
+    if (!(cv_mem->first_step_after_resize))
+    {
+      cvAdjustOrder(cv_mem, cv_mem->cv_qprime - cv_mem->cv_q);
+    }
     cv_mem->cv_q     = cv_mem->cv_qprime;
```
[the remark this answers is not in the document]
gardner48: If the order does not change these lines are not called (`if (cv_mem->cv_qprime != cv_mem->cv_q)` above). 

If the order is changing the resizing function updated the Nordsieck array so the `cvAdjustOrder` call is not needed but the values of `q`, `L`, and `qwait` still need to be updated below.

src/cvode/cvode_resize.c
```diff
@@ -0,0 +1,509 @@
...
+/* -----------------------------------------------------------------------------
+ * Build Adams Nordsieck array from f(t,y) history and y(t) value
+ * ---------------------------------------------------------------------------*/
+
+static int cvBuildNordsieckArrayAdams(sunrealtype* t, N_Vector y, N_Vector* f,
+                                      N_Vector* wrk, int order,
+                                      sunrealtype hscale, N_Vector* zn)
+{
+  /* Check for valid inputs */
+  if (!t || !y || !f || !wrk || order < 1 || !zn) { return CV_ILL_INPUT; }
```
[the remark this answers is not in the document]
gardner48: When we extend the usage of the function to other contexts I think these checks will be needed so we should keep them.

src/cvode/cvode_resize.c
```diff
@@ -0,0 +1,509 @@
...
+/* -----------------------------------------------------------------------------
+ * Build BDF Nordsieck array from y(t) history and f(t,y) value
+ * ---------------------------------------------------------------------------*/
+
+static int cvBuildNordsieckArrayBDF(sunrealtype* t, N_Vector* y, N_Vector f,
+                                    N_Vector* wrk, int order,
+                                    sunrealtype hscale, N_Vector* zn)
+{
+  /* Check for valid inputs */
+  if (!t || !y || !f || !wrk || order < 1 || !zn) { return CV_ILL_INPUT; }
```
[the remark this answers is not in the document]
gardner48: When we extend the usage of the function to other contexts I think these checks will be needed so we should keep them.

test/unit_tests/cvode/CXX_serial/cv_test_resize_history.cpp
gardner48: The first two are 1 and 2 to match the values of the macros they correspond to

doc/cvode/guide/source/Usage/index.rst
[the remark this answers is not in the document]
gardner48: The user is responsible for saving the history from CVODE returns which generally means they should call CVODE in one-step mode or build with monitoring enabled to access the state between steps. If insufficient history is supplied to resize for the order used in the next step an error is returned. I'll update the documentation to make this clear.

test/unit_tests/cvode/CXX_serial/cv_test_resize_history.cpp
gardner48: We could add an option to have CVODE save the needed state and RHS history as it goes but, for the current resize use case, this is not needed because the user must save the full solution history over time to compute history integrals however this might be useful in other cases.

Using `CVodeGetDky` with BDF methods may not work for retrieving state values further back than the most recently completed step because the corrector polynomial does not interpolate the state at the needed predictor times beyond `y_{n}` and `y_{n-1}`. So, older states would have some interpolation error (but maybe that's sufficiently small). For Adams methods, this might work except we need one history vector older than is interpolated by the corrector polynomial in order to compute the resized correction (this is also true for BDF methods).

test/unit_tests/cvode/CXX_serial/cv_test_resize_history.cpp
gardner48: In a case where CVODE saves the necessary history, yes having a function (or functions) like this would be nice way to fill the history vectors. For the current use case, saving the history in CVODE is unnecessary as the user code already has save the full history and this would require additional memory.

doc/cvode/guide/source/Usage/index.rst
```diff
@@ -3274,6 +3274,83 @@ vector.
...
+
+   For Adams methods the required history data is
+
+   * Solution vectors: :math:`y(t_n)` and :math:`y(t_{n-1})`
+   * Right-hand side vectors: :math:`f(t_n,y(t_n)), f(t_{n-1},y(t_{n-1})), \ldots, f(t_{n-k}, y(t_{n-k}))`
+
+   For BDF methods the required history data is:
+
+   * Solution vectors: :math:`y(t_n), y(t_{n-1}), \ldots, y(t_{n-k})`
+   * Right-hand side vectors: :math:`f(t_n, y(t_n))` and :math:`f(t_{n-1}, y(t_{n-1}))`,
```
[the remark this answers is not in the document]
gardner48: If you mean the user supplies only `y` values and we evaluate the right-hand side function to get the needed `f` values, then yes these requirement can be relaxed (with extra cost from the RHS evaluations). However, building the predictor Nordsieck array from the `y` values alone would not work as the underlying polynomial does not necessarily satisfy the same interpolation conditions on the derivative values. Similarly, an arbitrary set of `y` and `f` may not work depending on the method.

test/unit_tests/cvode/CXX_serial/cv_test_resize_history.cpp
drreynolds: Given that the problem is being resized anyways, and thus a user would likely need to interpolate in space already, then I'm not concerned with temporal interpolation that is only of the same order of accuracy as the method.  

That said, I would have some concerns about extrapolation further back in time.  However, it sounds like that extrapolated value would only actually be used for the predictor (not the corrector) on the first post-resize step, so perhaps the most "trouble" it could cause is for CVODE to reduce the step size and/or method order due to a failed post-resize-step.  With this in mind, I think that there's no need for a new "Get" routine, instead of possibly a paragraph in the user guide to mention using `CVodeGetDky` as a potential strategy (albeit with a warning).

gardner48: I'll have to look more into why `CVodeGetDky` does not allow interpolation outside of the most recent step and if there are any major issues there. 

The extrapolated value would be used to build a new predictor for the just completed step but at new problem size and from that get a new correction for the just completed step to potentially use in an order increase check in the next step. So this would only be a potential issue if the next step checks for an order increase.

doc/cvode/guide/source/Usage/index.rst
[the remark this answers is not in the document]
gardner48: Added an issue for this LLNL/sundials-private#109

test/unit_tests/cvode/CXX_serial/cv_test_resize_history.cpp
gardner48: Added in issue for this LLNL/sundials-private#109

doc/cvode/guide/source/Usage/index.rst
```diff
@@ -3279,10 +3279,16 @@ CVODE resize function
 
 For simulations involving changes to the number of equations and unknowns in the
 ODE system, CVODE maybe "resized" between steps by calling
```
drreynolds, line 3281:
```suggestion
ODE system, CVODE may be "resized" between steps by calling
```

doc/cvodes/guide/source/Usage/SIM.rst
```diff
@@ -3271,10 +3271,16 @@ CVODES resize function
 
 For simulations involving changes to the number of equations and unknowns in the
 ODE system, CVODES maybe "resized" between steps by calling
```
drreynolds, line 3273:
```suggestion
ODE system, CVODES may be "resized" between steps by calling
```

drreynolds (dismissed):
I noticed two tiny grammatical errors; otherwise this looks great.

drreynolds (approved):

PR #681, opened 2025-03-26
Feature: Anderson adaptive depth

gardner48:
Add support in KINSOL for setting a user-supplied function to compute the acceleration depth in fixed-point or Picard iterations.

### DIFF
--- a/doc/kinsol/guide/source/Mathematics.rst
+++ b/doc/kinsol/guide/source/Mathematics.rst
@@ -377,44 +377,50 @@ Anderson Acceleration
 The Picard and fixed point methods can be significantly accelerated using Anderson’s method
 :cite:p:`Anderson65, Walker-Ni09, Fang-Saad09, LWWY11`. Anderson acceleration can be formulated as follows:
 
-1. Set :math:`u_0 =` an initial guess and :math:`m \ge 1`
+1. Set :math:`u_0 =` an initial guess, :math:`m \ge 1`, and :math:`m_0 = 0`
 
 2. Set :math:`u_1 = G(u_0)`
 
 3. For :math:`n = 1, 2,...` until convergence do:
 
-   a.  Set :math:`m_n = \min\{m,n\}`
+   a.  Set :math:`m_n = \min\{m,n,m_{n-1} + 1\}`
 
    b.  Set :math:`F_{n} = (f_{n-m_n}, \ldots, f_n)`, where :math:`f_i=G(u_i)-u_i`
 
-   c.  Determine :math:`\alpha^{(n)} = (\alpha_0^{(n)}, \ldots, \alpha_{m_n}^{(n)})` that solves
+   c.  Optionally, update :math:`m_n` to remove some number of the left-most
+       columns from :math:`F_{n}`
+
+   d.  Determine :math:`\alpha^{(n)} = (\alpha_0^{(n)}, \ldots, \alpha_{m_n}^{(n)})` that solves
        :math:`\displaystyle\min_\alpha  \| F_n \alpha^T \|_2` such that :math:`\displaystyle\sum_{i=0}^{m_n} \alpha_i = 1`
 
-   d.  Set :math:`\displaystyle u_{n+1} = \beta_n \sum_{i=0}^{m_n} \alpha_i^{(n)} G(u_{n-m_n+i}) + (1-\beta_n) \sum_{i=0}^{m_n} \alpha_i^{(n)} u_{n-m_{n}+i}`
+   e.  Set :math:`\displaystyle u_{n+1} = \beta_n \sum_{i=0}^{m_n} \alpha_i^{(n)} G(u_{n-m_n+i}) + (1-\beta_n) \sum_{i=0}^{m_n} \alpha_i^{(n)} u_{n-m_{n}+i}`
 
-   e.  Test for convergence
+   f.  Test for convergence
 
 It has been implemented in KINSOL by turning the constrained linear least-squares problem in step 3c into an
 unconstrained one leading to the algorithm given below:
 
-1. Set :math:`u_0 =` an initial guess and :math:`m \ge 1`
+1. Set :math:`u_0 =` an initial guess, :math:`m \ge 1`, and :math:`m_0 = 0`
 
 2. Set :math:`u_1 = G(u_0)`
 
 3. For :math:`n = 1, 2,...` until convergence do:
 
-   a. Set :math:`m_n = \min\{m,n\}`
+   a. Set :math:`m_n = \min\{m,n,m_{n-1} + 1\}`
 
    b. Set :math:`\Delta F_{n} = (\Delta f_{n-m_n}, \ldots, \Delta f_{n-1})`, where :math:`\Delta f_i = f_{i+1} - f_i`
       and :math:`f_i=G(u_i)-u_i`
 
-   c. Determine :math:`\gamma^{(n)} = (\gamma_0^{(n)}, \ldots, \gamma_{m_n-1}^{(n)})` that solves
+   c. Optionally, update :math:`m_n` to remove some number of the left-most
+      columns from :math:`\Delta F_{n}`
+
+   d. Determine :math:`\gamma^{(n)} = (\gamma_0^{(n)}, \ldots, \gamma_{m_n-1}^{(n)})` that solves
       :math:`\displaystyle\min_\gamma  \| f_n - \Delta F_n \gamma^T \|_2`
 
-   d. Set :math:`\displaystyle u_{n+1} = G(u_n)-\sum_{i=0}^{m_n-1} \gamma_i^{(n)} \Delta g_{n-m_n+i} - (1-\beta_n)(f(u_n) - \sum_{i=0}^{m_n-1} \gamma_i^{(n)} \Delta f_{n-m_n+i})` with
+   e. Set :math:`\displaystyle u_{n+1} = G(u_n)-\sum_{i=0}^{m_n-1} \gamma_i^{(n)} \Delta g_{n-m_n+i} - (1-\beta_n)(f(u_n) - \sum_{i=0}^{m_n-1} \gamma_i^{(n)} \Delta f_{n-m_n+i})` with
       :math:`\Delta g_i = G(u_{i+1}) - G(u_i)`
 
-   e. Test for convergence
+   f. Test for convergence
 
 The least-squares problem in 3c is solved by applying a QR factorization to :math:`\Delta F_n = Q_n R_n` and solving
 :math:`R_n \gamma = Q_n^T f_n`. By default the damping is disabled i.e., :math:`\beta_n = 1.0`.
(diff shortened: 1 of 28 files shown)


doc/kinsol/guide/source/Usage/index.rst
```diff
@@ -1120,6 +1122,17 @@ negative, so a test ``retval`` :math:`<0` will catch any error.
       The function provided to :c:func:`KINSetDampingFn` will override any
       values set with :c:func:`KINSetDamping` or :c:func:`KINSetDampingAA`.
 
+.. c:function:: int KINSetDepthFn(void* kin_mem, KINDepthFn depth_fn)
+
+   Sets the function used to compute the update the depth, :math:`m_n`, in
+   fixed-point or Picard iterations.
```
drreynolds, line 1128:
```suggestion
   Sets the function used to compute the updated depth, :math:`m_n`, in
   fixed-point or Picard iterations.
```

doc/kinsol/guide/source/Usage/index.rst
```diff
@@ -2126,6 +2139,45 @@ of type :c:type:`KINDampingFn` to computing the damping factor, :math:`\beta_n`.
...
+Depth function
+~~~~~~~~~~~~~~
+
+When using the fixed-point or Picard iterations, the user may provide a function
+of type :c:type:`KINDepthFn` to modify the depth, :math:`m_n`.
+
+.. c:type:: int (*KINDepthFn)(long int iter, N_Vector u_val, N_Vector g_val, N_Vector f_val, N_Vector* df, sunrealtype* R_mat, long int depth, void* user_data, long int* new_depth, long int* remove_indices);
+
+   This function computes the depth parameter, :math:`m_{\textrm{max}} \leq m_n
+   \geq 0`, for fixed-point and Picard iterations.
```
drreynolds, line 2153: Are these inequalities correct?  

Also, the wording of this makes it difficult to determine if the function returns `m_n` or `m_{\textrm{max}}`.  In the text below, it looks like it actually returns `:math:`\hat{m}_n` -- I think that should be listed here for clarity.

gardner48: > Are these inequalities correct?

Nope, ends are reversed.

> Also, the wording of this makes it difficult to determine if the function returns `m_n` or `m_{\textrm{max}}`. In the text below, it looks like it actually returns `:math:`\hat{m}_n` -- I think that should be listed here for clarity.

Added some more detail to make this clearer.

drreynolds (approved):

doc/kinsol/guide/source/Usage/index.rst
```diff
@@ -2132,6 +2147,59 @@ from :numref:`KINSOL.Mathematics.FixedPoint` and
...
+     and elements are stored column-wise. On input, ``R_mat`` contains the
+     entries for an :math:`m_n \times m_n` matrix.
+   * **depth** -- the current size of the Anderson acceleration space,
+     :math:`m_n`.
+   * **user_data** -- the user data pointer passed to :c:func:`KINSetUserData`.
+   * **new_depth** -- the computed depth, :math:`\hat{m}_n`, to replace
+     the current depth, :math:`m_n`.
+   * **remove_indices** -- this parameter is currently unused (``NULL``) and
+     will be used to provide support for removing specific columns of
+     :math:`\Delta F_n` in the future.
```
Steven-Roberts, line 2195: I'm not familiar with how this is intended to be used in the future, but I'm surprised the type is `long int *` and not `sunbooleantype *` to indicate which to keep and remove.

doc/kinsol/guide/source/Mathematics.rst
```diff
@@ -381,44 +381,50 @@ Anderson Acceleration
...
 
 3. For :math:`n = 1, 2,...` until convergence do:
 
-   a.  Set :math:`m_n = \min\{m,n\}`
+   a.  Set :math:`m_n = \min\{m,n,m_{n-1} + 1\}`
 
    b.  Set :math:`F_{n} = (f_{n-m_n}, \ldots, f_n)`, where :math:`f_i=G(u_i)-u_i`
 
-   c.  Determine :math:`\alpha^{(n)} = (\alpha_0^{(n)}, \ldots, \alpha_{m_n}^{(n)})` that solves
+   c.  Optionally, update :math:`m_n` to remove some number of the left-most
```
Steven-Roberts, line 394: Not sure if this is what's actually done, but it seems wasteful to construct `F_{n}` only then to remove some columns. I.e., should b and c be switched?

examples/kinsol/serial/kinAnalytic_fp.c
```diff
@@ -234,6 +242,13 @@ int main(int argc, char* argv[])
     if (check_retval(&retval, "KINSetDampingFn", 1)) { return (1); }
   }
 
+  if (uopt->depth_fn)
+  {
+    /* Attach user defined depth function */
+    retval = KINSetDepthFn(kmem, DepthFn);
+    if (check_retval(&retval, "KINSetDepthFn", 1)) { return (1); }
+  }
```
Steven-Roberts, line 250: The if statement isn't really needed

src/kinsol/kinsol.c
```diff
@@ -3010,50 +2975,148 @@ static int KINFP(KINMem kin_mem)
...
+  }
+
+  /* Delete left-most column vector from QR factorization */
+  sunrealtype a, b, temp, c, s;
+
+  for (int i = 0; i < depth - 1; i++)
+  {
+    a                          = R[(i + 1) * depth + i];
+    b                          = R[(i + 1) * depth + i + 1];
+    temp                       = SUNRsqrt(a * a + b * b);
```
Steven-Roberts, line 2998: `hypot` would be more accurate. Looks like a Givens rotation, and there's a few tricks to improve the numerical stability: https://en.wikipedia.org/wiki/Givens_rotation#Stable_calculation. Not sure how important...

src/kinsol/kinsol.c
```diff
@@ -3010,50 +2975,148 @@ static int KINFP(KINMem kin_mem)
...
+{
+  /* local dot product flag for single buffer reductions */
+  sunbooleantype dotprodSB = SUNFALSE;
+
+  if ((kin_mem->kin_vtemp2->ops->nvdotprodlocal ||
+       kin_mem->kin_vtemp2->ops->nvdotprodmultilocal) &&
+      kin_mem->kin_vtemp2->ops->nvdotprodmultiallreduce)
+  {
+    dotprodSB = SUNTRUE;
+  }
```
Steven-Roberts, line 2989: Would it be easy to move this compatibility check outside the iteration loop?

src/kinsol/kinsol.c
```diff
@@ -3010,50 +2975,148 @@ static int KINFP(KINMem kin_mem)
...
+    N_VScale(ONE, kin_mem->kin_vtemp2, Q[i]);
+  }
+
+  /* Shift R to the left by one. */
+  for (int i = 1; i < depth; i++)
+  {
+    for (int j = 0; j < depth - 1; j++)
+    {
+      R[(i - 1) * depth + j] = R[i * depth + j];
+    }
```
Steven-Roberts, line 3026: Optional suggestion: use `memmove`

src/kinsol/kinsol.c
```diff
@@ -3010,50 +2975,148 @@ static int KINFP(KINMem kin_mem)
...
+    {
+      /* Move the left-most column vector (oldest value) to the end so it gets
+         overwritten with the newest value below. */
+      N_Vector tmp_dg = kin_mem->kin_dg_aa[0];
+      N_Vector tmp_df = kin_mem->kin_df_aa[0];
+      for (long int i = 1; i < kin_mem->kin_m_aa; i++)
+      {
+        kin_mem->kin_dg_aa[i - 1] = kin_mem->kin_dg_aa[i];
+        kin_mem->kin_df_aa[i - 1] = kin_mem->kin_df_aa[i];
+      }
```
Steven-Roberts, line 3092: Optional suggestion: use `memmove`

src/kinsol/kinsol.c
```diff
@@ -3085,121 +3148,117 @@ static int AndersonAcc(KINMem kin_mem, N_Vector gval, N_Vector fv, N_Vector x,
...
+      return KIN_SUCCESS;
     }
 
-    /* Add the new df vector */
-    kin_mem->kin_qr_func(kin_mem->kin_q_aa, R, kin_mem->kin_df_aa[i_pt],
-                         (int)kin_mem->kin_m_aa - 1, (int)kin_mem->kin_m_aa,
-                         (void*)kin_mem->kin_qr_data);
+    /* TODO(DJG): In the future, update QRDelete to support removing arbitrary
+       columns from the factorization */
+    if (new_depth < kin_mem->kin_current_depth)
```
Steven-Roberts, line 3231: Just want to double check this should be `if` and not `else if`

src/kinsol/kinsol.c
```diff
@@ -3085,121 +3148,117 @@ static int AndersonAcc(KINMem kin_mem, N_Vector gval, N_Vector fv, N_Vector x,
...
+        for (long int i = 1; i < kin_mem->kin_current_depth; i++)
+        {
+          kin_mem->kin_dg_aa[i - 1] = kin_mem->kin_dg_aa[i];
+          kin_mem->kin_df_aa[i - 1] = kin_mem->kin_df_aa[i];
+        }
+        kin_mem->kin_dg_aa[kin_mem->kin_current_depth - 1] = tmp_dg;
+        kin_mem->kin_df_aa[kin_mem->kin_current_depth - 1] = tmp_df;
+
+        retval = AndersonAccQRDelete(kin_mem, kin_mem->kin_q_aa, R,
+                                     (int)kin_mem->kin_current_depth);
```
Steven-Roberts, line 3250: Why not just have the last argument be `long int` to avoid casts?

doc/kinsol/guide/source/Usage/index.rst
gardner48: I was thinking it would an array with which indices to remove but using `sunbooleantype *` would be clearer. I'll switch to that.

doc/kinsol/guide/source/Mathematics.rst
gardner48: Some approaches for selecting the depth need to examine the system that will be solved for the acceleration weights. So we have to add a new column to `F_{n}` in this step before selecting the depth. The requires orthogonalizing the new vector against the existing vectors in `F_{n}` but this is usually a (relatively) small cost.

src/kinsol/kinsol.c
gardner48: Something to consider in the future (maybe when updating this to remove multiple columns). For now this function just encapsulates the exiting functionality.

src/kinsol/kinsol.c
gardner48: I'll add this to the issue for future enhancements

src/kinsol/kinsol.c
gardner48: I'll add this to the issue for future enhancements

drreynolds (approved):

PR #688, opened 2025-04-10
Fix Caliper compile test

balos1:


drreynolds (approved):

cmake/tpl/SundialsCaliper.cmake
```diff
@@ -86,16 +70,17 @@ if(CALIPER_FOUND AND (NOT CALIPER_WORKS))
...
     COMPILE_OK ${CALIPER_TEST_DIR}
-    ${CALIPER_TEST_DIR} ltest
+    ${CALIPER_TEST_DIR}/ltest.${_ext}
+    LINK_LIBRARIES caliper
     OUTPUT_VARIABLE COMPILE_OUTPUT)
 
+  # To ensure we do not use stuff from the previous attempts, we must remove the
+  # CMakeFiles directory.
+  file(REMOVE_RECURSE ${CALIPER_TEST_DIR}/CMakeFiles)
+
```
gardner48, line 83: `try_compile` cleans up after itself 
```suggestion
```

cmake/tpl/SundialsCaliper.cmake
```diff
@@ -56,28 +53,15 @@ if(CALIPER_FOUND AND (NOT CALIPER_WORKS))
...
-    "set(CMAKE_C_FLAGS_MINSIZE \"${CMAKE_C_FLAGS_MINSIZE}\")\n"
-    "add_executable(ltest ltest.c)\n"
-    "target_include_directories(ltest PRIVATE \"${caliper_INCLUDE_DIR}\")\n"
-    "target_link_libraries(ltest \"-L${caliper_LIB_DIR}\")\n"
-    "target_link_libraries(ltest caliper)\n")
+  if(CXX_FOUND)
+    set(_ext cpp)
+  else()
+    set(_ext c)
+  endif()
```
gardner48, line 60: Why does this need to be a C or C++ file?

balos1: There seems to be a bug (or at least a strange behavior when using `nvhpc`) with `try_compile` where if C++ is enabled in the parent project, the `try_compile` subproject fails with

```
CMake Error in CMakeLists.txt:
  No known features for CXX compiler

  ""

  version .
```

but if you build `ltest` as a C++ code then it identifies the features fine.

cmake/tpl/SundialsCaliper.cmake
balos1: To be clear, `try_compile` cleans up after itself only when using it in the "source file" mode (like we are now). This line was leftover from before where we used the "project" mode.

cmake/tpl/SundialsCaliper.cmake
gardner48: Interesting, it would be good to add a comment about this issue for future reference

balos1:
> @balos1 for some reason the GitHub CI seems to be stuck. Any chance this is related to `/autofix` somehow (looks like prior `/autofix` pushes might not have run all the GitHub CI tests too)?

It does look like `/autofix` causes the GitHub CI to get stuck. Closing and re-opening the PR unstuck them.

gardner48 (approved):

PR #698, opened 2025-04-29
Use different colors for fast and slow in MRI log file plot

Steven-Roberts:


Steven-Roberts:
\autofix

scripts/spack/packages/sundials/package.py
gardner48, line 1: @balos1 Any issues with this file getting formatted differently from the Spack style?

tools/log_example_mri.py
```diff
@@ -50,34 +67,74 @@ def main():
...
+    steps_s, times_s, vals_s = sunlog.get_history(
+        log,
+        "h",
+        step_range=args.step_range,
+        time_range=args.time_range,
+        group_by_level=True,
+    )
+
+    # get data for error test failures
+    steps_etf, times_etf, vals_etf = sunlog.get_history(
```
gardner48, line 79: Do you want to split up the failures by level as well?

gardner48 (commented):
Two minor questions but the formatting changes should be broken out into a separate PR so we can add the commit with just the formatting updates to `.git-blame-ignore-revs`

scripts/spack/packages/sundials/package.py
gardner48: Looks like Spack only changes two of the default settings in black:
```
line-length = 99
skip_magic_trailing_comma = true
```

tools/log_example_mri.py
Steven-Roberts: I tried that to start, but I though it created too many legend entries. I found it generally easy to distinguish failures of the different levels.

scripts/spack/packages/sundials/package.py
Steven-Roberts: Should we have a special case in the formatting script for that file then?

gardner48: That or adopt the Spack style in general. Either is good with me. @balos1 do you have a preference?

balos1: I think we should adopt the Spack style in general.

gardner48 (approved):

PR #699, opened 2025-05-05
Feature/more python formatting

Steven-Roberts:
Just the Python formatting improvements from #698

### DIFF
--- a/.cmake-format.py
+++ b/.cmake-format.py
@@ -3,329 +3,359 @@
 # ----------------------------------
 with section("parse"):
 
-  # Specify structure for custom cmake functions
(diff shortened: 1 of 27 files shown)


balos1:
To match Spack's formatting, I think we should add a pyproject.toml in the root of the repot with the two relevant settings: 

```
[tool.black]
line-length = 99
skip_magic_trailing_comma = true
``` 

Since the Python interfaces will add a bunch of Python code, we may want to set other configuration things in the near future, so a pyproject.toml file will be useful to have.

balos1 (approved):

PR #565, opened 2024-09-08
Feautre/adapt opt

Steven-Roberts:
Draft status while doing performance comparisons.

This branches of off`feature/rk_defaults` with `feature/min-err` merged in.

Current adaptivity controller changes:

- No deadzone
- Bias: 1.0
- Safety factor: 0.9
- Adjustment: 0
- I controller for ARKStep and ERKStep (to be changed based on Sylvia’s testing)

PR #705, opened 2025-05-15
fix default value of eta fixed step upper bound in arkode math section

balos1:
The default value was changed to `1.0` in v6.3.0 according to [here](https://sundials.readthedocs.io/en/latest/arkode/Usage/User_callable.html#c.ARKodeSetFixedStepBounds). [Code link](https://github.com/LLNL/sundials/blob/main/src/arkode/arkode_adapt_impl.h#L44).


balos1:
Value change was made in https://github.com/LLNL/sundials/pull/565.

drreynolds (approved):

PR #713, opened 2025-06-03
fix: correct parameter order in declaration of arkCheckConstraints

Dearsomeone:
### Problem
The function declaration of `arkCheckConstraints` in `arkode_impl.h` had a parameter order that was inconsistent with its definition in the source file.

Specifically, the `nflag` and `constrfails` parameters were swapped. This inconsistency can lead to undefined behavior when the function is used across multiple translation units.

### Fix
Reordered the parameters in the function declaration to match the definition.

### Impact
Ensures consistency between the function's declaration and its implementation, preventing misuse and potential compilation/linking issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,6 +129,9 @@ variables. It is currently an alias for `long int`.
 
 #### ARKODE
 
+Fixed a bug in the declaration of `arkCheckConstraints` by reordering parameters
+to match its definition.
+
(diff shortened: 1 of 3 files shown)


drreynolds (approved):
I wouldn't consider this a "bug", given that this is an internal function and prototypes are never required to declare the argument name, but I agree that this PR fixes an inconsistency.

PR #732, opened 2025-06-26
Use long specifier instead of long int

balos1:
This is the first of a series of PRs needed for the Python bindings to SUNDIALS

The Python binding generator we are using does not recognize `long int` as a specifier.  This PR changes `long int` to the equivalent specifier `long` in our header files (which is what the generator parses). There is no need to do this in the `src` directory, but if we want to do it for consistency sakes, we can.


balos1:
In addition to updating uses in `src/` another consideration is if we should update the uses in `doc/` to match as well. Again, updating these would just be for consistency wrt readability as `long` and `long int` are the exact same type.

Steven-Roberts (approved):
It would be nice to be consistent everywhere eventually, but I don't see the need to take care of that in this PR. I suppose examples and unit tests would also need updating.

gardner48:
Ultimately most of these occurrences should be replaced with `suncountertype` but I'm not sure if we want to make that switch right now since it's not strictly necessary of the Python work.

PR #741, opened 2025-07-07
Bugfix: MERK Unordered Groups

gardner48:
Update MERK to use the minimum stage number in a group when computing the fast time scale forcing. This fixes a bug where, in methods with unordered stages in a group (MERK43 and MERK54), the forcing computation would include stage RHS vectors that had not been computed yet. These vectors are scaled by zero, so in most cases would not impact results but do add extra vector operations. However, in cases where these vectors contain `inf` or `nan`, this would lead to erroneous forcing terms.


drreynolds (commented):
Interesting catch; did the `ark_kpr_mri.c` example end up with `inf` or `nan` in the RHS for a test on some architecture?

drreynolds (approved):

PR #742, opened 2025-07-09
Feature: KINSOL relax requirements on KINSetMAA and KINSetOrthAA

gardner48:
Update `KINSetMAA` and `KINSetOrthAA` to allow for setting the Anderson acceleration depth and orthogonalization method after `KINInit`. Additionally, `KINSetMAA` and `KINSetNumMaxIters` may now be called in any order.


gardner48:
@drreynolds This should fix the current CI failures with KINSOL in #662

drreynolds (changes requested):
Looks great. I saw only a couple of minor items to fix in the docs.

PR #749, opened 2025-07-22
Feature: Command-line control over scalar-valued set routines

drreynolds:
This PR adds functions to set package options from command line args, and includes hooks for eventual support for input file control over solver options.


src/sunlinsol/spgmr/fmod_int32/fsunlinsol_spgmr_mod.f90
```diff
@@ -142,6 +158,20 @@ function swigc_FSUNLinSolSetScalingVectors_SPGMR(farg1, farg2, farg3) &
 integer(C_INT) :: fresult
 end function
 
+function swigc_FSUNLinSolSetOptions_SPGMR(farg1, farg2, farg3, farg4, farg5) &
```
gardner48, line 161: Ignore `SetOptions` in swig file

src/sunlinsol/spgmr/sunlinsol_spgmr.c
```diff
@@ -134,6 +146,114 @@ SUNLinearSolver SUNLinSol_SPGMR(N_Vector y, int pretype, int maxl,
...
+  {
+    /* File-based option control is currently unimplemented */
+    return SUN_ERR_NOT_IMPLEMENTED;
+  }
+
+  if (argc > 0 && argv != NULL)
+  {
+    int retval = setFromCommandLine_SPGMR(S, LSid, argc, argv);
+    if (retval != SUN_SUCCESS) { return retval; }
+  }
```
gardner48, line 167:
```suggestion
  SUNFunctionBegin(C->sunctx);

  /* File-based option control is currently unimplemented */
  SUNAssert((file_name == NULL || strlen(file_name) == 0), SUN_ERR_ARG_INCOMPATIBLE)

  if (argc > 0 && argv != NULL)
  {
    SUNCheckCall(setFromCommandLine_SPGMR(S, LSid, argc, argv));
  }
```

src/sunlinsol/sptfqmr/fmod_int32/fsunlinsol_sptfqmr_mod.f90
```diff
@@ -131,6 +147,20 @@ function swigc_FSUNLinSolSetScalingVectors_SPTFQMR(farg1, farg2, farg3) &
 integer(C_INT) :: fresult
 end function
 
+function swigc_FSUNLinSolSetOptions_SPTFQMR(farg1, farg2, farg3, farg4, farg5) &
```
gardner48, line 150: Ignore `SetOptions` in swig file

src/sunlinsol/sptfqmr/sunlinsol_sptfqmr.c
```diff
@@ -149,6 +161,100 @@ SUNLinearSolver SUNLinSol_SPTFQMR(N_Vector y, int pretype, int maxl,
...
+  {
+    /* File-based option control is currently unimplemented */
+    return SUN_ERR_NOT_IMPLEMENTED;
+  }
+
+  if (argc > 0 && argv != NULL)
+  {
+    int retval = setFromCommandLine_SPTFQMR(S, LSid, argc, argv);
+    if (retval != SUN_SUCCESS) { return retval; }
+  }
```
gardner48, line 182:
```suggestion
  SUNFunctionBegin(C->sunctx);

  /* File-based option control is currently unimplemented */
  SUNAssert((file_name == NULL || strlen(file_name) == 0), SUN_ERR_ARG_INCOMPATIBLE)

  if (argc > 0 && argv != NULL)
  {
    SUNCheckCall(setFromCommandLine_SPTFQMR(S, LSid, argc, argv));
  }
```

src/sunlinsol/superlumt/sunlinsol_superlumt.c
```diff
@@ -203,6 +217,73 @@ SUNLinearSolver SUNLinSol_SuperLUMT(N_Vector y, SUNMatrix A, int num_threads,
...
+  {
+    /* File-based option control is currently unimplemented */
+    return SUN_ERR_NOT_IMPLEMENTED;
+  }
+
+  if (argc > 0 && argv != NULL)
+  {
+    int retval = setFromCommandLine_SuperLUMT(S, LSid, argc, argv);
+    if (retval != SUN_SUCCESS) { return retval; }
+  }
```
gardner48, line 238:
```suggestion
  SUNFunctionBegin(C->sunctx);

  /* File-based option control is currently unimplemented */
  SUNAssert((file_name == NULL || strlen(file_name) == 0), SUN_ERR_ARG_INCOMPATIBLE)

  if (argc > 0 && argv != NULL)
  {
    SUNCheckCall(setFromCommandLine_SuperLUMT(S, LSid, argc, argv));
  }
```

src/sunnonlinsol/fixedpoint/fmod_int32/fsunnonlinsol_fixedpoint_mod.f90
```diff
@@ -120,6 +136,20 @@ function swigc_FSUNNonlinSolSetConvTestFn_FixedPoint(farg1, farg2, farg3) &
 integer(C_INT) :: fresult
 end function
 
+function swigc_FSUNNonlinSolSetOptions_FixedPoint(farg1, farg2, farg3, farg4, farg5) &
```
gardner48, line 139: Ignore `SetOptions` in swig file

src/sunnonlinsol/newton/fmod_int32/fsunnonlinsol_newton_mod.f90
```diff
@@ -137,6 +153,20 @@ function swigc_FSUNNonlinSolSetConvTestFn_Newton(farg1, farg2, farg3) &
 integer(C_INT) :: fresult
 end function
 
+function swigc_FSUNNonlinSolSetOptions_Newton(farg1, farg2, farg3, farg4, farg5) &
```
gardner48, line 156: Ignore `SetOptions` in swig file

src/sunnonlinsol/newton/sunnonlinsol_newton.c
```diff
@@ -397,6 +408,65 @@ SUNErrCode SUNNonlinSolSetConvTestFn_Newton(SUNNonlinearSolver NLS,
...
+  {
+    /* File-based option control is currently unimplemented */
+    return SUN_ERR_NOT_IMPLEMENTED;
+  }
+
+  if (argc > 0 && argv != NULL)
+  {
+    int retval = setFromCommandLine_Newton(NLS, NLSid, argc, argv);
+    if (retval != SUN_SUCCESS) { return retval; }
+  }
```
gardner48, line 425:
```suggestion
  SUNFunctionBegin(C->sunctx);

  /* File-based option control is currently unimplemented */
  SUNAssert((file_name == NULL || strlen(file_name) == 0), SUN_ERR_ARG_INCOMPATIBLE)

  if (argc > 0 && argv != NULL)
  {
    SUNCheckCall(setFromCommandLine_Newton(NLS, NLSid, argc, argv));
  }
```

gardner48 (changes requested):

src/sunadaptcontroller/imexgus/sunadaptcontroller_imexgus.c
gardner48, line 1: Continuing a conversations started in #662. In the original thread started [here](https://github.com/LLNL/sundials/pull/662#discussion_r2072196366), @drreynolds commented:

> I am concerned with how the branching (file-based input vs command-line input) would occur at two separate levels, given the single SetOptions points of entry for both the base class and the implementations. If we want the base class to handle some keys before calling the implementations, then it would need to first branch between file-based vs command-line input, call its private function for file-based input, call its private function for command-line input, and then call the implementation's SetOptions function pointer to repeat the process. Since both the base class and implementations will share the same prefix (here, sunadaptcontroller), then I am concerned that doing (a) base class file-based, then (b) base class command-line, then (c) implementation file-based, then (d) implementation command-line, could open things up for odd behavior if the keys were not sufficiently distinct.

I'm not quite sure I understand the issue. The base class and derived class (like ARKODE steppers) should not have overlapping keys so the base class processes the only common keys first from a file then from the command line. Then the derived class processes only its keys first from a file then from the command line. The problems I see are (1) if we need to some how interleave common keys with implementation specific keys i.e., the Set calls need to be made in a specific order and (2) calling the derived class `SetOptions` would only set the implementation specific keys. We've generally tried to avoid (1) but maybe there are cases where a certain order is necessary. For (2), we could make the implementation functions internal (not place them in the public header), so users must call the generic version which is different from how we typically do things.

gardner48: One other thing to note about having the implementation process all the options is it requires implementing `SetOptions` even when there are only common options supported e.g., Newton and FixedPoint only have the `max_iters` option.

src/sundials/sundials_cli.h
```diff
@@ -0,0 +1,207 @@
...
+/* utilities for integer "set" routines */
+typedef int (*sunIntSetFn)(void*, int);
+
+struct sunKeyIntPair
+{
+  const char* key;
+  sunIntSetFn set;
+};
+
+SUNDIALS_EXPORT
```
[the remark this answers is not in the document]
gardner48: I'll take a closer look at this. I didn't see any tests that access these internal function directly so `SUNDIALS_EXPORT` shouldn't be needed.

src/sunadaptcontroller/imexgus/sunadaptcontroller_imexgus.c
drreynolds: I'm not too concerned with (2) since it's already standard practice in SUNDIALS to avoid order dependencies for calling "Set" routines on a specific module.

For (1) I suppose we will just need to remain diligent as authors/reviewers.  Fortunately, for now there are only a few places where base classes could handle command-line arguments (SUNAdaptController, SUNLinearSolver, SUNNonlinearSolver), and none of those have a large number of options.

I may try this out in SUNNonlinearSolver, since it's the simplest.  However, I still maintain that any changes here will not affect the API, and thus I don't see a strong reason to hold off on approving/merging this PR to wait for such a minor "code maintentance" optimization to be completed.

gardner48 (changes requested):
A few minor items, otherwise this looks good.

src/sundials/sundials_adaptcontroller.c
```diff
@@ -96,6 +102,65 @@ SUNAdaptController_Type SUNAdaptController_GetType(SUNAdaptController C)
...
+        free(prefix);
+        return retval;
+      }
+      continue;
+    }
+
+    /* Note that although SUNAdaptController_Write is part of the base class,
+       it should NOT be called until all options have been set, so we process
+       that in the implementations instead of here. */
+  }
```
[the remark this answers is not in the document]
drreynolds: I thought of that, but if it were in this file then it would be called *before* the implementations perform their own command-line processing, so the output would be inaccurate.  The reason this works in ARKODE is that the stepper function is called repeatedly, once for each command-line argument.  However, here, the implementation processing routines are called only once.

gardner48 (dismissed):

gardner48 (approved):

balos1 (approved):

PR #756, opened 2025-08-05
Feature: Provide access to IDA interpolant via new IDAGetInterpData function

zdodson:
Adds new public API function `IDAGetInterpData` to `ida.h` (implemented in `ida_io.cpp`) permitting access to IDA polynomial interpolant constructed during the last successful solver step.  This permits users to extract and save the necessary data to construct a piecewise-defined interpolant across multiple steps.  This can be useful for offline evaluation, checkpointing, custom output, etc.

SUNDIALS PR Checklist:
- [ ] After your PR is opened, ensure that all of the tests are passing (a SUNDIALS developer will have to allow the testing to run).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ setting the Anderson acceleration depth and orthogonalization method after
 `KINInit`. Additionally, `KINSetMAA` and `KINSetNumMaxIters` may now be called
 in any order.
 
+A new `IDAGetInterpData` function has been added to return the internal data
+defining the polynomial interpolant constructed during the last successful
+solver step.  This permits the construction of a piecewise-defined interpolant
+across multiple steps, useful for offline evaluation, checkpointing, custom
+output, etc.
+
 ### Bug Fixes
 
 The shared library version numbers for the oneMKL dense linear solver and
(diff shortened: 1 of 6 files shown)


Steven-Roberts:
Hi @zdodson, thanks for the sharing this PR. The existing functions `IDAGetDky` and `IDAGetCurrentStep` can, in principle, also be used to extract and save the data for a piecewise interpolant. Would that work for your needs? I'll let others on the team weigh in, but I have some reservations about exposing the lower-level `phi` and `psi` properties to users. User would have to be careful to only read from the properties or risk corrupting IDA's internal state.

zdodson:
Thank you, @Steven-Roberts, for examining the PR and commenting.  I quite agree that one can reconstruct the interpolating polynomial using the existing IDA functions as you describe, although it involves several calls to `IDAGetDky` to obtain the coefficients of all derivatives.  (Evaluating the recurrence for increasing values of `k` does introduce some inefficiencies.)

I appreciate your remarks about the risks of exposing IDA internals and quite understand if this PR is not aligned with SUNDIALS' trajectory or design goals.

IMO, it would be splendid to formalize the interpolating polynomial and make it more of a first-class object that users can interact with directly.  That said, this is a much more extensive change and, if done, should probably be applied across the entire suite of solvers rather than just IDA.

Thanks again for your comments and for your team's continued contributions to this excellent solver suite!

PR #764, opened 2025-09-16
Bugfix: KINSOL initialize Anderson acceleration depth

gardner48:
* Initialize the Anderson acceleration depth every `KINSol` call
* Fix a logging bug where KINSOL logging messages would not be output


gardner48:
Fixes https://github.com/mfem/mfem/issues/5004

PR #775, opened 2025-09-30
Docs: KINSOL and ARKODE example doc updates

gardner48:
* Add KINSOL example docs with Sphinx
* Remove old Shpinx themes and templates in ARKODE example docs
* Use shared bib and tex files in ARKODE example docs
* Split intro and landing page in ARKODE example docs
* Remove old README files
* Add index-superbuld.rst for example docs in superbuild
* Clean up some KINSOL examples
* Remove doc source from tarballs


doc/superbuild/source/index.rst
```diff
@@ -201,6 +201,16 @@ SUNDIALS License and Notices
    Changelog_link.rst
    FAQ_link.rst
 
+..
+   Uncomment once all example docs are updated
```
balos1: Should we just add this section now and add placeholders with links to the PDF versions for the other packages?

gardner48: Sounds good, I'll add a landing page for each package that directs users to the PDFs.

gardner48: Done, the other packages now have a landing page with a link to download the PDF version.

balos1:
I think it would be nice if the source code for the example was linked to at the top of every example subsection. Otherwise I think this is looking great. Just something like  "View the Source" as a link. This could be something for the future though.

balos1 (dismissed):

PR #777, opened 2025-10-03
add google analytics back

balos1:
readthedocs stopped supporting google analytics natively, so this is the only way to get it back.

### DIFF
--- a/doc/superbuild/source/conf.py
+++ b/doc/superbuild/source/conf.py
@@ -53,6 +53,12 @@ intersphinx_mapping = {
     "sphinx": ("https://www.sphinx-doc.org/en/master", ("../objects.inv", None))
 }
 
+# Google analytics
+if os.getenv("GITHUB_ACTIONS"):
+    extensions.append("sphinxcontrib-googleanalytics")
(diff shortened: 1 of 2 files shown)


doc/superbuild/source/conf.py
```diff
@@ -53,6 +53,13 @@
     "sphinx": ("https://www.sphinx-doc.org/en/master", ("../objects.inv", None))
 }
 
+# Only setup Google analytics for the readthedocs being deployed (not local).
+# We can do this by checking if we are running in GitHub actions.
+if os.getenv("GITHUB_ACTIONS"):
+    extensions.append("sphinxcontrib.googleanalytics")
+    googleanalytics_id = "G-3KESEG9QED"
+    googleanalytics_enabled = True
```
gardner48, line 61: When we build the docs as part of the CI tests on GitHub, will that cause any issues for the analytics?

drreynolds (approved):
LGTM

doc/superbuild/source/conf.py
balos1: It should not.

PR #780, opened 2025-10-06
CMake: add C23 option

gardner48:
Add C23 option to `CMAKE_C_STANDARD`


drreynolds (approved):
LGTM.  Have we confirmed that SUNDIALS can build without issue in C23?

balos1 (approved):

gardner48:
> Have we confirmed that SUNDIALS can build without issue in C23?

It works with gcc 15 (C23 became the default in gcc 15). I don't think other compilers have full C23 support yet.

PR #783, opened 2025-10-07
run ci only if [run-ci] is in commit message

balos1:
This will limit the GitHub actions CI to only commits with "[run-ci]" in the commit message. This will ensure we are not running the GitHub actions CI more than necessary.

### DIFF
--- a/.github/workflows/check-format.yml
+++ b/.github/workflows/check-format.yml
@@ -10,6 +10,8 @@ concurrency:
 
 jobs:
   format_check:
+    if: contains(github.event.head_commit.message, '[run-ci]') # This job only runs if commit message contains '[run-ci]'
+
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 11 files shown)


balos1:
With this requirement, we could also allow CI to run on pushes to branches, provided the [run-ci] tag is in the commit message.

drreynolds (dismissed):
This looks fine, but can you clarify the purpose of the PR?  Is it just that we run the CI too frequently and want to limit this?  Or is this a "security" question?

balos1:
We run CI too frequently and want to limit it, especially if we start hosting our own github runners. We can discuss this more at our next team meeting. For now, we should not merge this. I meant to open it as a draft.

balos1:
Ill also note that GitHub supports a skip ci command by default: https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs. We could use an opt-out approach instead.

balos1:
With the new CI staging, this may not be needed.

PR #784, opened 2025-10-07
Fix bug in SUNStepper_ResetCheckpointIndex

balos1:


gardner48 (approved):

drreynolds (approved):
It would be ideal if this error could have been caught by an example or unit test.

balos1:
> It would be ideal if this error could have been caught by an example or unit test.

Any example will also provide the Reset function, which is why it was not caught. We would have to start adding much more granular unit tests to catch these types of bugs (which I am all for as long as we commit to using googletest for them so the boiler plate is less).  

I actually caught it through smoke tests I am setting up in the Python branch to make sure that user-supplied functions are being handled properly. The smoke test just creates a SUNStepper and then sets only one of the user-supplied functions at a time, and checks to see if it is called.

PR #787, opened 2025-10-19
Bugfix: CVODE(S) Inequality Constraints

### LINKED ISSUE #702: [BUG] CVODE attempts an "infinite" step when constraints are enabled, then fails unrecoverably
### Current Behavior:

When solving a large system of ODEs that has non-negative constraints enabled, CVode will try to find a new stepsize after constraint checking fails. On several occasions, I have observed it trying an "infinite" step (of order 1e+308), which naturally fails unrecoverably, and the solver terminates with an error. This even happens if a maximum stepsize is set explicitly.

### Expected Behavior:

With a set maximum stepsize, I would expect that setting to take precedence no matter what. Even without it, an infinite stepsize is obviously unreasonable and bound to fail.

### Steps To Reproduce:

I do not currently have a MWE (apologies for that), and, while the behavior is deterministic for a particular build, small changes to the actual values used in the problem can actually sidestep the issue entirely. I'm not sure if this is a bug somewhere in my code, which I cannot currently share, but I have tracked the source of the infinity to the line referenced below and the behavior of `N_VMinQuotient` as specified in the documentation.

### Environment:

- SUNDIALS version: 7.2.1
- OS: Kubuntu 24.04
- Compiler: Clang 20.0.0

### Anything else:

I believe the issue stems from [this line of code](https://github.com/LLNL/sundials/blob/main/src/cvode/cvode.c#L3211). Because `N_VMinQuotient` should return `SUN_BIG_REAL` if there are no nonzero denominators, and because all the subsequent operations take the maximum value of eta, the problematic value will always be chosen and an infinite stepsize is attempted, _even when a maximum stepsize is explicitly set with `CVodeSetMaxStep`_. I haven't yet figured out what the absence of nonzero denominators means, though.

gardner48:
Fix a bug in the CVODE(S) inequality constraint handling where the predicted state was used to compute the step size reduction factor instead of the prior solution which could lead to an incorrect reduction in the step size or, when the prediction violates the constraints, an infinitely large step size in the next step attempt. (Fixes #702)

Add a unit test for inequality constraint handling with all the integrators.

In CVODE(S) and IDA(S), separate inequality constraint check from the nonlinear solve check.

Add functions to set the maximum number of inequality constraint failures and get total number of failures (constraint failures are no longer included in the number of step failures due to a solver failure).

Add a function to get the number of steps modified but not failed due to constraint violations. 

Synchronize some differences between CVODE/CVODES and IDA/IDAS source and docs.

Add missing return flag names to `CVodeGetReturnFlagName` in CVODES and `IDAGetReturnFlagName` in IDA.


examples/cvode/F2003_serial/cv_roberts_dns_constraints_f2003.out
```diff
@@ -25,19 +25,19 @@ Constraints cval = ( 1.00  1.00  1.00 )
   4.0000E+07   5.1479E-05   2.0593E-10   9.9995E-01
   4.0000E+08   6.0238E-06   2.4095E-11   9.9999E-01
   4.0000E+09   4.6711E-07   1.8685E-12   1.0000E+00
-  4.0000E+10   3.4054E-09   1.3622E-14   1.0000E+00
+  4.0000E+10  -1.2752E-24   4.5031E-30   1.0000E+00
```
gardner48: I suspect the negative solution values here are from interpolating the output (since they're essentially zero). I'll double check if that's the case.

examples/ida/F2003_parallel/ida_heat2D_kry_bbd_f2003.out
```diff
@@ -27,16 +27,16 @@
...
-    0.640000    1.06183E-02   1   84   125   603   125+603   2958    5.52E-02   29   728
-    1.280000    1.98532E-02   1   88   131   621   131+621   3162    2.21E-01   31   752
-    2.560000    3.00701E-02   1   92   138   651   138+651   3264    3.52E-01   32   789
-    5.120000    2.46999E-02   1   96   147   670   147+670   3570    1.07E+00   35   817
-   10.240000    1.82570E-02   1   98   150   682   150+682   3672    2.14E+00   36   832
+    0.640000    2.04253E-02   1   65    97   482    97+482   2550    6.90E-02   25   579
+    1.280000    2.84118E-02   1   69   103   508   103+508   2754    2.76E-01   27   611
+    2.560000    2.28392E-02   1   74   111   534   111+534   2856    7.37E-01   28   645
+    5.120000    1.50053E-02   1   76   114   546   114+546   2958    1.47E+00   29   660
+   10.240000    6.77062E-03   1   85   129   564   129+564   3876    1.89E+00   38   693
```
gardner48: Changes in the IDA output are not expected (other than solver failure counts). I'm looking what's happening in this example.

gardner48: Fixed, I was updating the step size before resetting the IDA history array.

examples/cvodes/serial/cvsRoberts_ASAi_dns_constraints.out
```diff
@@ -26,16 +26,16 @@ Backward integration from tB0 =   4.0000e+07
 --------------------------------------------------------
 returned t:   4.0000e+01
 tout:         4.0000e+01
-lambda(t):    3.9953e+07   3.9953e+07   3.9953e+07
+lambda(t):    3.9934e+07   3.9934e+07   3.9934e+07
 y(t):         7.1588e-01   9.1883e-06   2.8411e-01
 --------------------------------------------------------
 
-Done ( nst = 161 )
+Done ( nst = 242 )
```
gardner48, line 33: @drreynolds commented [here](https://github.com/sundials-codes/answers/pull/67#discussion_r2445241664) about the increased number of steps needed in the first backward integration. Because of the constraint handling bug, CVODES would originally fail the constraint test multiple times near the end of the integration before cutting the step size. As a result, the old version it does not make it as far in the forward integration as the version with the bug fix (final time of 6.95926e+07 in the old code versus 9.30497e+07 in the new code). So, in the first backward integration in the new version starts with a different initial condition and integrates over a longer time span than the old code.

examples/cvode/F2003_serial/cv_roberts_dns_constraints_f2003.out
gardner48: This appears to be due to round off differences in applying the constraint modification to the state vector versus the history correction vector. For the constraint vector $c$, the constraint modification is
```math
v_i =
\begin{cases}
y_i \mp 0.2 w_i & \quad \text{for} \quad c_i = \pm 2 \quad \text{and} \quad y_i \lesseqqgtr 0 \\
y_i & \quad \text{for} \quad c_i = \pm 1 \quad \text{and} \quad y_i \lessgtr 0 \\
0 & \quad \text{otherwise}
\end{cases}
```
so that $y_i - v_i$ will satisfy the constraints. 

If $\lVert v \rVert$ is sufficiently small, the correction vector is updated using $v$ as `acor_new = acor - v`. Expanding the `acor` update (and masking entries appropriately) we have `acor_new = acor - v = acor - (y - x) = acor - (y_p + acor - x) = -y_p + x` where `y_p` is the predictor, `zn[0]`, and `x` is `0.2 w` or `0`. So, the history update `zn[0] += acor_new` should satisfy the constraints. 

`y_new = y - v` satisfies the constraints but `zn[0] += acor` might not depending on round off. In this test the last two steps have very small negative values. Breaking the update into three steps (with appropriate masking): `acor -= acor`, `acor -= y_p`, and `acor += x` eliminates the small negative values however, (at least on my laptop) towards the end of the simulation a step hits the max error test failure limit (although each step attempt reduces the error).

src/cvode/cvode.c
```diff
@@ -3390,9 +3392,7 @@ static int cvDoErrorTest(CVodeMem cv_mem, int* nflagPtr, sunrealtype saved_t,
 
   dsm = cv_mem->cv_acnrm * cv_mem->cv_tq[2];
 
-  SUNLogDebug(CV_LOGGER, "error-test",
-              "step = %li, h = " SUN_FORMAT_G ", dsm = " SUN_FORMAT_G,
-              cv_mem->cv_nst, cv_mem->cv_h, dsm);
+  SUNLogDebug(CV_LOGGER, "error-test", "dsm = " SUN_FORMAT_G, dsm);
```
drreynolds, line 3460: I'm surprised that this debug line removed information -- was the previous one incompatible with the current `tools/suntools/logs.py`?

src/cvode/cvode.c
```diff
@@ -3540,8 +3540,7 @@ static void cvCompleteStep(CVodeMem cv_mem)
   }
 #endif
 
-  SUNLogDebug(CV_LOGGER, "return", "nst = %d, nscon = %d", cv_mem->cv_nst,
```
drreynolds, line 3526: Similarly here, was this line incompatible with the Python log parser?

doc/cvode/guide/source/Usage/index.rst
```diff
@@ -2254,12 +2277,16 @@ the rootfinding algorithm.
      * ``CV_MEM_NULL`` -- The CVODE memory block was not initialized through a previous call to :c:func:`CVodeCreate`.
 
    **Notes:**
-      CVODE will not report the initial conditions as a possible zero-crossing  (assuming that one or more components :math:`g_i` are zero at the initial time).  However, if it appears that some :math:`g_i` is identically zero at the initial  time (i.e., :math:`g_i` is zero at the initial time and after the first step),  CVODE will issue a warning which can be disabled with this optional input  function.
+      CVODES will not report the initial conditions as a possible zero-crossing
+      (assuming that one or more components :math:`g_i` are zero at the initial time).
+      However, if it appears that some :math:`g_i` is identically zero at the initial
+      time (i.e., :math:`g_i` is zero at the initial time and after the first step),
+      CVODES will issue a warning which can be disabled with this optional input
+      function.
```
drreynolds:
```suggestion
      CVODE will not report the initial conditions as a possible zero-crossing
      (assuming that one or more components :math:`g_i` are zero at the initial time).
      However, if it appears that some :math:`g_i` is identically zero at the initial
      time (i.e., :math:`g_i` is zero at the initial time and after the first step),
      CVODE will issue a warning which can be disabled with this optional input
      function.
```

doc/cvodes/guide/source/Usage/SIM.rst
```diff
@@ -1253,6 +1257,23 @@ Main solver optional input functions
       simultaneous corrector option is currently disallowed and will result in an
       illegal input return.
 
+.. c:function:: int CVodeSetMaxNumConstraintFails(void* cvode_mem, int max_fails)
+
```
drreynolds, line 1261: Can this be controlled using `CVodeSetOptions`?  If so, the option name should be specified here.  If not, it should be added.

doc/cvode/guide/source/Usage/index.rst
```diff
@@ -1239,7 +1241,28 @@ Main solver optional input functions
...
    **Notes:**
-      The presence of a non-``NULL`` constraints vector that is not 0.0 in  all components will cause constraint checking to be performed.  However, a call with 0.0 in all components of ``constraints`` will  result in an illegal input return. A ``NULL`` constraints vector will disable  constraint checking.
+      The presence of a non-``NULL`` constraints vector that is not 0.0 in  all
+      components will cause constraint checking to be performed. However, a call
+      with 0.0 in all components of ``constraints`` will  result in an illegal
+      input return. A ``NULL`` constraints vector will disable  constraint
+      checking.
+
+.. c:function:: int CVodeSetMaxNumConstraintFails(void* cvode_mem, int max_fails)
+
```
drreynolds, line 1251: Can this be controlled using `CVodeSetOptions`?  If so, the option name should be specified here.  If not, it should be added.

doc/ida/guide/source/Usage/index.rst
```diff
@@ -1280,6 +1281,23 @@ Main solver optional input functions
       input return. A ``NULL`` input will disable constraint checking.
 
 
+.. c:function:: int IDASetMaxNumConstraintFails(void* ida_mem, int max_fails)
+
```
drreynolds, line 1285: Can this be controlled using `IDASetOptions`?  If so, the option name should be specified here.  If not, it should be added.

doc/idas/guide/source/Usage/SIM.rst
```diff
@@ -915,36 +915,37 @@ Main solver optional input functions
...
-   +--------------------------------------------------------------------+---------------------------------+----------------+
-   | Variable types (differential/algebraic)                            | :c:func:`IDASetId`              | NULL           |
-   +--------------------------------------------------------------------+---------------------------------+----------------+
-   | Inequality constraints on solution                                 | :c:func:`IDASetConstraints`     | NULL           |
-   +--------------------------------------------------------------------+---------------------------------+----------------+
-
+   +--------------------------------------------------------------------+---------------------------------------+----------------+
+   | **Optional input**                                                 | **Function name**                     | **Default**    |
+   +--------------------------------------------------------------------+---------------------------------------+----------------+
+   | Set IDA options from the command line or file                      | :c:func:`IDASetOptions`               |                |
```
drreynolds:
```suggestion
   | Set IDAS options from the command line or file                     | :c:func:`IDASetOptions`               |                |
```

doc/idas/guide/source/Usage/SIM.rst
```diff
@@ -1291,6 +1291,22 @@ Main solver optional input functions
       simultaneous corrector option is currently disallowed and will result in
       an illegal input return.
 
+.. c:function:: int IDASetMaxNumConstraintFails(void* ida_mem, int max_fails)
+
```
drreynolds, line 1295: Can this input be controlled using `IDASetOptions`?  If so, the option name should be stated below.  If not, then it should be added.

src/cvode/cvode.c
gardner48: The step number and size are logged at the beginning of the step attempt so I removed them from here so they're not duplicated in the output. Also we print `nst + 1`, so this was off by one when I was debugging.

src/cvode/cvode.c
gardner48: The step number is already logged at the beginning of the step attempt so this removes the duplicate output (we print `nst + 1`, so this was off by one when I was debugging). I don't remember why I only logged `nscon` and not other quantities in this function, so I removed it to reduce the output some.

examples/cvode/F2003_serial/cv_roberts_dns_constraints_f2003.out
drreynolds: This is helpful information, but I think it leaves us in a difficult position.  If a user code applies something like `sqrt` or `log` to the new time step then keeping the current approach would create `NaN` or `Inf` throughout the simulation.  However, the current approach seems more robust than the proposed 3-step process.

How much do you think it is just this problem (Roberts) causing trouble, versus something potentially more indicative of user applications?

src/ida/ida_bbdpre.c
```diff
@@ -18,7 +18,7 @@
  * -----------------------------------------------------------------
  * This file contains implementations of routines for a
  * band-block-diagonal preconditioner, i.e. a block-diagonal
- * matrix with banded blocks, for use with IDA, the IDASPILS
+ * matrix with banded blocks, for use with IDA, the IDASLS
```
drreynolds:
```suggestion
 * matrix with banded blocks, for use with IDA, the IDALS
```

src/ida/ida_io.c
```diff
@@ -610,23 +608,67 @@ int IDASetConstraints(void* ida_mem, N_Vector constraints)
...
+  if (max_fails <= 0) { IDA_mem->max_constraint_fails = MAX_CONSTRAINT_FAILS; }
+  else { IDA_mem->max_constraint_fails = max_fails; }
+
+  return IDA_SUCCESS;
+}
+
+/*
+ * IDASetMaxNumConstraintFails
+ *
+ * Setup for constraint handling feature
```
drreynolds:
```suggestion
 * IDAGetMaxNumConstraintFails
 *
 * Routine to retrieve the number of constraint failures
```

src/idas/idas_io.c
```diff
@@ -628,23 +626,67 @@ int IDASetConstraints(void* ida_mem, N_Vector constraints)
...
 
   N_VScale(ONE, constraints, IDA_mem->ida_constraints);
 
-  IDA_mem->ida_constraintsSet = SUNTRUE;
-
   return (IDA_SUCCESS);
 }
 
+/*
+ * IDAGetNumConstraintFails
```
drreynolds:
```suggestion
 * IDASetMaxNumConstraintFails
```

src/idas/idas_io.c
```diff
@@ -628,23 +626,67 @@ int IDASetConstraints(void* ida_mem, N_Vector constraints)
...
+  if (max_fails <= 0) { IDA_mem->max_constraint_fails = MAX_CONSTRAINT_FAILS; }
+  else { IDA_mem->max_constraint_fails = max_fails; }
+
+  return IDA_SUCCESS;
+}
+
+/*
+ * IDASetMaxNumConstraintFails
+ *
+ * Setup for constraint handling feature
```
drreynolds:
```suggestion
 * IDAGetNumConstraintFails
 *
 * Routine to return the number of constraint failures
```

examples/cvode/F2003_serial/cv_roberts_dns_constraints_f2003.out
gardner48: Increasing the allowed number of error test failures, the step size is eventually driven to zero but the error test still fails. Suggesting the "sufficiently small" condition on `v` might be too loose. Tightening the threshold removes the failure and far fewer corrector modification are needed.

test/unit_tests/problems/constant.hpp
```diff
@@ -0,0 +1,114 @@
...
+ *
+ *   y' = a
+ *
+ * with the initial condition
+ *
+ *   y(t_0) = y_0
+ *
+ * This problem has analytical solution
+ *
+ *   y(t) = a * (t - t_0) + y_0
```
drreynolds, line 27: In the file itself, `a=1`:
```suggestion
 *   y' = 1
 *
 * with the initial condition
 *
 *   y(t_0) = y_0
 *
 * This problem has analytical solution
 *
 *   y(t) = (t - t_0) + y_0
```

examples/cvode/F2003_serial/cv_roberts_dns_constraints_f2003.out
gardner48: The new and old approach are computing the same modification to the correction so I'm not sure the current approach is more robust just a different way of applying the update (in fewer operations).

test/unit_tests/problems/constant.hpp
gardner48: The comments are correct e.g., the RHS function is `N_VScale(SUN_RCONST(1.0), udata->rhs_, ydot);` where `rhs_` is an input to the constructor.

examples/cvode/F2003_serial/cv_roberts_dns_constraints_f2003.out
gardner48: 3942e23d927f1356dc4fda202ac86148c8aefaca splits the correction update in to three steps to reduce the impact of roundoff errors.

gardner48:
@drreynolds I've updated the correction modification to address the roundoff issues and resolved the open comments. However, this PR also includes updates from #788 (makes adding logging to the constraint checks easier), so hold off on taking another pass over this PR until that one is merged.

doc/cvodes/guide/source/Usage/SIM.rst
```diff
@@ -914,28 +918,29 @@ Main solver optional input functions
...
       using the default ``cvid``, the option ``cvodes.max_order`` followed by the value
       can be used to set the maximum method order of accuracy.
 
-      CVODES options set via command-line arguments to :c:func:`CVodeSetOptions` will
-      overwrite any previously-set values.  Options are set in the order they are given
-      in ``argv`` and, if an option with the same prefix appears multiple times in
+      CVODES options set via :c:func:`CVodeSetOptions` will overwrite
+      any previously-set values.  Options are set in the order they are given in
+      ``argv`` and, if an option with the same prefix appears multiple times in
       ``argv``, the value of the last occurrence will used.
```
drreynolds:
```suggestion
      ``argv``, the value of the last occurrence will be used.
```

doc/cvodes/guide/source/Usage/SIM.rst
```diff
@@ -2447,109 +2471,115 @@ the preconditioner.
...
-   | Last return from a CVDIAG function              | :c:func:`CVDiagGetLastFlag`              |
-   +-------------------------------------------------+------------------------------------------+
-   | Name of constant associated with a return flag  | :c:func:`CVDiagGetReturnFlagName`        |
-   +-------------------------------------------------+------------------------------------------+
+   +-------------------------------------------------+--------------------------------------------+
+   |              **Optional output**                |            **Function name**               |
+   +=================================================+============================================+
+   | **CVODE main solver**                           |                                            |
+   +-------------------------------------------------+--------------------------------------------+
+   | Size of CVODE real and integer workspaces       | :c:func:`CVodeGetWorkSpace`                |
```
drreynolds: Was it intentional that these were changed from `CVODES` to `CVODE` here in the CVODES documentation?  If not, here's a suggestion to revert:
```suggestion
   | **CVODES main solver**                          |                                            |
   +-------------------------------------------------+--------------------------------------------+
   | Size of CVODES real and integer workspaces      | :c:func:`CVodeGetWorkSpace`                |
```

doc/cvodes/guide/source/Usage/SIM.rst
```diff
@@ -2447,109 +2471,115 @@ the preconditioner.
...
+   +-------------------------------------------------+--------------------------------------------+
+   | Estimated local error vector                    | :c:func:`CVodeGetEstLocalErrors`           |
+   +-------------------------------------------------+--------------------------------------------+
+   | No. of nonlinear solver iterations              | :c:func:`CVodeGetNumNonlinSolvIters`       |
+   +-------------------------------------------------+--------------------------------------------+
+   | No. of nonlinear convergence failures           | :c:func:`CVodeGetNumNonlinSolvConvFails`   |
+   +-------------------------------------------------+--------------------------------------------+
+   | All CVODE integrator statistics                 | :c:func:`CVodeGetIntegratorStats`          |
+   +-------------------------------------------------+--------------------------------------------+
+   | CVODE nonlinear solver statistics               | :c:func:`CVodeGetNonlinSolvStats`          |
```
drreynolds: As above, if the change from `CVODES` to `CVODE` here in the CVODES documentation was unintentional, here's a suggestion to revert:
```suggestion
   | All CVODES integrator statistics                | :c:func:`CVodeGetIntegratorStats`          |
   +-------------------------------------------------+--------------------------------------------+
   | CVODES nonlinear solver statistics              | :c:func:`CVodeGetNonlinSolvStats`          |
```

doc/idas/guide/source/Usage/SIM.rst
```diff
@@ -2332,91 +2348,95 @@ preconditioner.
...
-  | No. of Jacobian-vector product evaluations                         | :c:func:`IDAGetNumJtimesEvals`         |
-  +--------------------------------------------------------------------+----------------------------------------+
-  | Last return from a linear solver function                          | :c:func:`IDAGetLastLinFlag`            |
-  +--------------------------------------------------------------------+----------------------------------------+
-  | Name of constant associated with a return flag                     | :c:func:`IDAGetLinReturnFlagName`      |
-  +--------------------------------------------------------------------+----------------------------------------+
+  +--------------------------------------------------------------------+------------------------------------------+
+  | **Optional output**                                                | **Function name**                        |
+  +====================================================================+==========================================+
+  | Size of IDA real and integer workspace                             | :c:func:`IDAGetWorkSpace`                |
```
drreynolds: Was this `IDAS` -> `IDA` change intentional (here in the IDAS docs)?  If not, here's a suggestion to revert:
```suggestion
  | Size of IDAS real and integer workspace                            | :c:func:`IDAGetWorkSpace`                |
```

src/cvode/cvode_io.c
```diff
@@ -1649,6 +1712,10 @@ int CVodePrintAllStats(void* cvode_mem, FILE* outfile, SUNOutputFormat fmt)
   sunfprintf_long(outfile, fmt, SUNFALSE, "Steps", cv_mem->cv_nst);
   sunfprintf_long(outfile, fmt, SUNFALSE, "Error test fails", cv_mem->cv_netf);
   sunfprintf_long(outfile, fmt, SUNFALSE, "NLS step fails", cv_mem->cv_ncfn);
+  sunfprintf_long(outfile, fmt, SUNFALSE, "Constraint fails",
+                  cv_mem->constraint_fails);
+  sunfprintf_long(outfile, fmt, SUNFALSE, "Constraint corrections",
+                  cv_mem->constraint_corrections);
```
drreynolds, line 1718: Do you really want to print these by default, or should this be inside a conditional based on whether constraint-handling is used, e.g.
```suggestion
  if (cv_mem->cv_constraints == NULL)
  {
    sunfprintf_long(outfile, fmt, SUNFALSE, "Constraint fails",
                    cv_mem->constraint_fails);
    sunfprintf_long(outfile, fmt, SUNFALSE, "Constraint corrections",
                    cv_mem->constraint_corrections);
  }
```

(and similar question for CVODES, IDA, and IDAS)

src/cvodes/cvodes_io.c
```diff
@@ -2543,10 +2608,15 @@ int CVodePrintAllStats(void* cvode_mem, FILE* outfile, SUNOutputFormat fmt)
...
 
+  /* step and method stats */
   sunfprintf_real(outfile, fmt, SUNTRUE, "Current time", cv_mem->cv_tn);
   sunfprintf_long(outfile, fmt, SUNFALSE, "Steps", cv_mem->cv_nst);
   sunfprintf_long(outfile, fmt, SUNFALSE, "Error test fails", cv_mem->cv_netf);
   sunfprintf_long(outfile, fmt, SUNFALSE, "NLS step fails", cv_mem->cv_ncfn);
+  sunfprintf_long(outfile, fmt, SUNFALSE, "Constraint fails",
+                  cv_mem->constraint_fails);
+  sunfprintf_long(outfile, fmt, SUNFALSE, "Constraint corrections",
+                  cv_mem->constraint_corrections);
```
drreynolds, line 2619: As with CVODE, do you really want to print these by default?

src/ida/ida_io.c
```diff
@@ -1426,10 +1488,15 @@ int IDAPrintAllStats(void* ida_mem, FILE* outfile, SUNOutputFormat fmt)
...
 
+  /* step and method stats */
   sunfprintf_real(outfile, fmt, SUNTRUE, "Current time", IDA_mem->ida_tn);
   sunfprintf_long(outfile, fmt, SUNFALSE, "Steps", IDA_mem->ida_nst);
   sunfprintf_long(outfile, fmt, SUNFALSE, "Error test fails", IDA_mem->ida_netf);
   sunfprintf_long(outfile, fmt, SUNFALSE, "NLS step fails", IDA_mem->ida_ncfn);
+  sunfprintf_long(outfile, fmt, SUNFALSE, "Constraint fails",
+                  IDA_mem->constraint_fails);
+  sunfprintf_long(outfile, fmt, SUNFALSE, "Constraint corrections",
+                  IDA_mem->constraint_corrections);
```
drreynolds, line 1499: As with CVODE, do you really want to print these by default?

src/idas/idas_io.c
```diff
@@ -2258,6 +2320,10 @@ int IDAPrintAllStats(void* ida_mem, FILE* outfile, SUNOutputFormat fmt)
   sunfprintf_long(outfile, fmt, SUNFALSE, "Steps", IDA_mem->ida_nst);
   sunfprintf_long(outfile, fmt, SUNFALSE, "Error test fails", IDA_mem->ida_netf);
   sunfprintf_long(outfile, fmt, SUNFALSE, "NLS step fails", IDA_mem->ida_ncfn);
+  sunfprintf_long(outfile, fmt, SUNFALSE, "Constraint fails",
+                  IDA_mem->constraint_fails);
+  sunfprintf_long(outfile, fmt, SUNFALSE, "Constraint corrections",
+                  IDA_mem->constraint_corrections);
```
drreynolds, line 2326: As with CVODE, do you really want to print these by default?

src/cvode/cvode.c
```diff
@@ -3170,68 +3193,128 @@ static int cvNls(CVodeMem cv_mem, int nflag)
...
+  cv_mem->constraint_fails++;
+
+  /* restore zn */
+  cvRestore(cv_mem, saved_t);
+
+  /* Check for |h| == hmin or max step attempt failures */
+  if ((SUNRabs(cv_mem->cv_h) <= cv_mem->cv_hmin * ONEPSM) ||
+      (*step_constraint_fails == cv_mem->max_constraint_fails))
   {
+    SUNLogInfo(CV_LOGGER, "end-constraint-check", "status = failed max attempts");
```
drreynolds, line 3300: Minor item: this error message could also be displayed if |h|==hmin, so couldn't this "status" be misleading?  The same comment is relevant for CVODES, IDA, and IDAS.

doc/cvodes/guide/source/Usage/SIM.rst
gardner48: No, this was an error when aligning the tables in the CVODE and CVODES versions of the docs.

doc/cvodes/guide/source/Usage/SIM.rst
gardner48: No, this was an error when aligning the tables in the CVODE and CVODES versions of the docs.

doc/idas/guide/source/Usage/SIM.rst
gardner48: No, this was an error when aligning the tables in the IDA and IDAS versions of the docs.

src/cvode/cvode_io.c
gardner48: Yes the intent was to always print these values, the thinking was if constraints where ever enabled and later disabled, the statistics from when constraint checking was enabled would still be printed.

balos1 (dismissed):

Steven-Roberts (dismissed):

drreynolds (dismissed):

PR #788, opened 2025-10-25
Logging: Update region labels

gardner48:
Create a convention for logging region labels and update the log parsing script to automatically read logging region labels so new regions do not need to be explicitly added to the logging script.


gardner48:
The main logging updates are in `tools/suntools/logs.py`. The changes to the source files and outputs are to update logging region labels to the new convention of `<begin|end>-<region-name>[-list]` (mostly adding `-list` where needed).

balos1 (approved):

PR #800, opened 2025-11-15
Maintenance: Use SUN_FORMAT specifiers

gardner48:
Replace uses of `%g` and `%e` with `SUN_FORMAT_G` and `SUN_FORMAT_E`


gardner48:
Output updates in sundials-codes/answers#69

Steven-Roberts (dismissed):

Issue #806, opened 2025-12-18
[BUG] Build fails with MPI under Win, latest release 2025


horstsoft0815:
When trying to build on Windows (after generating with CMake) within Visual Studio the build fails simply due to the following line in sundials_mpi_errors.c

#include <unistd.h>

which is only valid for unix systems.

This could for example be fixed by replacing this line by

#ifdef _WIN32
#include <io.h>
#define access _access
#else
#include <unistd.h>
#endif

balos1:
Thanks for the report. This header is unused, so I have removed it entirely in #807.

gardner48:
This is fixed in the `develop` branch and will be included in the next release

PR #807, opened 2025-12-19
Don't include unistd

balos1:
Addresses #806.  This header isnt used.

### DIFF
--- a/src/sundials/sundials_mpi_errors.c
+++ b/src/sundials/sundials_mpi_errors.c
@@ -16,7 +16,6 @@
 #include <mpi.h>
 #include <stdlib.h>
 #include <string.h>
-#include <unistd.h>
 
 #include <sundials/priv/sundials_mpi_errors_impl.h>
 #include <sundials/sundials_core.h>


drreynolds (approved):
I'm assuming everything in the CI passes

src/sundials/sundials_mpi_errors.c
```diff
@@ -16,7 +16,6 @@
 #include <mpi.h>
 #include <stdlib.h>
 #include <string.h>
-#include <unistd.h>
```
gardner48, line 19: This header is also included in:
- benchmarks/nvector/test_nvector_performance.c
- cmake/tpl/SundialsPOSIXTimers.cmake
- src/sundials/sundials_profiler_impl.h
- src/sundials/sundials_mpi_errors.c
- test/unit_tests/sunlinsol/test_sunlinsol.c
- test/unit_tests/sunmatrix/test_sunmatrix.c
- test/unit_tests/sundomeigest/test_sundomeigest.c
- test/unit_tests/nvector/test_nvector.c

We should update the windows tests to enable the units tests, benchmarks, and profiling (maybe MPI too if that's easy).

gardner48 (changes requested):

src/sundials/sundials_mpi_errors.c
balos1: In all of those other places, it is properly guarded with `SUNDIALS_HAVE_POSIX_TIMERS`.

gardner48: Ah, thanks for checking. Looking that the Windows testing configs we don't have an MPI test so that how this slipped through the CI, should add that as a to do item for a future PR.

gardner48 (approved):

PR #809, opened 2026-01-10
Update llvm used for clang-format to 17.0.6

balos1:
llvm 17.0.4 is deprecated in spack, so this updates to the final release in the 17.x.x series (which is not deprecated).


gardner48:
Or should we switch to the `clang-format` installed in the github runner image? `ubuntu-latest` should have 16.0.6, 17.0.6, and 18.1.3 installed.

balos1:
> Or should we switch to the `clang-format` installed in the github runner image? `ubuntu-latest` should have 16.0.6, 17.0.6, and 18.1.3 installed.

I dont think we want to rely on the one in the github runner image, because as the runners are updated the clang-format version may change.

gardner48:
> I dont think we want to rely on the one in the github runner image, because as the runners are updated the clang-format version may change.

I think that might be ok (as long at the results are not radially different). The runner images are not updated frequently, so this would be a way of moving to newer versions as the "latest" runner is updated.

balos1:
> I think that might be ok (as long at the results are not radially different). The runner images are not updated frequently, so this would be a way of moving to newer versions as the "latest" runner is updated.

In my experience, the clang-format output can change quite a bit from patch to patch even.

PR #811, opened 2026-01-13
 Step/stage preprocessing

drreynolds:
This PR adds stage and step preprocessing, as well as failed-step postprocessing, to ARKODE.  

While existing support for stage and step postprocessing has been useful for some advanced collaborator codes, the lack of symmetry increases challenges when applications must compute auxiliary data (e.g., solving Poisson equations for a potential field) but want do not wish to do this more than absolutely necessary.  Additionally, when a step fails our existing step postprocessing is not called, eliminating the possibility for applications to rewind auxiliary data to the previous "accepted" step values.

### DIFF
--- a/doc/arkode/guide/source/Constants.rst
+++ b/doc/arkode/guide/source/Constants.rst
@@ -470,9 +470,9 @@ contains the ARKODE output constants.
    Commented-out table rows:
 
       +-------------------------------------+------+------------------------------------------------------------+
-      | :index:`ARK_POSTPROCESS_STEP_FAIL`  | -37  | An error occurred when calling the user-provided           |
+      | :index:`ARK_POSTPROCESS_STEP_FAIL`  | -37  | An error occurred when calling a user-provided.            |
       |                                     |      | step-based :c:func:`ARKPostProcessFn` routine.             |
       +-------------------------------------+------+------------------------------------------------------------+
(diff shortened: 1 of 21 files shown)


drreynolds:
Note: this is not planned for our imminent release, so reviewing can wait a few weeks,

drreynolds:
I just overhauled the role of "preprocessing" within stages.  While stage postprocessing should always have been called when the state is updated, stage-based preprocessing should be called just prior to evaluating RHS function(s).  To this end, I also updated the name to PreProcessRHS (instead of the earlier PreProcessStage).

I also noticed that for ImEx methods in ARKStep and MRIStep, we did not follow a clear pattern for ordering calls to the explicit and implicit RHS functions when evaluated at identical inputs (even when those calls were independent).  This PR also updates the order of those calls to uniformly call the implicit RHS before the explicit, in case users want to leverage that structure.

drreynolds:
I'm closing this PR to move to a more streamlined approach, since we'll have a number of small PRs to support our FES collaborators.

To simplify reviewing, each item will be a separate small PR.  

To simplify merging, I'll create a separate "staging" branch off `develop`, and then chain the small PRs together, e.g. `staging <- PR1 <- PR2 <- PR3` etc.  When merging these we should _not_ squash the commits (to make the merges simpler).  

Once that set has been merged together, we can open a final PR for `develop <- staging`.

PR #812, opened 2026-01-14
Add suntools.table module

balos1:
Parses stats that were printed in the table format instead of csv.


tools/suntools/table.py
```diff
@@ -0,0 +1,57 @@
...
+            value = match.group(2).strip()
+
+            # Attempt to convert value to a float or int if numeric
+            try:
+                if "." in value:
+                    value = float(value)
+                else:
+                    value = int(value)
+            except ValueError:
+                pass  # Keep value as string if conversion fails
```
Steven-Roberts: Similar parsing is done at
https://github.com/llnl/sundials/blob/c8dabcea90ca8bf195474da120e4f3dd39aa711f/tools/suntools/csv.py#L22-L31
Could consider making this into a shared function that both files can use.

tools/suntools/table.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def parse_table(data):
+    """
+    Parses "table" formatted SUNDIALS stats output. I.e., data is in key-value format,
+    and this converts it into the same dictionary format as csv.py.
+    Args:
+        data (str): Multiline string containing repeated key-value tables.
+    Returns:
+        list: A dictionary, where each key is a list of the values.
```
Steven-Roberts:
```suggestion
        stats: A dictionary, where each key is a list of the values.
```
Calling the output list when it is a dict is a bit confusing

tools/suntools/table.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+
+    stats = {}
+    lines = data.splitlines()
+    for line in lines:
+        # Extract key-value pairs
+        match = re.match(r"^(.*?)\s+=\s+(.*)$", line)
```
Steven-Roberts:
```suggestion
    lines = data.splitlines()
    regexp = re.compile(r"^(.*?)\s+=\s+(.*)$")
    for line in lines:
        # Extract key-value pairs
        match = regexp.match(line)
```
Untested, but should be a bit more efficient.

Steven-Roberts (dismissed):

gardner48 (approved):

PR #813, opened 2026-01-15
Feature/stageprocessing

drreynolds:
This PR adds step preprocessing, failed-step postprocessing, and stage RHS preprocessing to ARKODE.

While existing support for stage and step postprocessing has been useful for some advanced collaborator codes, the lack of symmetry increases challenges when applications must compute auxiliary data (e.g., solving Poisson equations for a potential field) but want do not wish to do this more than absolutely necessary. Additionally, when a step fails our existing step postprocessing is not called, eliminating the possibility for applications to rewind auxiliary data to the previous "accepted" step values.

*As with previous postprocessing support, these should be considered advanced/experimental features.  As such, they are not advertised in the SUNDIALS documentation*

## Step preprocessing vs postprocessing

The fundamental difference between these is when the processing routine is called inside the ARKODE time-stepping loop.  It is unlikely that a user would use both, but this allows them to more precisely place their processing function.

## Stage RHS preprocessing vs stage postprocessing 

Stage RHS preprocessing occurs just prior to any set of RHS calls that would use an identical state (e.g., if both explicit and implicit RHS functions would be called on the same input), allowing auxiliary data (e.g., MPI exchange buffers) to be reused in both calls.

On the other hand, stage postprocessing is called just after updating the current stage solution.

Again, it is unlikely that a user would leverage both in the same application, but this allows increased precision over the timings when they are performed.

PR #814, opened 2026-01-15
Removed extraneous yout copy

drreynolds:
Removed extraneous copy of `yn` into `yout` in ONE_STEP mode for ARKODE (since `ycur` aliases `yout`, and just prior to this copy `ycur` was copied into `yn`).


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 The functions `CVodeGetUserDataB` and `IDAGetUserDataB` were added to CVODES
 and IDAS, respectively.
 
+Removed extraneous copy of output vector when using ARKODE in ONE_STEP mode.
```
Steven-Roberts:
```suggestion
Removed extraneous copy of output vector when using ARKODE in `ARK_ONE_STEP` mode.
```

doc/shared/RecentChanges.rst
```diff
@@ -8,6 +8,8 @@
 The functions ``CVodeGetUserDataB`` and ``IDAGetUserDataB`` were added to CVODES
 and IDAS, respectively.
 
+Removed extraneous copy of output vector when using ARKODE in ONE_STEP mode.
```
Steven-Roberts:
```suggestion
Removed extraneous copy of output vector when using ARKODE in ``ARK_ONE_STEP`` mode.
```

Steven-Roberts (approved):

drreynolds:
This is the second in a sequence of PRs for ARKODE to support our FES collaborators.  Please review this one after #813.  When the stack is complete, we'll do a final PR into `develop`

drreynolds:

balos1 (approved):

balos1:
Does this need to be merged after #813?

drreynolds:
> Does this need to be merged after #813?

That was the plan, so that each individual review focused only on one theme.  However, this one is so small I don't think it would pollute the review for #813, so it could be reviewed/merged at any time.

CHANGELOG.md
gardner48, line 1: Also update `doc/shared/RecentChanges.rst`

src/arkode/arkode.c
```diff
@@ -1154,8 +1154,7 @@ int ARKodeEvolve(void* arkode_mem, sunrealtype tout, N_Vector yout,
     {
       istate            = ARK_SUCCESS;
       ark_mem->tretlast = *tret = ark_mem->tcur;
-      N_VScale(ONE, ark_mem->yn, yout);
-      ark_mem->next_h = ark_mem->hprime;
+      ark_mem->next_h           = ark_mem->hprime;
```
gardner48, line 1165: If rootfinding is enabled, the call to `arkRootfind` in `arkRootCheck3` above will overwrite `ycur`. We'll need to use a different temporary vector in that function to remove the copy here. 

A similar unnecessary copy is done in `arkRootCheck3` in one step mode or when `tcur` has not passed `tout` in normal mode. If `tcur` is past `tout`, then `ycur` is overwritten*. In that case we could also use a different temporary vector (i.e., the same one used in `arkRootfind`) to avoid overwriting `ycur`. That would also allow us to remove some of the other copies above for a tstop return without interpolation or a normal mode return without interpolation.

In each case where the copy is removed it would be helpful to add a comment for why the copy is not needed e.g., `/* yn and ycur (an alias to yout) are the same at this point because arkCompleteStep copies the new state from ycur into yn, so we don't need to copy yn into yout */`

*The value `ycur` is overwritten with is the interpolated solution at `tout` so, if a root is not found, we could potentially replace an extra interpolation call with a copy if we use different temporary vectors in `arkRootCheck3` and `arkRootfind`.

gardner48 (changes requested):

src/arkode/arkode.c
drreynolds: I realize that you were staring at this a couple of weeks ago, and I'm only now catching up, so I apologize in advance for simple questions.  I'll write my interpretation of your concern and proposed fix here; let me know if I have it correct.

When merely determining **if** roots occurred that should be found (in the top of `arkRootCheck2`, top of `arkRootCheck3`, and throughout `arkRootfind`), we overwrite `ycur`; if we could perform these checks with a different temporary vector, that would leave `ycur` unchanged when roots are not encountered.

However, if a root **is** encountered, then we must fill `ycur` with the interpolated solution at the root before `ARKodeEvolve` returns.  Thus the calls to `ARKodeGetDky` inside `arkRootCheck2` and at the end of `arkRootCheck3`, and the call to `N_VLinearSum` in `arkRootCheck1`, should still fill `ycur`.


Now, you also provided suggestions for other optimizations related to this PR.  I'll need to stare at things for a while to see those (allowing us to remove other copies and add comments).  However, I wonder if a two-phase approach would be reasonable here, since I expect that these other optimizations are specific to the case when rootfinding is enabled, and thus could be deferred to a later PR.

gardner48: > When merely determining **if** roots occurred that should be found (in the top of `arkRootCheck2`, top of `arkRootCheck3`, and throughout `arkRootfind`), we overwrite `ycur`; if we could perform these checks with a different temporary vector, that would leave `ycur` unchanged when roots are not encountered.
 
Yes, `arkRootfind` and `arkRootCheck3` overwrite `ycur` when searching for a root so if a root is not found we will need to copy `yn` to `ycur`. If we can use a different temporary vector in those functions, then the copy can be removed. I don't think we need to worry about `arkRootCheck2` since it is only called before the internal step loop.

> However, if a root **is** encountered, then we must fill `ycur` with the interpolated solution at the root before `ARKodeEvolve` returns. Thus the calls to `ARKodeGetDky` inside `arkRootCheck2` and at the end of `arkRootCheck3`, and the call to `N_VLinearSum` in `arkRootCheck1`, should still fill `ycur`.

Yes, I think the call to `ARKodeGetDky` to fill `ycur` at the end of `arkRootCheck3` is all that's needed and we don't need to worry about `arkRootCheck1` and `arkRootCheck2` since these only happen at the initial setup or before the internal step loop, respectively. 

> Now, you also provided suggestions for other optimizations related to this PR. I'll need to stare at things for a while to see those (allowing us to remove other copies and add comments). However, I wonder if a two-phase approach would be reasonable here, since I expect that these other optimizations are specific to the case when rootfinding is enabled, and thus could be deferred to a later PR.

Agreed, these other optimizations are specific to when rootfinding is enabled and can be left to a follow-on PR.

drreynolds: Thanks.  I just made those changes in `arkode_root.c`.  I went ahead and updated `arkRootCheck2` so that the entire file is consistent, in that nothing touches `ycur` unless a root is found; in all other cases they use `tempv4` when assessing the existence of roots.

drreynolds:
I have addressed all PR comments. I couldn't mark some as "resolved" because they were adjacent to other code modifications, and thus GitHub marked them as "outdated."

Once this passes the CI then I think it should be good to merge into the staging branch.

CHANGELOG.md
```diff
@@ -53,6 +55,8 @@ failing the step.
 The functions `CVodeGetUserDataB` and `IDAGetUserDataB` were added to CVODES and
 IDAS, respectively.
 
+Removed extraneous copy of output vector when using ARKODE in `ARK_ONE_STEP` mode.
+
```
gardner48: Looks like a bad merge, this is duplicated above
```suggestion
```

src/arkode/arkode.c
gardner48:
```suggestion
    /* In ONE_STEP mode, exit loop (arkCompleteStep already copied yn to ycur, an alias to yout) */ 
```

gardner48 (changes requested):
Two minor edits, otherwise this looks good

gardner48 (approved):

gardner48 (approved):

gardner48 (approved):

gardner48 (approved):

gardner48 (approved):

PR #815, opened 2026-01-16
LSRKStep updates

drreynolds:
General LSRKStep cleanup, including always using (tcur,ycur) as the current stage solution, and to only call RHS on that vector.


drreynolds:
This is the third in a sequence of PRs for ARKODE to support our FES collaborators.  Please review this one after #814.  When the stack is complete, we'll do a final PR into `develop`

drreynolds:

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 ### New Features and Enhancements
 
+  The default numbers of stages for the SSP Runge--Kutta methods `ARKODE_LSRK_SSP_S_2`
```
gardner48: Use single dash in markdown 
```suggestion
  The default number of stages for the SSP Runge-Kutta methods `ARKODE_LSRK_SSP_S_2`
```

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
...
+  and `ARKODE_LSRK_SSP_S_3` in LSRKStep were changed from 10 and 9, respectively, to
+  their minimum allowable values of 2 and 4.  Users may revert to the previous values
```
```suggestion
  their minimum allowable values of 2 and 4. Users may revert to the previous values
```

doc/shared/RecentChanges.rst
```diff
@@ -5,6 +5,11 @@
 
 **New Features and Enhancements**
 
+  The default numbers of stages for the SSP Runge--Kutta methods :c:enumerator:`ARKODE_LSRK_SSP_S_2`
```
```suggestion
  The default number of stages for the SSP Runge--Kutta methods :c:enumerator:`ARKODE_LSRK_SSP_S_2`
```

doc/shared/RecentChanges.rst
```diff
@@ -5,6 +5,11 @@
...
+  and :c:enumerator:`ARKODE_LSRK_SSP_S_3` in LSRKStep were changed from 10 and 9, respectively, to
+  their minimum allowable values of 2 and 4.  Users may revert to the previous values by calling
```
```suggestion
  their minimum allowable values of 2 and 4. Users may revert to the previous values by calling
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -537,6 +537,10 @@ int lsrkStep_FullRHS(ARKodeMem ark_mem, sunrealtype t, N_Vector y, N_Vector f,
   weighted local error if an embedding is present; otherwise it
   should be 0.
 
+  The variables (ark_mem->tcur, ark_mem->ycur) should
+  contain the current time and solution at each stage of within
```
gardner48:
```suggestion
  contain the current time and solution at each stage within
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -625,14 +629,14 @@ int lsrkStep_TakeStepRKC(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
...
-  /* Compute RHS function, if necessary. */
+  /* Compute RHS function for the start of the step, if necessary. */
   if ((!ark_mem->fn_is_current && ark_mem->initsetup) ||
       (step_mem->step_nst != ark_mem->nst))
   {
     /* apply user-supplied stage preprocessing function (if supplied) */
     if (ark_mem->PreProcessRHS != NULL)
     {
-      retval = ark_mem->PreProcessRHS(ark_mem->tn, ark_mem->yn,
+      retval = ark_mem->PreProcessRHS(ark_mem->tcur, ark_mem->ycur,
```
gardner48: This change (and the update to the `fe` call below) relies on the copy of `yn` into `ycur` at the start of each step attempt added in #813. As noted in that PR, this copy is potentially redundant and could be avoided by passing `yn` to the preprocessing function.

src/arkode/arkode_lsrkstep.c
```diff
@@ -731,12 +726,11 @@ int lsrkStep_TakeStepRKC(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
       }
     }
 
-    /* Use the ycur array for temporary storage here */
-    retval = step_mem->fe(ark_mem->tcur + ark_mem->h * thjm1, ark_mem->tempv2,
-                          ark_mem->ycur, ark_mem->user_data);
+    retval = step_mem->fe(ark_mem->tcur, ark_mem->ycur, ark_mem->tempv3,
```
gardner48: Looks like this function did not use `tempv3` before so this adds an extra storage requirement to `TakeStepRKC`. The vector `tempv3` is also used in `DQJtimes`, but that could be replaced with a different temporary vector and in calls to a user-supplied dominant Eigenvalue function (where it may be unused and we could remove it in the future).

src/arkode/arkode_lsrkstep.c
```diff
@@ -994,14 +975,14 @@ int lsrkStep_TakeStepRKL(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
...
-  /* Compute RHS function, if necessary. */
+  /* Compute RHS function for the start of the step, if necessary. */
   if ((!ark_mem->fn_is_current && ark_mem->initsetup) ||
       (step_mem->step_nst != ark_mem->nst))
   {
     /* apply user-supplied stage preprocessing function (if supplied) */
     if (ark_mem->PreProcessRHS != NULL)
     {
-      retval = ark_mem->PreProcessRHS(ark_mem->tn, ark_mem->yn,
+      retval = ark_mem->PreProcessRHS(ark_mem->tcur, ark_mem->ycur,
```
gardner48: Like above, this relies on a potentially unnecessary copy of `yn` to `ycur`

src/arkode/arkode_lsrkstep.c
```diff
@@ -1083,11 +1057,11 @@ int lsrkStep_TakeStepRKL(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
     }
 
     /* Use the ycur array for temporary storage here */
-    retval = step_mem->fe(ark_mem->tcur + ark_mem->h * cjm1, ark_mem->tempv2,
-                          ark_mem->ycur, ark_mem->user_data);
+    retval = step_mem->fe(ark_mem->tcur, ark_mem->ycur, ark_mem->tempv3,
```
gardner48: Like with RKC, this function did not use `tempv3` before, so this adds an extra vector storage requirement.

src/arkode/arkode_lsrkstep.c
```diff
@@ -1164,12 +1149,11 @@ int lsrkStep_TakeStepRKL(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
       return ARK_PREPROCESS_RHS_FAIL;
     }
   }
-
-  retval = step_mem->fe(ark_mem->tcur + ark_mem->h, ark_mem->ycur,
-                        ark_mem->tempv2, ark_mem->user_data);
+  retval = step_mem->fe(ark_mem->tcur, ark_mem->ycur, ark_mem->tempv3,
```
gardner48: Originally noted in [this comment](https://github.com/llnl/sundials/pull/813/changes/BASE..66985faa5999923c84f9d5b40e2d4b09c2e24888#r2757604198) in #813, it looks like this evaluation is unnecessary when using fixed step sizes.

src/arkode/arkode_lsrkstep.c
```diff
@@ -1243,37 +1230,42 @@ int lsrkStep_TakeStepSSPs2(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
...
+    bt3 = ark_mem->h * (ONE - SUN_RCONST(0.694021459207626));
   }
   else
   {
-    bt1 = (rs + ONE) / (rs * rs);
-    bt2 = ONE / rs;
-    bt3 = (rs - ONE) / (rs * rs);
+    bt1 = ark_mem->h * (rs + ONE) / (rs * rs);
+    bt2 = hrsinv;
+    bt3 = ark_mem->h * (rs - ONE) / (rs * rs);
```
gardner48: Maybe rename these to `hbt*` to indicate the multiplication by h is already incorporated

src/arkode/arkode_lsrkstep.c
```diff
@@ -1243,37 +1230,42 @@ int lsrkStep_TakeStepSSPs2(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
...
   /* The method is not FSAL. Therefore, fn ​is computed at the beginning
-     of the step unless a renewed step or ARKODE updated fn. */
+     of the step unless the previous step failed or ARKODE updated fn. */
   if (!ark_mem->fn_is_current)
   {
     /* apply user-supplied stage preprocessing function (if supplied) */
     if (ark_mem->PreProcessRHS != NULL)
     {
-      retval = ark_mem->PreProcessRHS(ark_mem->tn, ark_mem->yn,
+      retval = ark_mem->PreProcessRHS(ark_mem->tcur, ark_mem->ycur,
```
gardner48: As above, this relies on a potentially unnecessary copy of `yn` to `ycur`

src/arkode/arkode_lsrkstep.c
```diff
@@ -1510,7 +1508,7 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
...
```
gardner48: Like above, this relies on a potentially unnecessary copy of `yn` to `ycur`

src/arkode/arkode_lsrkstep.c
```diff
@@ -1741,14 +1741,17 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
     }
   }
 
+  /* Evaluate final stage group */
   for (int j = (in * (in + 1) / 2 + 1); j <= step_mem->req_stages; j++)
   {
+    /* Complete the previous stage */
+    ark_mem->tcur = ark_mem->tn + (j - in - 1) * hrat;
```
gardner48, line 1754: Before this was `(j - rn - 1) * hrat`. Looks like that was incorrect because it `rn` here might not be rounded the same as `in`. We should note the fix in the change log.

src/arkode/arkode_lsrkstep.c
```diff
@@ -1771,24 +1773,23 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
...
 
     SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
-    SUNLogInfo(ARK_LOGGER, "begin-stages-list",
-               "stage = %i, tcur = " SUN_FORMAT_G, j,
-               ark_mem->tn + ((sunrealtype)j - rn) * rat * ark_mem->h);
 
-    N_VLinearSum(ONE, ark_mem->ycur, ark_mem->h * rat, ark_mem->tempv3,
-                 ark_mem->ycur);
+    /* Begin the j-th stage by updating the state and embedding */
+    ark_mem->tcur = ark_mem->tn + (j - in) * hrat;
```
gardner48, line 1784: Similarly here, before this was `(j - rn) * hrat` and it looks like that was incorrect because it `rn` here might not be rounded the same as `in`. We should note the fix in the change log.

src/arkode/arkode_lsrkstep.c
```diff
@@ -1871,7 +1879,7 @@ int lsrkStep_TakeStepSSP43(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
     /* apply user-supplied stage preprocessing function (if supplied) */
     if (ark_mem->PreProcessRHS != NULL)
     {
-      retval = ark_mem->PreProcessRHS(ark_mem->tn, ark_mem->yn,
+      retval = ark_mem->PreProcessRHS(ark_mem->tcur, ark_mem->ycur,
```
gardner48: Like above, this relies on a potentially unnecessary copy of `yn` to `ycur`

src/arkode/arkode_lsrkstep.c
```diff
@@ -1881,35 +1889,37 @@ int lsrkStep_TakeStepSSP43(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
       }
     }
 
-    retval = step_mem->fe(ark_mem->tn, ark_mem->yn, ark_mem->fn,
+    retval = step_mem->fe(ark_mem->tcur, ark_mem->ycur, ark_mem->fn,
                           ark_mem->user_data);
     step_mem->nfe++;
-    ark_mem->fn_is_current = SUNTRUE;
```
gardner48, line 1915: Checking this elsewhere, the `ARK_FULLRHS_END` case in `lsrkStep_FullRHS` also needs to move the flag update to after the return check.

src/arkode/arkode_lsrkstep.c
```diff
@@ -2054,16 +2068,17 @@ int lsrkStep_TakeStepSSP43(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
...
   SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
-  SUNLogInfo(ARK_LOGGER, "begin-stages-list", "stage = %i, tcur = " SUN_FORMAT_G,
-             step_mem->req_stages, ark_mem->tn + ark_mem->h);
 
-  N_VLinearSum(ONE, ark_mem->ycur, ark_mem->h * p5, ark_mem->tempv3,
-               ark_mem->ycur);
+  /* Compute the time step solution and embedding */
+  ark_mem->tcur = ark_mem->tn + ark_mem->h;
+  SUNLogInfo(ARK_LOGGER, "begin-stages-list", "stage = %i, tcur = " SUN_FORMAT_G,
+             step_mem->req_stages, ark_mem->tcur);
```
gardner48:
```suggestion
             4, ark_mem->tcur);
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2141,7 +2165,7 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
     /* apply user-supplied stage preprocessing function (if supplied) */
     if (ark_mem->PreProcessRHS != NULL)
     {
-      retval = ark_mem->PreProcessRHS(ark_mem->tn, ark_mem->yn,
+      retval = ark_mem->PreProcessRHS(ark_mem->tcur, ark_mem->ycur,
```
gardner48: Like above, this relies on a potentially unnecessary copy of `yn` to `ycur`

src/arkode/arkode_lsrkstep.c
```diff
@@ -2121,15 +2139,21 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
...
 
   sunrealtype* cvals = step_mem->cvals;
   N_Vector* Xvecs    = step_mem->Xvecs;
 
+  /* Initialize method coefficients */
+  const sunrealtype hsixth = ark_mem->h / SIX;
+  const sunrealtype hfifth = ark_mem->h / FIVE;
+
+  /* Copy yn into tempv2 for use in later stages */
+  N_VScale(ONE, ark_mem->yn, ark_mem->tempv2);
```
gardner48: Move the copy to after the RHS evaluation below (avoids an unnecessary copy in the corner case of a recoverable error return from the RHS)

src/arkode/arkode_lsrkstep.c
```diff
@@ -2192,15 +2214,17 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
     }
   }
 
-  /* Evaluate stages j = 2,...,step_mem->req_stages */
+  /* Evaluate stages j = 2,...,4 */
```
gardner48:
```suggestion
  /* Evaluate stages j = 2,...,5 */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2223,12 +2246,13 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
...
     SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
-    SUNLogInfo(ARK_LOGGER, "begin-stages-list",
-               "stage = %i, tcur = " SUN_FORMAT_G, j,
-               ark_mem->tn + j * onesixth * ark_mem->h);
 
-    N_VLinearSum(ONE, ark_mem->ycur, onesixth * ark_mem->h, ark_mem->tempv3,
-                 ark_mem->ycur);
+    /* Begin the j-th stage by updating the state and embedding */
+    if (j == 5) { ark_mem->tcur = ark_mem->tn + 2 * hsixth; }
+    else { ark_mem->tcur = ark_mem->tn + j * hsixth; }
```
gardner48, line 2261: Before the `PostProcessStage` calls has these values reversed (i.e., the `j == 5` case used `tcur = tn + j * hsixth`), I think this version is correct. I don't see is error in the `develop` branch, so seems like this change was made in #813 when moving the `PostProcessStage` calls.

src/arkode/arkode_lsrkstep.c
```diff
@@ -2166,24 +2190,22 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
...
+    N_VLinearSum(ONE, ark_mem->yn, hfifth, ark_mem->fn, ark_mem->tempv1);
   }
 
   /* apply user-supplied stage postprocessing function (if supplied) */
   if (ark_mem->PostProcessStage != NULL)
   {
-    retval = ark_mem->PostProcessStage(ark_mem->tn + onesixth * ark_mem->h,
-                                       ark_mem->ycur, ark_mem->user_data);
+    retval = ark_mem->PostProcessStage(ark_mem->tcur, ark_mem->ycur,
+                                       ark_mem->user_data);
```
gardner48: Related to [this comment](https://github.com/llnl/sundials/pull/813/changes/BASE..66985faa5999923c84f9d5b40e2d4b09c2e24888#r2757676412) in #813, I believe the `PostProcessStage` call was done inside the loop before to avoid calling PostProcessStage at the end of the loop so the final `ycur` is not processed since it's not passed to a RHS call.

src/arkode/arkode_lsrkstep.c
```diff
@@ -2313,16 +2332,16 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
     if (retval > 0) { return RHSFUNC_RECVR; }
 
     SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
-    SUNLogInfo(ARK_LOGGER, "begin-stages-list",
-               "stage = %i, tcur = " SUN_FORMAT_G, j + 1,
-               ark_mem->tn + (j - 3) * onesixth * ark_mem->h);
 
-    N_VLinearSum(ONE, ark_mem->ycur, onesixth * ark_mem->h, ark_mem->tempv3,
-                 ark_mem->ycur);
+    /* Begin the j-th stage by updating the state and embedding */
```
gardner48, line 2333:
```suggestion
    /* Begin the j+1-th stage by updating the state and embedding */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2630,7 +2651,7 @@ void lsrkStep_DomEigUpdateLogic(ARKodeMem ark_mem, ARKodeLSRKStepMem step_mem,
 {
   if (dsm <= ONE)
   {
-    N_VScale(ONE, ark_mem->tempv2, ark_mem->fn);
+    N_VScale(ONE, ark_mem->tempv3, ark_mem->fn);
```
gardner48: Make the vector storing `fn` an input to this function so we don't have to keep track of which temp vector holds the evaluation. Also, this relies on the changes to RKC and RKL to use `tempv3` which adds an extra storage vector to those take step functions.

src/arkode/arkode_lsrkstep.c
```diff
@@ -2773,8 +2794,8 @@ int lsrkStep_DQJtimes(void* arkode_mem, N_Vector v, N_Vector Jv)
                                         &step_mem);
   if (retval != ARK_SUCCESS) { return retval; }
 
-  sunrealtype t = ark_mem->tn;
-  N_Vector y    = ark_mem->yn;
+  sunrealtype t = ark_mem->tcur;
+  N_Vector y    = ark_mem->ycur;
```
gardner48: This also relies on the potentially unnecessary copy of `yn` to `ycur`

test/unit_tests/logging/test_logging_arkode_lsrkstep.cpp
```diff
@@ -83,6 +83,7 @@ int main(int argc, char* argv[])
   else if (method == 3)
   {
     flag = LSRKStepSetSSPMethodByName(arkode_mem, "ARKODE_LSRK_SSP_S_3");
+    if (flag == 0) { flag = LSRKStepSetNumSSPStages(arkode_mem, 9); }
```
gardner48, line 87: Should the `method == 2` case above similar update the number of stages to the old default?

gardner48 (changes requested):

src/arkode/arkode_lsrkstep.c
drreynolds: I'm not sure that I follow.  ARKODE always allocates `tempv3`, and so this should be considered as "available" for steppers to leverage.  

It sounds like you're concerned about a future effort to remove allocation of some temporary vectors.  Thus, this line (and the corresponding one for RKL) would require `tempv3` be allocated, whereas with the previous implementation it could be pruned.  I'm fine with this tradeoff, since this PR does not increase *current* memory usage, and the previous implementation inconsistently jumped between using `ycur`. to hold the current state and the RHS of the current state.

src/arkode/arkode_lsrkstep.c
drreynolds: I believe @maggul needed this because RKL is FSAL, and by including it here he can skip evaluation of the RHS at the start of the next step.

src/arkode/arkode_lsrkstep.c
gardner48: Yes, I am thinking of when we can prune down the temporary vectors. When that does happen, we would have to revet the change to use `tempv3` back to the current version that uses `ycur` like any other temporary vector that stores the state or rhs as needed to minimize the storage. Is there a problem with using `ycur` in this way?

src/arkode/arkode_lsrkstep.c
gardner48: Sounds like it's FASL only when the embedding is used and this RHS could be unused in some cases with fixed steps e.g., using in operator splitting scheme or other cases where a `ReSet` could be performed between steps.

src/arkode/arkode_lsrkstep.c
drreynolds: Yes, @maggul noticed the error when going through this and notified me of the issue offline, so it's fixed here.

src/arkode/arkode_lsrkstep.c
drreynolds: I'd like to leave this as-is for now, since it boils down to how one numbers stages and decides precisely when each stage begins/ends.  In PR #819 I add a "get" routine to retrieve the current stage index, so let's defer adjusting this internal comment until we've hashed out our definition of the stage indices and dividing line.

src/arkode/arkode_lsrkstep.c
drreynolds: The problem with the previous implementation was that it ended up calling the user RHS routines on temporary vectors instead of consistently calling the RHS on `ycur`.  The next PR in this stack, #817, makes this consistent across nearly all of ARKODE, so that users who store auxiliary data in a custom N_Vector that they pass to `ARKodeEvolve` will always get that data back when evaluating the RHS (e.g., external potential fields), instead of getting other vectors that may have incorrect auxiliary data.

I'm fine with adjusting how we store temporaries in RKC and RKL, but I did not see a way to accomplish my `ycur` promise without leveraging a temporary vector, though I can stare at this again if desired.  However, if it turns out that the price to consistently use `ycur` in RHS routines is that RKC and RKL require a temporary vector that a few other methods can someday omit, I think it's worth it.

gardner48: I don't think we should guarantee that a user will always get `ycur` in a RHS callbacks. I'm not sure this can be done in all cases so the convention would be inconsistent and could easily be broken leading to hard to track down bugs. Is there a reason this auxiliary data can't leverage the user data pointer?

gardner48: It also seems like requiring using `ycur` could force us to make choices that might be undesirable for other applications like performing extra copies or increasing the memory footprint.

src/arkode/arkode_lsrkstep.c
drreynolds: I agree.  I'll revert the placement of this call.

src/arkode/arkode_lsrkstep.c
drreynolds: I think with fixed steps and when `ARKodeReset` will be called, then this RHS evaluation would be extraneous; however I don't it's possible to know here whether `ARKodeReset` will be called later.

I suppose we could re-engineer how this is currently handled, including rebuilding `lsrkStep_DomEigUpdateLogic` and `lsrkStep_FullRHS`, to improve performance in the case where `ARKodeReset` is called frequently (perhaps emulating a bit of how ERKStep handles FSAL methods).  

Thus, this "lsrk-cleanup" PR doesn't perfect everything from the original implementation.  Whether this optimization is required in this PR, or deferred for a separate effort, is a separate question.

src/arkode/arkode_lsrkstep.c
drreynolds: The other steppers did not require anything significant to accomplish this, and places that would have required a larger memory footprint (e.g., SPRKStep, and DQ Jacobian approximation routines) were left as exceptions to this rule (which I believe I document in that PR).  The big anomaly was RKC and RKL, which is why I updated it here.

I agree that the user data pointer is a natural place for auxiliary data.  I'd need to check with our collaborators as to why that was a sub-optimal option (I believe for both Gkeyll and GENE-X).  Meanwhile, I'll stare at this to see if I can use only `tempv1`, `tempv2` and `ycur` as before, but consistently call the RHS on `ycur`.

src/arkode/arkode_lsrkstep.c
```diff
@@ -1243,37 +1231,42 @@ int lsrkStep_TakeStepSSPs2(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
...
   if (step_mem->req_stages == 2)
   {
-    bt1 = SUN_RCONST(
-      0.694021459207626); // due to https://doi.org/10.1016/j.cam.2022.114325 pg 5
-    bt2 = ZERO;
-    bt3 = ONE - bt1;
+    // from https://doi.org/10.1016/j.cam.2022.114325 pg 5
+    hbt1 = ark_mem->h * SUN_RCONST(0.694021459207626);
+    hbt2 = ZERO;
+    hbt3 = ark_mem->h * (ONE - SUN_RCONST(0.694021459207626));
```
maggul:
```suggestion
    hbt3 = ark_mem->h - hbt1;
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1243,37 +1231,42 @@ int lsrkStep_TakeStepSSPs2(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
...
   }
   else
   {
-    bt1 = (rs + ONE) / (rs * rs);
-    bt2 = ONE / rs;
-    bt3 = (rs - ONE) / (rs * rs);
+    hbt1 = ark_mem->h * (rs + ONE) / (rs * rs);
+    hbt2 = hrsinv;
+    hbt3 = ark_mem->h * (rs - ONE) / (rs * rs);
```
maggul:
```suggestion
    hbt1 = hrsinv * (ONE + rsinv);
    hbt2 = hrsinv;
    hbt3 = hrsinv * (ONE - rsinv);
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1494,11 +1489,15 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
...
-  sunrealtype rs  = (sunrealtype)step_mem->req_stages;
-  sunrealtype rn  = SUNRsqrt(rs);
-  sunrealtype rat = ONE / (rs - rn);
-  int in          = (int)SUNRround(rn);
+  /* Initialize method coefficients */
+  const sunrealtype rs     = (sunrealtype)step_mem->req_stages;
+  const sunrealtype rn     = SUNRsqrt(rs);
+  const sunrealtype hrat   = ark_mem->h / (rs - rn);
+  const sunrealtype rsinv  = ONE / rs;
+  const sunrealtype hrsinv = ark_mem->h * rsinv;
```
maggul:
```suggestion
  const sunrealtype hrsinv = ark_mem->h / rs;
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1857,9 +1862,13 @@ int lsrkStep_TakeStepSSP43(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
   sunrealtype* cvals = step_mem->cvals;
   N_Vector* Xvecs    = step_mem->Xvecs;
 
-  sunrealtype rs = SUN_RCONST(4.0);
-  sunrealtype p5 = SUN_RCONST(0.5);
+  /* Initialize method coefficients */
+  const sunrealtype rs     = SUN_RCONST(4.0);
+  const sunrealtype hp5    = ark_mem->h * SUN_RCONST(0.5);
+  const sunrealtype rsinv  = ONE / rs;
+  const sunrealtype hrsinv = ark_mem->h * rsinv;
```
maggul:
```suggestion
  const sunrealtype hrsinv = ark_mem->h / rs;
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2300,9 +2308,8 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
...
     }
 
-    retval = step_mem->fe(ark_mem->tcur +
-                            ((sunrealtype)j - FOUR) * onesixth * ark_mem->h,
-                          ark_mem->ycur, ark_mem->tempv3, ark_mem->user_data);
+    retval = step_mem->fe(ark_mem->tcur, ark_mem->ycur, ark_mem->tempv3,
+                          ark_mem->user_data);
     step_mem->nfe++;
 
     SUNLogExtraDebugVec(ARK_LOGGER, "stage RHS", ark_mem->tempv3, "F_%i(:) =", j);
```
maggul:
```suggestion
    SUNLogExtraDebugVec(ARK_LOGGER, "stage RHS", ark_mem->tempv3, "F_%i(:) =", j - 1);
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2359,8 +2368,8 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
     }
   }
 
-  retval = step_mem->fe(ark_mem->tcur + ark_mem->h, ark_mem->ycur,
-                        ark_mem->tempv3, ark_mem->user_data);
+  retval = step_mem->fe(ark_mem->tcur, ark_mem->ycur, ark_mem->tempv3,
+                        ark_mem->user_data);
   step_mem->nfe++;
 
   SUNLogExtraDebugVec(ARK_LOGGER, "stage RHS", ark_mem->tempv3, "F_10(:) =", 9);
```
maggul:
```suggestion
  SUNLogExtraDebugVec(ARK_LOGGER, "stage RHS", ark_mem->tempv3, "F_9(:) =", 9);
```

maggul (approved):

src/arkode/arkode_lsrkstep.c
```diff
@@ -537,6 +537,10 @@ int lsrkStep_FullRHS(ARKodeMem ark_mem, sunrealtype t, N_Vector y, N_Vector f,
   weighted local error if an embedding is present; otherwise it
   should be 0.
 
+  The variables (ark_mem->tcur, ark_mem->ycur) should
+  contain the current time and solution at each stage within
+  the time step.
```
gardner48: Removing the copy in #813 to update `ycur` with `yn` will require some updates below to revert `ycur` back to `yn` (looks like just the first RHS call at the start of a step)

src/arkode/arkode_lsrkstep.c
gardner48: Noted optimization when using fixed steps in https://github.com/llnl/sundials-private/issues/144

gardner48 (changes requested):
Other than a side effect of removing the `yn` to `ycur` copy in #813, this looks good to go

src/arkode/arkode_lsrkstep.c
```diff
@@ -625,14 +630,14 @@ int lsrkStep_TakeStepRKC(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
...
-  /* Compute RHS function, if necessary. */
+  /* Compute RHS function for the start of the step, if necessary. */
   if ((!ark_mem->fn_is_current && ark_mem->initsetup) ||
       (step_mem->step_nst != ark_mem->nst))
   {
     /* call the user-supplied pre-rhs function (if supplied) */
     if (ark_mem->PreRhsFn)
     {
-      retval = ark_mem->PreRhsFn(ark_mem->tn, ark_mem->yn, ark_mem->user_data);
+      retval = ark_mem->PreRhsFn(ark_mem->tcur, ark_mem->yn, ark_mem->user_data);
```
gardner48: Minor suggestion to use `tn` when using `yn`, I'll try to suggest this change anywhere else it shows up.

```suggestion
      retval = ark_mem->PreRhsFn(ark_mem->tn, ark_mem->yn, ark_mem->user_data);
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -642,7 +647,7 @@ int lsrkStep_TakeStepRKC(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
     }
 
     /* call fe */
-    retval = step_mem->fe(ark_mem->tn, ark_mem->yn, ark_mem->fn,
+    retval = step_mem->fe(ark_mem->tcur, ark_mem->yn, ark_mem->fn,
```
gardner48:
```suggestion
    retval = step_mem->fe(ark_mem->tn, ark_mem->yn, ark_mem->fn,
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1004,22 +981,22 @@ int lsrkStep_TakeStepRKL(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
...
-  /* Compute RHS function, if necessary. */
+  /* Compute RHS function for the start of the step, if necessary. */
   if ((!ark_mem->fn_is_current && ark_mem->initsetup) ||
       (step_mem->step_nst != ark_mem->nst))
   {
     /* call the user-supplied pre-RHS function (if supplied) */
     if (ark_mem->PreRhsFn)
     {
-      retval = ark_mem->PreRhsFn(ark_mem->tn, ark_mem->yn, ark_mem->user_data);
+      retval = ark_mem->PreRhsFn(ark_mem->tcur, ark_mem->yn, ark_mem->user_data);
```
gardner48:
```suggestion
      retval = ark_mem->PreRhsFn(ark_mem->tn, ark_mem->yn, ark_mem->user_data);
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1004,22 +981,22 @@ int lsrkStep_TakeStepRKL(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
...
       if (retval != 0)
       {
         SUNLogInfo(ARK_LOGGER, "end-stages-list",
                    "status = failed preprocess rhs, retval = %i", retval);
         return ARK_PRERHSFN_FAIL;
       }
     }
-    retval = step_mem->fe(ark_mem->tn, ark_mem->yn, ark_mem->fn,
+    retval = step_mem->fe(ark_mem->tcur, ark_mem->yn, ark_mem->fn,
```
gardner48:
```suggestion
    retval = step_mem->fe(ark_mem->tn, ark_mem->yn, ark_mem->fn,
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1263,37 +1242,42 @@ int lsrkStep_TakeStepSSPs2(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
...
   /* The method is not FSAL. Therefore, fn ​is computed at the beginning
-     of the step unless a renewed step or ARKODE updated fn. */
+     of the step unless the previous step failed or ARKODE updated fn. */
   if (!ark_mem->fn_is_current)
   {
     /* call the user-supplied pre-RHS function (if supplied) */
     if (ark_mem->PreRhsFn)
     {
-      retval = ark_mem->PreRhsFn(ark_mem->tn, ark_mem->yn, ark_mem->user_data);
+      retval = ark_mem->PreRhsFn(ark_mem->tcur, ark_mem->yn, ark_mem->user_data);
```
gardner48:
```suggestion
      retval = ark_mem->PreRhsFn(ark_mem->tn, ark_mem->yn, ark_mem->user_data);
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1302,12 +1286,11 @@ int lsrkStep_TakeStepSSPs2(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
       }
     }
 
-    retval = step_mem->fe(ark_mem->tn, ark_mem->yn, ark_mem->fn,
+    retval = step_mem->fe(ark_mem->tcur, ark_mem->yn, ark_mem->fn,
```
gardner48:
```suggestion
    retval = step_mem->fe(ark_mem->tn, ark_mem->yn, ark_mem->fn,
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1530,7 +1516,7 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
     /* call the user-supplied pre-RHS function (if supplied) */
     if (ark_mem->PreRhsFn)
     {
-      retval = ark_mem->PreRhsFn(ark_mem->tn, ark_mem->yn, ark_mem->user_data);
+      retval = ark_mem->PreRhsFn(ark_mem->tcur, ark_mem->yn, ark_mem->user_data);
```
gardner48:
```suggestion
      retval = ark_mem->PreRhsFn(ark_mem->tn, ark_mem->yn, ark_mem->user_data);
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1539,7 +1525,7 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
       }
     }
 
-    retval = step_mem->fe(ark_mem->tn, ark_mem->yn, ark_mem->fn,
+    retval = step_mem->fe(ark_mem->tcur, ark_mem->yn, ark_mem->fn,
```
gardner48:
```suggestion
    retval = step_mem->fe(ark_mem->tn, ark_mem->yn, ark_mem->fn,
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1900,7 +1894,7 @@ int lsrkStep_TakeStepSSP43(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
     /* call the user-supplied pre-RHS function (if supplied) */
     if (ark_mem->PreRhsFn)
     {
-      retval = ark_mem->PreRhsFn(ark_mem->tn, ark_mem->yn, ark_mem->user_data);
+      retval = ark_mem->PreRhsFn(ark_mem->tcur, ark_mem->yn, ark_mem->user_data);
```
gardner48:
```suggestion
      retval = ark_mem->PreRhsFn(ark_mem->tn, ark_mem->yn, ark_mem->user_data);
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1909,35 +1903,37 @@ int lsrkStep_TakeStepSSP43(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
       }
     }
 
-    retval = step_mem->fe(ark_mem->tn, ark_mem->yn, ark_mem->fn,
+    retval = step_mem->fe(ark_mem->tcur, ark_mem->yn, ark_mem->fn,
```
gardner48:
```suggestion
    retval = step_mem->fe(ark_mem->tn, ark_mem->yn, ark_mem->fn,
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2169,7 +2172,7 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
     /* call the user-supplied pre-RHS function (if supplied) */
     if (ark_mem->PreRhsFn)
     {
-      retval = ark_mem->PreRhsFn(ark_mem->tn, ark_mem->yn, ark_mem->user_data);
+      retval = ark_mem->PreRhsFn(ark_mem->tcur, ark_mem->yn, ark_mem->user_data);
```
gardner48:
```suggestion
      retval = ark_mem->PreRhsFn(ark_mem->tn, ark_mem->yn, ark_mem->user_data);
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2178,7 +2181,7 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
       }
     }
 
-    retval = step_mem->fe(ark_mem->tn, ark_mem->yn, ark_mem->fn,
+    retval = step_mem->fe(ark_mem->tcur, ark_mem->yn, ark_mem->fn,
```
gardner48:
```suggestion
    retval = step_mem->fe(ark_mem->tn, ark_mem->yn, ark_mem->fn,
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2804,7 +2790,7 @@ int lsrkStep_DQJtimes(void* arkode_mem, N_Vector v, N_Vector Jv)
                                         &step_mem);
   if (retval != ARK_SUCCESS) { return retval; }
 
-  sunrealtype t = ark_mem->tn;
+  sunrealtype t = ark_mem->tcur;
```
gardner48:
```suggestion
  sunrealtype t = ark_mem->tn;
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2725,7 +2711,7 @@ int lsrkStep_ComputeNewDomEig(ARKodeMem ark_mem, ARKodeLSRKStepMem step_mem)
   }
   else if (step_mem->dom_eig_fn != NULL)
   {
-    retval = step_mem->dom_eig_fn(ark_mem->tn, ark_mem->ycur, ark_mem->fn,
+    retval = step_mem->dom_eig_fn(ark_mem->tn, ark_mem->yn, ark_mem->fn,
```
gardner48, line 2714: Looks like this was a bug before unless the output vector passed to `ARKodeEvolve` had the initial condition on the first call or the last returned solution on a subsequent calls (if an estimate was requested on the first step in the `ARKodeEvolve` call. 

I'll suggest an addition to the recent changes.

doc/shared/RecentChanges.rst
```diff
@@ -42,6 +47,9 @@ Fixed a bug in the ARKODE discrete adjoint checkpointing where an incorrect
 state would be stored on the first step if the output vector passed to
 :c:func:`ARKodeEvolve` did not contain the initial condition on the first call.
 
+Fixed a potential bug in LSRKStep's :c:enumerator:`ARKODE_LSRK_SSP_S_3` method, where a real
+number was used instead of an integer, potentially resulting in a rounding error.
```
gardner48, line 67:
```suggestion
number was used instead of an integer, potentially resulting in a rounding error.

Fixed a bug in LSRKStep where an incorrect state vector could be passed to a
user-supplied dominant eigenvalue function on the first step unless the output
vector passed to :c:func:`ARKodeEvolve` contained the initial condition and when
an eigenvalue estimate is requested on the first step in a subsequent call to
:c:func:`ARKodeEvolve` unless the output vector passed contained the most recently
returned solution.
```

CHANGELOG.md
```diff
@@ -43,6 +48,9 @@ Fixed a bug in the ARKODE discrete adjoint checkpointing where an incorrect
 state would be stored on the first step if the output vector passed to
 `ARKodeEvolve` did not contain the initial condition on the first call.
 
+Fixed a potential bug in LSRKStep's `ARKODE_LSRK_SSP_S_3` method, where a real
+number was used instead of an integer, potentially resulting in a rounding error.
```
gardner48, line 66:
```suggestion
number was used instead of an integer, potentially resulting in a rounding error.

Fixed a bug in LSRKStep where an incorrect state vector could be passed to a
user-supplied dominant eigenvalue function on the first step unless the output
vector passed to `ARKodeEvolve` contained the initial condition and when an
eigenvalue estimate is requested on the first step in a subsequent call to
`ARKodeEvolve` unless the output vector passed contained the most recently returned
solution.
```

gardner48 (changes requested):
Minor suggestion to use `tn` with `yn` and note a bugfix in the change log. Otherwise this looks good to me.

gardner48 (approved):

PR #816, opened 2026-01-17
CI: Chain workflows

gardner48:
Run CI in stages (formatting checks, interface checks, doc checks, and tests) and only run a stage if there are changes in the relevant directories and the prior stage was successful (or skipped).


drreynolds (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,183 @@
+name: ci
+
+# Workflow triggers
+on:
```
balos1, line 4: Do we still want these to run on push to main or develop?

gardner48: We could but it seemed redundant since the same tests just ran for the PR

balos1 (approved):

PR #817, opened 2026-01-17
Consistently use `ycur` for user RHS routines

drreynolds:
This PR has been modified from its original implementation.  It now includes a set of relatively minor changes so that steppers leverage the fact that `ycur` has already been initialized with `yn` at the ARKODE level.


drreynolds:
This is the fourth in a sequence of PRs for ARKODE to support our FES collaborators.  Please review this one after #815.  When the stack is complete, we'll do a final PR into `develop`

drreynolds:

gardner48 (commented):
After removing the `yn` to `ycur` copy in #813, I think we can drop this PR from the series and keep the current calls using `yn`

PR #818, opened 2026-01-17
Optional allocation of ARKODE internal data prior to ARKodeEvolve

drreynolds:
This PR adds a function `ARKodeAllocateInternalData` to enable stage-related data allocation before the first call to `ARKodeEvolve` (but after all other optional input routines have been called), to support users who measure memory usage before beginning a simulation.

PR #819, opened 2026-01-20
Feature: User-callable routine to retrieve the current stage index

drreynolds:
This PR adds the function `ARKodeGetStageIndex` that returns the index of the stage currently being processed, and the total number of stages in the method, for users who must compute auxiliary quantities in their IVP right-hand side functions during some stages and not others (e.g., in all but the first or last stage).  It also adds functions to retrieve the last saved state (`yn`) and the corresponding time (`tn`).


drreynolds:
This is the sixth (and last) in a sequence of PRs for ARKODE to support our FES collaborators.  Please review this one after #818.  When the stack is complete, we'll do a final PR into `develop`

manauref:
A data point: I've tested this in combination with
```
ARKodeSetPreprocessRHSFn
ARKodeSetPostprocessStageFn
ARKodeSetPostprocessStepFailFn
```
in [Gkeyll](https://github.com/ammarhakim/gkeyll) and it seems to work fine.

doc/arkode/guide/source/Usage/User_callable.rst
```diff
@@ -4090,6 +4099,38 @@ Retrieve the accumulated temporal error estimate       :c:func:`ARKodeGetAccumul
...
+
+   Returns the last saved solution reached by the solver.
+
+   :param arkode_mem: pointer to the ARKODE memory block.
+   :param yn: last saved solution.
+
+   :retval ARK_SUCCESS: the function exited successfully.
+   :retval ARK_MEM_NULL: ``arkode_mem`` was ``NULL``.
+
+   .. note::
```
gardner48: Maybe upgrade this to a "danger" message
```suggestion
   .. danger::
```

doc/arkode/guide/source/Usage/User_supplied.rst
gardner48, line 1: Bad merge, revert to the version in #813 that uses the right style for [user-supplied functions](https://sundials.readthedocs.io/en/latest/developers/documentation/Style.html#user-supplied-functions)

doc/shared/RecentChanges.rst
```diff
@@ -34,6 +34,11 @@ stage-related data allocation before the first call to :c:func:`ARKodeEvolve`
 (but after all other optional input routines have been called), to support
 users who measure memory usage before beginning a simulation.
 
+Added the function :c:func:`ARKodeGetStageIndex` that returns the index of the
+stage currently being processed, and the total number of stages in the method, for
+users who must compute auxiliary quantities in their IVP right-hand side functions
+during some stages and not others (e.g., in all but the first or last stage).
```
gardner48, line 44: Note the addition of the functions to get the last time and state.

doc/arkode/guide/source/Usage/User_callable.rst
```diff
@@ -4463,6 +4504,21 @@ Retrieve the accumulated temporal error estimate       :c:func:`ARKodeGetAccumul
    .. versionadded:: 6.2.0
 
 
+.. c:function:: int ARKodeGetStageIndex(void* arkode_mem, int* stage, int *max_stages)
+
+   Returns the index of the current stage (0-based) and the total number of
+   stages in the method.
```
gardner48: Just to be extra clear
```suggestion
   Returns the index of the current stage (0-based) and the total number of
   stages in the method i.e., for an :math:`s`-stage method `stage`
   :math:`\in 0,\dots,s-1` and `max_stages` :math:`= s`.
```

src/arkode/arkode_erkstep.c
```diff
@@ -904,6 +909,9 @@ int erkStep_TakeStep(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
       }
     }
 
+    /* update current stage index */
+    step_mem->istage = is;
```
gardner48: Like in ARKStep, this should be set at the start of the loop. Currently, if a user calls `ARKodeGetStageIndex` in `PostProcessStageFn` they will get `0` for the stage index when `is = 1` indicating `ycur` is the first stage state but `ycur` has already been updated and holds the second stage state.

src/arkode/arkode_io.c
```diff
@@ -3163,6 +3203,38 @@ int ARKodeGetUserData(void* arkode_mem, void** user_data)
   return (ARK_SUCCESS);
 }
 
+/*---------------------------------------------------------------
+  ARKodeGetStageIndex:
+
+  Returns the index of the current stage and the total number of
+  stages.  If this is not supplied by the time-stepping module
+  then it returns (0,1), indicating that it is currently in the
```
gardner48: I think we should return an error and maybe negative output values to indicate the stepper does not supply this function so users don't trust the values in that case.

src/arkode/arkode_lsrkstep.c
```diff
@@ -709,6 +714,9 @@ int lsrkStep_TakeStepRKC(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
   /* Evaluate stages j = 2,...,step_mem->req_stages */
   for (int j = 2; j <= step_mem->req_stages; j++)
   {
+    /* set stage index (0-based) */
+    step_mem->istage = j - 1;
```
gardner48: Because of how the computations are ordered with `tmp2` computed and `PostProcessStageFn` called before the loop, `istage = 1` needs to be added above before computing `tmp2`. Currently when `PostProcessStageFn` is first called `istage = 0` however, the input state (`tmp2`) corresponds to `z_1` not `z_0 = y_n`. Similarly, updating `istage` in the loop needs to move to after the RHS evaluation below (when updating `tcur`) and be changed to `step_mem->istage = j`. I think the "begin/end-stages-list" logging statements have the correct stage indexing.

Also, equation 2.13 for RKC/RKL methods in the mathematical consideration section of the docs should probably be updated to 1-based indexing so we're consistent across the methods.

src/arkode/arkode_lsrkstep.c
```diff
@@ -821,7 +829,8 @@ int lsrkStep_TakeStepRKC(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
   SUNLogInfo(ARK_LOGGER, "begin-compute-embedding", "");
 
   /* final stage processing */
-  ark_mem->tcur = ark_mem->tn + ark_mem->h;
+  step_mem->istage = step_mem->req_stages;
```
gardner48: Following the comment above to relocate the `istage` update within the loop, this can be removed as it will already be `req_stages` at this point.
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1044,6 +1057,9 @@ int lsrkStep_TakeStepRKL(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
   /* Evaluate stages j = 2,...,step_mem->req_stages */
   for (int j = 2; j <= step_mem->req_stages; j++)
   {
+    /* set stage index (0-based) */
+    step_mem->istage = j - 1;
```
gardner48: Same comments as for RKC:
* Add `istage = 1` when computing `z_1 = tmp2`
* Change this line to `istage = j` and move it to after the RHS evaluation when updating `tcur`

src/arkode/arkode_lsrkstep.c
```diff
@@ -1143,10 +1159,11 @@ int lsrkStep_TakeStepRKL(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
 
   SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
   SUNLogExtraDebugVec(ARK_LOGGER, "updated solution", ark_mem->ycur, "ycur(:) =");
+  SUNLogInfo(ARK_LOGGER, "begin-compute-embedding", "");
 
   /* final stage processing */
-  SUNLogInfo(ARK_LOGGER, "begin-compute-embedding", "");
-  ark_mem->tcur = ark_mem->tn + ark_mem->h;
+  step_mem->istage = step_mem->req_stages;
```
gardner48: Like in RKC this can be removed as it will be set within the loop
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -709,6 +714,9 @@ int lsrkStep_TakeStepRKC(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
   /* Evaluate stages j = 2,...,step_mem->req_stages */
   for (int j = 2; j <= step_mem->req_stages; j++)
   {
+    /* set stage index (0-based) */
+    step_mem->istage = j - 1;
+
     /* Evaluate RHS and store in ycur */
     ark_mem->tcur = ark_mem->tn + ark_mem->h * thjm1;
```
gardner48: This update to `tcur` duplicates the update to `tcur` above (or the prior iteration) and can be removed
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1044,6 +1057,9 @@ int lsrkStep_TakeStepRKL(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
   /* Evaluate stages j = 2,...,step_mem->req_stages */
   for (int j = 2; j <= step_mem->req_stages; j++)
   {
+    /* set stage index (0-based) */
+    step_mem->istage = j - 1;
+
     /* Evaluate RHS and store in ycur */
     ark_mem->tcur = ark_mem->tn + ark_mem->h * cjm1;
```
gardner48: This update to `tcur` duplicates the update to `tcur` above (or the prior iteration) and can be removed
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -821,7 +829,8 @@ int lsrkStep_TakeStepRKC(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
   SUNLogInfo(ARK_LOGGER, "begin-compute-embedding", "");
 
   /* final stage processing */
-  ark_mem->tcur = ark_mem->tn + ark_mem->h;
+  step_mem->istage = step_mem->req_stages;
+  ark_mem->tcur    = ark_mem->tn + ark_mem->h;
```
gardner48: Similarly this `tcur` update duplicates the last `tcur` update within the loop and could be removed (the last `cj` should be 1).
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1143,10 +1159,11 @@ int lsrkStep_TakeStepRKL(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
 
   SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
   SUNLogExtraDebugVec(ARK_LOGGER, "updated solution", ark_mem->ycur, "ycur(:) =");
+  SUNLogInfo(ARK_LOGGER, "begin-compute-embedding", "");
 
   /* final stage processing */
-  SUNLogInfo(ARK_LOGGER, "begin-compute-embedding", "");
-  ark_mem->tcur = ark_mem->tn + ark_mem->h;
+  step_mem->istage = step_mem->req_stages;
+  ark_mem->tcur    = ark_mem->tn + ark_mem->h;
```
gardner48: Like in RKC this duplicates the last `tcur` update from within the loop and could be removed (the last `cj` should be 1).
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1327,6 +1347,9 @@ int lsrkStep_TakeStepSSPs2(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
   /* Evaluate stages j = 2,...,step_mem->req_stages - 1 */
   for (int j = 2; j < step_mem->req_stages; j++)
   {
+    /* set stage index (0-based) */
+    step_mem->istage = j - 1;
```
gardner48: Same comments as in RKC:
* Add `istage = 1` above before computing `z_1 = ycur`
* Change this line to `istage = j` and move it to after the RHS evaluation when updating `tcur`

src/arkode/arkode_lsrkstep.c
```diff
@@ -1327,6 +1347,9 @@ int lsrkStep_TakeStepSSPs2(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
...
+
     /* Complete previous stage by evaluating RHS and storing it in tempv2 */
     ark_mem->tcur = ark_mem->tn + (j - 1) * hsm1inv;
```
gardner48: Like in RKC this duplicates the `tcur` update above (or the prior iteration) and can be removed.
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1382,7 +1405,8 @@ int lsrkStep_TakeStepSSPs2(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
   }
 
   /* Complete the next-to-last stage by evaluating the RHS and storing it in tempv2 */
-  ark_mem->tcur = ark_mem->tn + ark_mem->h;
+  step_mem->istage = step_mem->req_stages - 1;
```
gardner48: Like in RKC this can be removed as it will be set within the loop
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1382,7 +1405,8 @@ int lsrkStep_TakeStepSSPs2(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
...
+  ark_mem->tcur    = ark_mem->tn + ark_mem->h;
```
gardner48: I'm not sure this `tcur` update is correct (unless `j * hsm1inv = 1` in the loop above). This should probably be moved below before the linear combination to update `ycur` for the final solution. Similarly, `istage = req_stages` should be added below when updating `ycur`.
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1567,6 +1594,9 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
   /* Evaluate first stage group */
   for (int j = 2; j <= ((in - 1) * (in - 2) / 2); j++)
   {
+    /* set stage index (0-based) */
+    step_mem->istage = j - 1;
```
gardner48: Same comments as in RKC:
* Add `istage = 1` above before computing `z_1 = ycur`
* Change this line to `istage = j` and move it to after the RHS evaluation when updating `tcur`

src/arkode/arkode_lsrkstep.c
```diff
@@ -1567,6 +1594,9 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
...
+
     /* Complete previous stage by evaluating RHS and storing it in tempv3 */
     ark_mem->tcur = ark_mem->tn + (j - 1) * hrat;
```
gardner48: Like in RKC this duplicates the `tcur` update before the loop (or the prior iteration) and could be removed
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1628,6 +1658,9 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
   /* Evaluate second stage group */
   for (int j = ((in - 1) * (in - 2) / 2 + 1); j <= (in * (in + 1) / 2 - 1); j++)
   {
+    /* set stage index (0-based) */
+    step_mem->istage = j - 1;
```
gardner48: After moving `istage` in the prior loop its value will be correct at the start of this loop and this update should be changed to `istage` and moved after the RHS evaluation when updating `tcur`.

src/arkode/arkode_lsrkstep.c
```diff
@@ -1684,7 +1717,8 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
   }
 
   /* Complete the last stage from the second stage group */
-  ark_mem->tcur = ark_mem->tn + hrat * (rn * (rn + ONE) / TWO - ONE);
+  step_mem->istage = (in * (in + 1) / 2 - 1);
```
gardner48: `istage` should already be correct from the last iteration of the loop so this can be removed
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1684,7 +1717,8 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
...
+  ark_mem->tcur    = ark_mem->tn + hrat * (rn * (rn + ONE) / TWO - ONE);
```
gardner48: This should also already be set from the last iteration of the loop and could be removed. 
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1750,6 +1784,9 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
   /* Evaluate final stage group */
   for (int j = (in * (in + 1) / 2 + 1); j <= step_mem->req_stages; j++)
   {
+    /* set stage index (0-based) */
+    step_mem->istage = j - 1;
```
gardner48: * Add `istage = in * (in + 1) / 2` before the loop when updating `tcur` before the linear combination
* Change this to `istage = j` and move to after the RHS evaluation when updating `tcur`

src/arkode/arkode_lsrkstep.c
```diff
@@ -1750,6 +1784,9 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
...
+
     /* Complete the previous stage */
     ark_mem->tcur = ark_mem->tn + (j - in - 1) * hrat;
```
gardner48: This should be the same as the `tcur` computed before the loop (or on prior iterations) and could be removed
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1942,6 +1982,9 @@ int lsrkStep_TakeStepSSP43(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
     }
   }
 
+  /* update stage index */
+  step_mem->istage = 1;
```
gardner48: This needs to move up to when `tcur` is updated and `z_1 = ycur` otherwise `istage = 0` when calling `PostProcessStage` with `z_1`

src/arkode/arkode_lsrkstep.c
```diff
@@ -1991,6 +2034,9 @@ int lsrkStep_TakeStepSSP43(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
     }
   }
 
+  /* update stage index */
+  step_mem->istage = 2;
```
gardner48: This needs to move up to when `tcur` is updated and `z_2 = ycur` otherwise `istage = 1` when calling `PostProcessStage` with `z_2`

src/arkode/arkode_lsrkstep.c
```diff
@@ -2053,6 +2099,9 @@ int lsrkStep_TakeStepSSP43(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
     }
   }
 
+  /* update stage index */
+  step_mem->istage = 3;
```
gardner48: This needs to move up to when `tcur` is updated and `z_3 = ycur` otherwise `istage = 2` when calling `PostProcessStage` with `z_3`

src/arkode/arkode_lsrkstep.c
```diff
@@ -2053,6 +2099,9 @@ int lsrkStep_TakeStepSSP43(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
...
+
   /* Evaluate stage RHS */
```
gardner48, line 2077: Need to add `istage = 4` below when updating `tcur` for the updated solution computation

src/arkode/arkode_lsrkstep.c
```diff
@@ -2229,6 +2281,9 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
       }
     }
 
+    /* update stage index (0-based) */
+    step_mem->istage = j - 1;
```
gardner48: Since `PostProcessStage` is called inside the loop in this case, `istage = j - 1` could be moved to the start of the loop so the value is correct when calling `PostProcessStage`. However, from the changes above, it seems like a good convention is to update `istage` when updating `tcur`, so instead I think it would be better to add `istage = 1` outside the loop when updating `tcur` for `z_1` and add `istage = j` after the RHS call when updating `tcur`.

src/arkode/arkode_lsrkstep.c
gardner48: This duplicates the `tcur` update outside the loop and could be removed
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2300,6 +2355,9 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
   /* Evaluate stages j = 6,...,9 */
   for (int j = 6; j <= 9; j++)
   {
+    /* set stage index (0-based) */
+    step_mem->istage = j - 1;
```
gardner48: The indexing is strange in this case since the RHS is ever evaluated at "stage 5". 

If we do update `istage` to account for "stage 5" then 
 
* Add `istage = 6` outside the loop when `tcur` is updated
* Change this to `istage = j + 1` and move it to after the RHS evaluation when `tcur` is updated

In which case the last value from the loop is `istage = 10` plus one more for the final time step solution (i.e., 12 stages in total) 

Alternatively, we ignore "stage 5" in the `istage` updates: 
* No need to update `istage` above when `tcur` is updated as it's already correct from the end of the last loop
* Change this to `istage = j` and move it to after the RHS evaluation when `tcur` is updated

In which case the last value from the loop is `istage = 9` plus one more for the final time step solution (i.e., 11 stages in total)

src/arkode/arkode_lsrkstep.c
```diff
@@ -2300,6 +2355,9 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
...
+
     /* Complete previous stage by evaluating RHS and storing in tempv3 */
     ark_mem->tcur = ark_mem->tn + (j - 4) * hsixth;
```
gardner48: This is already computed outside the loop when updating `tcur` (or on prior iterations) and can be removed
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2361,7 +2419,10 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
     }
   }
 
-  /* Complete the 9th stage by evaluating RHS and storing in tempv3 */
+  /* set stage index (0-based) */
+  step_mem->istage = 9;
```
gardner48: This will already be set at the end of the last loop
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2361,7 +2419,10 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
...
+
+  /* Complete the last stage by evaluating RHS and storing in tempv3 */
   ark_mem->tcur = ark_mem->tn + ark_mem->h;
```
gardner48: This is already set at the end of the loop and could be removed
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2361,7 +2419,10 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
...
 
   /* apply user-supplied stage preprocessing function (if supplied) */
```
gardner48, line 2388: Update `istage` for the final solution evaluation. In which case, the logging statements need to be corrected for the last stage end the prior stage list and start/end a new stage list to compute the linear combination for the final solution

src/arkode/arkode_mristep.c
```diff
@@ -2013,6 +2017,9 @@ int mriStep_TakeStepMRIGARK(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
       }
     }
 
+    /* Update current stage index */
+    step_mem->istage = is;
```
gardner48: Move this up to the start of the loop when `tcur` so `istage = 1` when `PostProcessStageFn` is called with `z_1`

src/arkode/arkode_mristep.c
```diff
@@ -2141,7 +2148,8 @@ int mriStep_TakeStepMRIGARK(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
   /* perform embedded stage (if needed) */
   if (do_embedding)
   {
-    is = step_mem->stages;
+    /* Set current stage index */
+    step_mem->istage = is = step_mem->stages;
```
gardner48: Note `istage = s` corresponds to the MRI embedding in the `GetStageIndex` docs. Seems like this is unique to MRIStep, does this happen for any other steppers?

src/arkode/arkode_mristep.c
```diff
@@ -2517,6 +2529,9 @@ int mriStep_TakeStepMRISR(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
   /* Loop over stages */
   for (stage = 1; stage < max_stages; stage++)
   {
+    /* Set current stage index */
+    step_mem->istage = stage;
```
gardner48: Should probably move the `tcur` update from below to here and replace uses of `ark_mem->tn + cstage * ark_mem->h` with `tcur`

src/arkode/arkode_mristep.c
gardner48: Looks like MRI-SR and mriStep_StageDIRKNoFast` use `istage = stages - 1` for the embedding (and it's tied to the `c` value used in `mriStep_Predict` and `mriStep_StageSetup`). We should try to be consistent if possible but note the difference if not.

src/arkode/arkode_mristep.c
```diff
@@ -3028,7 +3046,7 @@ int mriStep_TakeStepMERK(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
     {
       /* Get stage index from group; skip to the next group if
          we've reached the end of this one */
-      stage = step_mem->MRIC->group[ig][is];
+      step_mem->istage = stage = step_mem->MRIC->group[ig][is];
```
gardner48: I don't remember, is the stage index unique across groups or is there overlap i.e., would a user need to distinguish between `istage = 1` in group `0` and group `1`?

src/arkode/arkode_splittingstep.c
```diff
@@ -458,6 +461,31 @@ static int splittingStep_SetOrder(ARKodeMem ark_mem, int order)
...
+{
+  ARKodeSplittingStepMem step_mem;
+  int retval = splittingStep_AccessStepMem(ark_mem, __func__, &step_mem);
+  if (retval != ARK_SUCCESS) { return (retval); }
+
+  /* if coefficients structure is not yet available, return defaults */
+  if (step_mem->coefficients == NULL)
+  {
+    *istage     = 0;
+    *num_stages = 1;
```
gardner48: I think we should return an error in this case. To other methods need to similarly check if their method tables are `NULL`?

src/arkode/arkode_sprkstep_io.c
```diff
@@ -275,6 +275,35 @@ int sprkStep_SetOrder(ARKodeMem ark_mem, int ord)
...
+
+  /* access ARKodeSPRKStepMem structure */
+  retval = sprkStep_AccessStepMem(ark_mem, __func__, &step_mem);
+  if (retval != ARK_SUCCESS) { return (retval); }
+
+  /* if table is not yet set, return defaults */
+  if (step_mem->method == NULL)
+  {
+    *stage      = 0;
+    *max_stages = 1;
```
gardner48: Same comment as in splitting step

CHANGELOG.md
```diff
@@ -35,6 +35,11 @@ stage-related data allocation before the first call to `ARKodeEvolve`
 (but after all other optional input routines have been called), to support
 users who measure memory usage before beginning a simulation.
 
+Added the function `ARKodeGetStageIndex` that returns the index of the stage
+currently being processed, and the total number of stages in the method, for users
+who must compute auxiliary quantities in their IVP right-hand side functions during
+some stages and not others (e.g., in all but the first or last stage).
```
gardner48: Note the addition of the functions to get the last time and state.

gardner48 (changes requested):

doc/arkode/guide/source/Usage/User_callable.rst
gardner48: Also see the [comment](https://github.com/llnl/sundials/pull/819#discussion_r2973134428) later about MRI methods and the stage index for embeddings

gardner48:
To summarize the comments above, `istage` should be updated when `tcur` is updated so calls to `PostprocessStage`, `PreRhs`, and the RHS functions with the same `t_i, y_i` have the same value of `istage`. This is already the case in ARKStep and some of the MRI steppers but the other steppers have `istage = i-1` when `PostporcessStage` is called and `istage = i` when `PreRhs` and the RHS functions are called.

doc/arkode/guide/source/Usage/User_supplied.rst
drreynolds: I'd ensured that the style of the new functions matched the rest of this file.  I've now adjusted **all** functions in this file to match the style guide.

src/arkode/arkode_mristep.c
drreynolds: This only occurs for MRI methods, since they are the only ones (so far) that have any "work" to do during an embedding stage (unlike Runge--Kutta methods that compute the embedded error using available information).  I'll update the documentation to clarify this point.

src/arkode/arkode_splittingstep.c
drreynolds: No, all the other steppers store the number of stages in the main stepper memory; SplittingStep omits it there, so it is only accessible from within the coefficient table.

I'll change this to return an error.

src/arkode/arkode_lsrkstep.c
drreynolds: The last iteration of the loop has `j=req_stages` but sets `step_mem->istage = j - 1`, so the loop here (and in RKL) never sets `istage` to equal `req_stages`.  I want this value to remain here, since RKL and RKC evaluate the RHS at the end of the step to compute the temporal error estimate and to prepare for the next step, so unlike ERK methods where `istage` never equals `s`, it will equal it here.

src/arkode/arkode_lsrkstep.c
drreynolds: The last iteration of the loop would set `step_mem->istage = step_mem->req_stages - 2`, so this update is still needed (whether here or elsewhere below).

src/arkode/arkode_lsrkstep.c
drreynolds: The last iteration of the loop would end with `step_mem->istage` one smaller than this value, so the update is still needed (whether it is placed here or somewhere below)

src/arkode/arkode_lsrkstep.c
drreynolds: The end of the last loop would have `step_mem->istage` one smaller, so the update is still needed (here or elsewhere below)

src/arkode/arkode_mristep.c
drreynolds: Yes, stage indices are unique, we just need to execute these out of order to exploit structure, which is why they are "grouped" separately.

src/arkode/arkode_lsrkstep.c
gardner48: This suggestion relies on the comment [here](https://github.com/llnl/sundials/pull/819#discussion_r2972735627) to change how and where `istage` is set.

Using `istage = j - 1` at the start of the loop is correct but there is an issue with mismatched indexes when calling `PostProcessStageFn` that makes this redundant:

* `istage = 1` is missing before the loop when `tmp2 = z_1` is computed and passed to `PostProcessStageFn`. Currently `istage = 0` when that call is made but `t_1, z_1` is passed to the function which I think will be confusing to users that would expect `istage = 0` to correspond to the inputs being `t_0 = t_n` and `z_0 = y_n`

* Adding `istage = 1` outside the loop makes setting `istage = j - 1` at the start of the loop redundant

* This same mismatched index issue occurs within the loop after the RHS evaluation is made and we compute `z_j` then call `PostProcessStageFn` with `t_j, z_j` but `istage = j - 1`. Moving the `istage` update to after the RHS evaluation where we update `tcur` and making it `istage = j` instead would fix this issue

* Now that `istage = j` is used within the loop, the last value will be `istage = req_stages` so setting `istage = req_stages` here is redundant

This same pattern shows up in the other `TakeStep` functions as well

src/arkode/arkode_splittingstep.c
```diff
@@ -334,6 +335,7 @@ static int splittingStep_TakeStep(ARKodeMem ark_mem, sunrealtype* dsmPtr,
 
   for (int i = 1; i < coefficients->sequential_methods; i++)
   {
+    step_mem->istage = i;
```
Steven-Roberts: This is counterintuitive to the standard definition of stage in the context of operator splitting. Normally it refers to the 2nd loop in the [algorithm in the docs](https://sundials.readthedocs.io/en/latest/arkode/Mathematics_link.html#arkode-mathematics-splittingstep). Here it is referring to the 1st loop. If you want to keep what's here, I think the docs should clarify this.

doc/arkode/guide/source/Mathematics.rst
```diff
@@ -527,9 +527,9 @@ LSRKStep currently supports two families of second-order, explicit, and temporal
 Runge--Kutta--Chebyshev (RKC), :cite:p:`VSH:04` and Runge--Kutta--Legendre (RKL), :cite:p:`MBA:14`.   These methods have the form
 
 .. math::
-   z_0 &= y_n,\\
-   z_1 &= z_0 + h_n \tilde{\mu}_1 f(t_n,z_0),\\
-   z_j &= (1-\mu_j-\nu_j)z_0 + \mu_j z_{j-1} + \nu_jz_{j-2} + h_n \tilde{\gamma}_j f(t_n,z_0) + h_n \tilde{\mu}_j f(t_n + c_{j-1}h, z_{j-1}) \\
+   z_1 &= y_n,\\
+   z_2 &= z_1 + h_n \tilde{\mu}_2 f(t_n,z_1),\\
+   z_j &= (1-\mu_j-\nu_j)z_1 + \mu_j z_{j-1} + \nu_jz_{j-2} + h_n \tilde{\gamma}_j f(t_n,z_1) + h_n \tilde{\mu}_j f(t_n + c_{j-1}h, z_{j-1}) \\
    y_{n+1} &= z_s.
```
gardner48: Should we change this to `z_{s+1}`  to align with the note in the `ARKodeGetStageIndex` docs? Also, add the range for `j`.
```suggestion
   z_j &= (1-\mu_j-\nu_j)z_1 + \mu_j z_{j-1} + \nu_jz_{j-2} + h_n \tilde{\gamma}_j f(t_n,z_1) + h_n \tilde{\mu}_j f(t_n + c_{j-1}h, z_{j-1}), \quad j = 3,\ldots,s+1 \\
   y_{n+1} &= z_{s+1}.
```

src/arkode/arkode_io.c
```diff
@@ -3174,6 +3214,41 @@ int ARKodeGetUserData(void* arkode_mem, void** user_data)
...
+  first of only a single stage.
```
gardner48: Update comment for revised function behavior
```suggestion
  stages. If this is not supplied by the time-stepping module
  then an error is returned and the values are set to (-1, -1).
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -677,7 +682,8 @@ int lsrkStep_TakeStepRKC(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
   mus  = w1 * bjm1;
 
   /* Evaluate the first stage (store in tmp2) and initialize embedding */
```
gardner48: There is a minor inconsistency I noticed when looking at the changes/comments in SSP43. This is stage index 1, so it is the second stage. Do you prefer
```
/* Begin stage 1 ... */
```
so the comment matches the index or
```
/* Begin the second stage ...*/
```
so the comment matches the count?

It looks like we mostly use the former, so I'll suggest that when I see it, but either convention is good with me.

```suggestion
  /* Begin stage 1 (store in tmp2) and initialize embedding */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -709,7 +715,6 @@ int lsrkStep_TakeStepRKC(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
   for (int j = 2; j <= step_mem->req_stages; j++)
   {
     /* Evaluate RHS and store in ycur */
```
gardner48:
```suggestion
    /* Complete the previous stage (evaluate the RHS and store it in ycur) */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -737,16 +742,17 @@ int lsrkStep_TakeStepRKC(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
     SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
 
     /* Compute new stage value and store in ycur */
```
gardner48:
```suggestion
    /* Begin stage j (store in ycur) */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1019,10 +1029,10 @@ int lsrkStep_TakeStepRKL(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
   bjm2 = ONE / THREE;
   bjm1 = bjm2;
   mus  = w1 * bjm1;
-  cjm1 = mus;
 
   /* Evaluate the first stage (store in tmp2) and initialize embedding */
```
gardner48:
```suggestion
  /* Begin stage 1 (store in tmp2) and initialize embedding */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1044,7 +1054,6 @@ int lsrkStep_TakeStepRKL(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
   for (int j = 2; j <= step_mem->req_stages; j++)
   {
     /* Evaluate RHS and store in ycur */
```
gardner48:
```suggestion
    /* Complete the previous stage (evaluate the RHS and store it in ycur) */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1072,14 +1081,15 @@ int lsrkStep_TakeStepRKL(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
     SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
 
     /* Compute new stage value and store in ycur */
```
gardner48:
```suggestion
    /* Begin stage j (store in ycur) */
```

src/arkode/arkode_lsrkstep.c
gardner48:
```suggestion
  /* Begin stage 0 */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1301,7 +1313,8 @@ int lsrkStep_TakeStepSSPs2(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
   SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
 
   /* Begin the second stage, and accumulate embedding into tempv1 */
```
gardner48:
```suggestion
  /* Begin stage 1 and accumulate embedding into tempv1 */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1327,7 +1340,6 @@ int lsrkStep_TakeStepSSPs2(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
   for (int j = 2; j < step_mem->req_stages; j++)
   {
     /* Complete previous stage by evaluating RHS and storing it in tempv2 */
```
gardner48:
```suggestion
    /* Complete the previous stage (evaluate the RHS and store it in tempv2) */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1357,7 +1369,8 @@ int lsrkStep_TakeStepSSPs2(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
     SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
 
     /* Begin the j-th stage by updating the state and embedding */
```
gardner48:
```suggestion
    /* Begin stage j (update the state and embedding) */
```

src/arkode/arkode_lsrkstep.c
gardner48:
```suggestion
  /* Begin stage 0 */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1541,7 +1558,8 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
   SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
 
   /* Begin the second stage, and accumulate embedding into tempv1 */
```
gardner48:
```suggestion
  /* Begin second 1 and accumulate embedding into tempv1 */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1567,7 +1585,6 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
   for (int j = 2; j <= ((in - 1) * (in - 2) / 2); j++)
   {
     /* Complete previous stage by evaluating RHS and storing it in tempv3 */
```
gardner48:
```suggestion
    /* Complete the previous stage (evaluate the RHS and store it in tempv3) */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1597,7 +1614,8 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
     SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
 
     /* Begin the j-th stage by updating the state and embedding */
```
gardner48:
```suggestion
    /* Begin stage j (update the state and embedding) */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1628,7 +1646,6 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
   for (int j = ((in - 1) * (in - 2) / 2 + 1); j <= (in * (in + 1) / 2 - 1); j++)
   {
     /* Complete previous stage by evaluating RHS and storing it in tempv3 */
```
gardner48:
```suggestion
    /* Complete the previous stage (evaluate the RHS and store it in tempv3) */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1658,7 +1675,8 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
     SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
 
     /* Begin the j-th stage by updating the state and embedding */
```
gardner48:
```suggestion
    /* Begin stage  j (update the state and embedding) */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1750,7 +1766,6 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
   for (int j = (in * (in + 1) / 2 + 1); j <= step_mem->req_stages; j++)
   {
     /* Complete the previous stage */
```
gardner48:
```suggestion
    /* Complete the previous stage (evaluate the RHS and store it in tempv3) */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1780,7 +1795,8 @@ int lsrkStep_TakeStepSSPs3(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
     SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
 
     /* Begin the j-th stage by updating the state and embedding */
```
gardner48:
```suggestion
    /* Begin stage j (update the state and embedding) */
```

src/arkode/arkode_lsrkstep.c
gardner48:
```suggestion
  /* Begin stage 0 */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1919,7 +1938,8 @@ int lsrkStep_TakeStepSSP43(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
   SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
 
   /* Perform the second stage, and accumulate embedding into tempv1 */
```
gardner48:
```suggestion
  /* Begin stage 1 and accumulate embedding into tempv1 */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -1968,7 +1988,8 @@ int lsrkStep_TakeStepSSP43(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
   SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
 
   /* Perform the third stage */
```
gardner48:
```suggestion
  /* Begin stage 2 */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2018,7 +2039,8 @@ int lsrkStep_TakeStepSSP43(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr
   SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
 
   /* Perform the fourth stage */
```
gardner48:
```suggestion
  /* Begin stage 3 */
```

src/arkode/arkode_lsrkstep.c
gardner48:
```suggestion
  /* Begin stage 0 */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2200,7 +2226,8 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
   N_VScale(ONE, ark_mem->yn, ark_mem->tempv2);
 
   /* Begin the second stage, and accumulate embedding into tempv1 */
```
gardner48:
```suggestion
  /* Begin stage 1 and accumulate embedding into tempv1 */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2228,6 +2252,8 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
       }
     }
 
+    /* Complete previous stage by evaluating RHS and storing in tempv3 */
+
```
gardner48:
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2212,9 +2239,6 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
   /* Evaluate stages j = 2,...,5 */
   for (int j = 2; j <= 5; j++)
   {
```
gardner48, line 2241:
```suggestion
  {
    /* Complete the previous stage (postprocesses the stage, evaluate the RHS, and
       store it in tempv3) */

```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2258,6 +2284,7 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
     /* Begin the j-th stage by updating the state and embedding */
```
gardner48:
```suggestion
    /* Begin stage j (update the state and embedding) */
```

src/arkode/arkode_lsrkstep.c
gardner48: `tcur` is already computed above
```suggestion
/* Begin stage 6 */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2300,7 +2327,6 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
   for (int j = 6; j <= 9; j++)
   {
     /* Complete previous stage by evaluating RHS and storing in tempv3 */
```
gardner48:
```suggestion
    /* Complete previous stage (evaluate RHS and store it in tempv3) */
```

src/arkode/arkode_lsrkstep.c
gardner48: `istage = 5` at this point. Should we update the comment (`/* Continue stage 5 */`) and logging (remove `end-stages-list` above and `being-stages-list` here) to match or add `istage = 6` here? Also, the `tcur` update can be dropped since it's the same from the end of the loop above.

Based on the comments/changes in the loop below, I'll suggest the former changes (assuming it works correctly on lines unchanged by the PR) but either is good as long as `istage` and the logging are consistent.
```suggestion
/* Continue stage 5 */
```

src/arkode/arkode_lsrkstep.c
gardner48:
```suggestion
    /* Complete the previous stage (evaluate the RHS and store in tempv3) */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2330,9 +2356,10 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
     SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
 
     /* Begin the j-th stage by updating the state and embedding */
```
gardner48:
```suggestion
    /* Begin stage j (update the state and embedding) */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2360,8 +2387,7 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
     }
   }
 
-  /* Complete the 9th stage by evaluating RHS and storing in tempv3 */
-  ark_mem->tcur = ark_mem->tn + ark_mem->h;
+  /* Complete the last stage by evaluating RHS and storing in tempv3 */
```
gardner48:
```suggestion
  /* Complete the previous stage (evaluate the RHS and store it in tempv3) */
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2379,21 +2405,22 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
                         ark_mem->user_data);
   step_mem->nfe++;
 
-  SUNLogExtraDebugVec(ARK_LOGGER, "stage RHS", ark_mem->tempv3, "F_9(:) =", 9);
+  SUNLogExtraDebugVec(ARK_LOGGER, "stage RHS", ark_mem->tempv3, "F_9(:) =");
   SUNLogInfoIf(retval != 0, ARK_LOGGER, "end-stages-list",
                "status = failed rhs eval, retval = %i", retval);
 
   if (retval < 0) { return ARK_RHSFUNC_FAIL; }
   if (retval > 0) { return RHSFUNC_RECVR; }
```
gardner48, line 2409: End the logging for the previous stage and start the logging for `istage = 10` below.
```suggestion
  if (retval > 0) { return RHSFUNC_RECVR; }
  
  SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
  SUNLogInfo(ARK_LOGGER, "begin-stages-list",
           "stage = %i, tcur = " SUN_FORMAT_G, 10, ark_mem->tcur);
```

src/arkode/arkode_mristep.c
gardner48: I think this needs to be changed to local variable instead of using `istage` otherwise the solution and embedding have the same stage `istage` value when getting the stage index. The adjusted value with an embedding is still needed as an input to `mriStep_Predict` to prevent using an out of bound index there. 

This issue similarly happens in `mriStep_StageDIRKNoFast` which also calls `mriStep_Predict`. However, using a local variable there would also impact the `mriStep_StageSetup` call which access `istage` from the memory structure. Making the stage index an input to `mriStep_StageSetup` would fix that issue. 

Also the check here could use the `embedding` flag set above instead of checking `stage == stages`.

src/arkode/arkode_mristep_io.c
```diff
@@ -828,6 +828,26 @@ int mriStep_GetNonlinSolvStats(ARKodeMem ark_mem, long int* nniters,
...
+{
+  ARKodeMRIStepMem step_mem;
+  int retval;
+
+  /* access ARKodeMRIStepMem structure */
+  retval = mriStep_AccessStepMem(ark_mem, __func__, &step_mem);
+  if (retval != ARK_SUCCESS) { return (retval); }
+
+  *stage      = step_mem->istage;
+  *max_stages = step_mem->stages;
```
gardner48, line 847: Does this need to be `+1` for embeddings?

gardner48 (changes requested):
* A minor inconsistency between the `istage` value and the logging output in SSP104 (`istage = 5` ends early and the logging for `istage = 9` and `istage = 10` is combined)
* A minor issue in MRIStep where `istage` has the same value for the final stage and embedding
* Suggestions to consistently refer to a stage by its index instead of count in LSRKStep comments 
* Small doc and other suggestions

src/arkode/arkode_mristep_io.c
drreynolds: Yes, but only if we're using an MRI method with its embedding.  I'll update this to key off the `fixed_step` flag.

drreynolds: Actually, I'm not sure.  In the papers that introduce MRI methods (even those with embeddings), the number of "stages" does not include the embedding.  Thus in the documentation for `ARKodeGetStageIndex` I added the note that "For temporally adaptive computations in MRIStep, the "embedding" stage is indicated using `stage` **equal to** `max_stages`."

If we keep this function as I'd implemented, then it is consistent with the literature, our documentation of the mathematics, and my documentation note for `ARKodeGetStageIndex`.  Whereas if I use your recommendation then it redefines the number of "stages" to be consistent with other ARKODE solvers, but it may not be what users expect when finding these methods in the literature.

src/arkode/arkode_mristep.c
drreynolds: It is more complicated than that, since `step_mem->istage` is also referenced throughout `arkode_mristep_nls.c` to indicate the vector storage location to use throughout the nonlinear solve.

Instead of adding function arguments and potentially breaking the [somewhat delicate] use of `step_mem->istage`, I think a much cleaner solution is to just add a counter to the MRIStep memory to keep track of the "actual" current stage for reporting by `ARKodeGetStageIndex` (as I added for some other steppers).

I've pushed an update that takes this path, which should resolve the underlying issue in this comment.

drreynolds:
@gardner48: I've addressed all your PR comments, taking action on all but one, where instead I explained why I'd like to keep it as-is.  If you're good with that explanation, then once this passes the CI I think it's ready to approve.

(It's possible that some of these PR revisions might adjust some of the logging output, so it could take a few commits to resolve CI testing).

src/arkode/arkode_mristep_io.c
gardner48: This is good as it is now and I think it is consistent with the other ARKODE solvers. I still had my head in the LSRKStep space where the number of "stages" in the literature is one less because the trivial initial stage is not counted, but we include it to make the indexing consistent across the steppers.

src/arkode/arkode_lsrkstep.c
```diff
@@ -2273,8 +2301,7 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
 
   SUNLogInfo(ARK_LOGGER, "end-stages-list", "status = success");
```
gardner48: I think it wasn't clear because of how GitHub shows suggestions on unchanged lines, this logging line needs to be deleted (the `end-stages-list` line inside the loop will close this stage's log, if successful).
```suggestion
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2273,8 +2301,7 @@ int lsrkStep_TakeStepSSP104(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
...
 
-  /* Begin the sixth stage */
-  ark_mem->tcur = ark_mem->tn + TWO * hsixth;
+  /* Finish stage 5 by preparing for the final stage group */
   SUNLogInfo(ARK_LOGGER, "begin-stages-list",
              "stage = %i, tcur = " SUN_FORMAT_G, 6, ark_mem->tcur);
```
gardner48: Like above I think it wasn't clear because of how GitHub shows suggestions on unchanged lines, this logging line also needs to be deleted (the `begin-stages-list` line inside the loop will open the stage 6 log).
```suggestion
```

src/arkode/arkode_mristep.c
gardner48: This looks good, but I think it would be helpful to co-locate the declaration of `istage` and `cur_stage` in the `arkode_mristep_impl.h` and add a comment to clarify that unlike in the other steppers `istage` can't be used for both the current stage index and the stage index for the nonlinear solve in MRIStep because of the "additional" stage for embeddings which needs to use the same `istage` value as the solution stage. So in this case `cur_stage` is like `istage` in the other steppers.

gardner48 (changes requested):
Minor logging fix and a suggestion to note that `istage` is slightly different in MRIStep

src/arkode/arkode_lsrkstep.c
gardner48: Note this suggestion (and the one below) will change the logging output. Currently, stage 6 is logged twice with inconsistent times e.g., 
```
[INFO][rank 0][lsrkStep_TakeStepSSP104][begin-stages-list] stage = 6, tcur = 2.03450520833333e-12
[INFO][rank 0][lsrkStep_TakeStepSSP104][end-stages-list] status = success
[INFO][rank 0][lsrkStep_TakeStepSSP104][begin-stages-list] stage = 6, tcur = 3.0517578125e-12
[INFO][rank 0][lsrkStep_TakeStepSSP104][end-stages-list] status = success
```

gardner48 (approved):
Looks good, just waiting on the CI now

drreynolds:
Strangely, that previous litgen CI failure is resurfacing periodically.

gardner48 (approved):

PR #821, opened 2026-01-21
Fix adjoint indices in ARKODE docs

Steven-Roberts:


doc/arkode/guide/source/Mathematics.rst
```diff
@@ -2762,14 +2762,14 @@ The discrete adjoint variables represent the gradients of the discrete cost func
...
 
 .. math::
-   \Lambda_i &= h_n f_y^*(t_{n,i}, z_i, p) \left(b_i \lambda_{n+1} + \sum_{j=i+1}^s a_{j,i} \Lambda_j \right), \quad \quad i = s, \dots, 1,\\
-   \lambda_n &= \lambda_{n+1} + \sum_{j=1}^{s} \Lambda_j, \\
-   \nu_i     &= h_n f_p^*(t_{n,i}, z_i, p) \left(b_i \lambda_{n+1} + \sum_{j=i+1}^{s} a_{j,i} \Lambda_j \right), \\
-   \mu_n     &= \mu_{n+1} + \sum_{j=1}^{s} \nu_j.
+   \Lambda_i &= h_k f_y^*(t_{k,i}, z_i, p) \left(b_i \lambda_{k+1} + \sum_{j=i+1}^s a_{j,i} \Lambda_j \right), \quad \quad i = s, \dots, 1,\\
+   \lambda_k &= \lambda_{k+1} + \sum_{j=1}^{s} \Lambda_j, \\
+   \nu_i     &= h_k f_p^*(t_{k,i}, z_i, p) \left(b_i \lambda_{k+1} + \sum_{j=i+1}^{s} a_{j,i} \Lambda_j \right), \\
+   \mu_k     &= \mu_{k+1} + \sum_{j=1}^{s} \nu_j.
```
gardner48: Why the change to `k`? We usually use `n` for the time step index.

Steven-Roberts: Because `n` is used as the final time step and `k` is an intermediate step in these two equations
https://sundials.readthedocs.io/en/latest/arkode/Mathematics_link.html#equation-arkode-discrete-optimization-problem
https://sundials.readthedocs.io/en/latest/arkode/Mathematics_link.html#equation-arkode-discrete-adjoint

The other option is to make `N` the final step and `n` an intermediate step.

gardner48: I'd prefer switching to `N` for the final step

doc/arkode/guide/source/Mathematics.rst
```diff
@@ -2712,12 +2712,12 @@ Now, suppose we have a functional :math:`g(t_f, y(t_f), p)` for which we would l
...
 
 .. math::
-   \frac{dg(t_f,y(t_n),p)}{dy}, \quad \text{and optionally}, \quad \frac{dg(t_f,y(t_n),p)}{dp}.
+   \frac{dg(t_f,y(t_N),p)}{dy}, \quad \text{and optionally}, \quad \frac{dg(t_f,y(t_N),p)}{dp}.
 
 This most often arises in the form of an optimization problem such as
 
 .. math::
-   \min_{y(t_0), p} g(t_f, y(t_n), p).
+   \min_{y(t_0), p} g(t_f, y(t_f), p).
```
drreynolds, line 2720: Should we specify that `t_N` is the same as `t_f` here?

drreynolds (commented):
One minor question; otherwise LGTM

doc/arkode/guide/source/Mathematics.rst
Steven-Roberts: For consistency, we are now only using `t_f`

drreynolds: Sure, but in the equations below we still use `t_N` as the end of the final time step, but here we use `t_f`.  I'm good with retaining both, since `t_f` comes from the continuous IVP, while `t_N` is the end of the final time step. If we never state that `t_f` is the same as `t_N`, then it could be challenging for readers to connect the above continuous problem with the discrete formulation below.  Perhaps the simplest approach for tidying this up would be to just state that these are the same?

drreynolds (dismissed):

gardner48 (approved):

drreynolds (approved):

PR #822, opened 2026-01-23
Fix target linking for shared+static builds

### LINKED ISSUE #692: Programm statically linked but still needs dll to run
I'm new to cmake and to sundials. I'm following the instructions for projects that use CMake in the documentation (https://sundials.readthedocs.io/en/v7.3.0/sundials/Install_link.html#cmake-projects) but changed the last line to `target_link_libraries(myexec PUBLIC SUNDIALS::core_static SUNDIALS::cvode_static) ` and added `#define SUNDIALS_STATIC_DEFINE` to my test file (cvRocket_dns.c). The resulting .exe fails to run with a missing sundials_core.dll error.  Changing the SUNDIALSTargets.cmake in INSTALL_DIR from 
```
set_target_properties(SUNDIALS::cvode_static PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "SUNDIALS_STATIC_DEFINE"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "SUNDIALS::core_shared"
)
```
to 
```
set_target_properties(SUNDIALS::cvode_static PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "SUNDIALS_STATIC_DEFINE"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "SUNDIALS::core_static"    <========
)
```
solves the problem.

Is it intended that all static libraries depend on SUNDIALS::core_shared?

balos1:
Fixes #692.


cmake/macros/SundialsAddLibrary.cmake
```diff
@@ -548,3 +555,22 @@ macro(append_static_suffix libs_in libs_out)
...
+      elseif(aliased_target_name MATCHES ${oppo_lib_suffix})
+        list(APPEND ${libs_out} ${_lib}${lib_suffix})
+      else()
+        list(APPEND ${libs_out} ${_lib})
+      endif()
+    else()
+      # handle visibility modifiers PUBLIC, PRIVATE etc.
+      list(APPEND ${libs_out} ${_lib})
+    endif()
+  endforeach()
```
gardner48, line 550: Do we really need `oppo_lib_suffix` or could this be
```suggestion
  foreach(_lib ${${libs_in}})
    if(TARGET ${_lib}${lib_suffix})
      list(APPEND ${libs_out} ${_lib}${lib_suffix})
    else()
      # handle imported targets (e.g., LAPACK::LAPACK, BLAS::BLAS, etc.) and visibility modifiers (PUBLIC, PRIVATE etc.)
      list(APPEND ${libs_out} ${_lib})
    endif()
  endforeach()
```
Also should we update (or remove) blocks like
```
      if(${_libtype} MATCHES "STATIC")
        append_static_suffix(sundials_add_library_LINK_LIBRARIES _all_libs)
      else()
        set(_all_libs ${sundials_add_library_LINK_LIBRARIES})
      endif()
```

gardner48: I think we can probably replace `append_static_suffix` with `dealias_libraries`

gardner48 (dismissed):

gardner48 (approved):

PR #824, opened 2026-01-25
Release: v7.6.0

gardner48:
SUNDIALS Release v7.6.0


gardner48:
@balos1 Seems like it's not an issue, but we should do something about warnings like this in the CI from `pypa/cibuildwheel@v2.22`?
```
sundials4py-build-and-test / Build extra wheels on macos-latest (single/32)
Bumping MACOSX_DEPLOYMENT_TARGET (10.15) to the minimum required (11.0).
```

drreynolds (approved):

balos1:
> @balos1 Seems like it's not an issue, but we should do something about warnings like this in the CI from `pypa/cibuildwheel@v2.22`?
> 
> ```
> sundials4py-build-and-test / Build extra wheels on macos-latest (single/32)
> Bumping MACOSX_DEPLOYMENT_TARGET (10.15) to the minimum required (11.0).
> ```

It should be fine for the release since it automatically is bumping the target to the minimum, but to get rid of the warnings I will open up a PR that fixes how we set the `MACOSX_DEPLOYMENT_TARGET` variable.

balos1 (approved):

PR #827, opened 2026-01-26
CI: Fix wheels publishing

balos1:


.github/workflows/sundials4py-deploy.yml
```diff
@@ -22,7 +22,14 @@ jobs:
   sundials4py-deploy-build-and-test:
     name: Build and test
     if: github.event_name == 'release' && github.event.action == 'published'
-    uses: ./.github/workflows/sundials4py-build-and-test.yml
+    steps:
+    - name: Clone repository
+      uses: actions/checkout@v5
+      with:
+        submodules: true
```
gardner48, line 29: Is this necessary? The jobs in `sundials4py-build-and-test.yml` also clone the repo

balos1: Yes, otherwise [it doesn't seem to find the workflow file](https://github.com/llnl/sundials/actions/runs/21370677495/job/61518720588) to use because the repo has not been cloned yet.

gardner48: Interesting, the main ci workflow only has a checkout in the detect-changes job but not the others. Maybe that initial workflow with a checkout is all that's needed for the uses statements work in the other jobs.

balos1: Yeah, there has to be a clone before the uses statement with a file path. Since the deploy workflow doesn't start with detect-changes, there is no clone until this first stage.

gardner48 (approved):

PR #803, opened 2025-11-21
CI: Add -Wundef

gardner48:
Add `-Wundef` with `ENABLE_ALL_WARNINGS`

PR #835, opened 2026-01-28
CMake: Set SUPERLUMT_WORKS=TRUE after success

jschueller:
Amends #803

if SUPERLUMT_WORKS is not set to TRUE after a successful compilation test then it is considered disabled

/cc @gardner48


gardner48:
@jschueller Thanks for reporting this issue. I'm closing this PR in favor of the fix in #836. Until that's merged, setting the `SUPERLUMT_WORKS` option to `TRUE` when configuring will workaround the bug in CMake.

jschueller:
thanks

PR #836, opened 2026-01-28
Bugfix: Remove dependency on `<TPL>_WORKS` variables

gardner48:
Remove `<TPL>_WORKS` variables as dependencies for building the corresponding interface library. Fatal errors will occur before these are set to `TRUE` in the various checks, so the `WORKS` variables are not actually needed to check if the library is found and works with SUNDIALS. The options are only needed to skip the compatibility and compile checks.


gardner48:
#837 deprecates the `<TPL>_WORKS` options and replaces them with `SUNDIALS_ENABLE_<TPL>_CHECKS` to make the purpose of the option clearer (and namespace the option).

balos1 (approved):

PR #850, opened 2026-02-22
Bugfix: Fix the embedded coefficients of the ARKODE_TSITOURAS_7_4_5 method

Steven-Roberts:


src/arkode/arkode_butcher_erk.def
```diff
@@ -486,13 +486,18 @@ ARK_BUTCHER_TABLE(ARKODE_TSITOURAS_7_4_5, { /* Tsitouras-ERK */
...
+     * lists bHat coefficients but actually reports (bHat - b) with a sign error
+     * in the last coefficient. The Mathematica code uses bHat_7 = 1/40 when it
+     * should be 1/66. This is why these values do not match the paper. */
+    B->d[0] = SUN_RCONST(0.09468075576583945807478876255758922856118);
+    B->d[1] = SUN_RCONST(0.009183565540343253096776363936645313759814);
+    B->d[2] = SUN_RCONST(0.4877705284247615707855642599631228241517);
+    B->d[3] = SUN_RCONST(1.234297566930478985655109673884237654036);
+    B->d[4] = SUN_RCONST(-2.707712349983525454881109975059321670690);
+    B->d[5] = SUN_RCONST(1.866628418170587035753719399566211498666);
+    B->d[6] = SUN_RCONST(0.01515151515151515151515151515151515151515);
```
gardner48, line 500: Update the table in the documentation and add a similar note there as well. Also note the fix in the change log.

gardner48 (changes requested):

doc/arkode/guide/source/Butcher.rst
Steven-Roberts, line 758: TODO: Update the stability region

drreynolds (dismissed):
I find it interesting that although these updates slightly change the solutions, the accuracy is essentially the same as before, and the number of adaptive steps is also essentially unchanged.

Steven-Roberts:
> I find it interesting that although these updates slightly change the solutions, the accuracy is essentially the same as before, and the number of adaptive steps is also essentially unchanged.

Yes, I thought the same. But it is consistent with the fact that the new and old embedded coefficients have nearly identical quality metrics ($B^{(6)}, C^{(6)}$).

gardner48 (approved):

PR #853, opened 2026-03-03
Feature: stageprocessing updates

gardner48:
Suggested updates to #813

* Replace `PreProcessStep` with `PreStepFn` -- removes "process" from the name, adds the current step and attempt index as inputs, and always call before a step attempt
* Remove `PostProcessStepFail` -- users can handle this case by checking the attempt index in `PreProcessStep`
* Replace `PostProcessStep` call in `arkCompleteStep` with a `PostStepFn`
* Replace `PostProcessStage` calls on the new state in `TakeStep` functions with `PostProcessStep` calls and account cases where the last stage is the new state
* Replace `PreRHSProcess` with `PreRhsFn` to remove the idea of processing from the name

This does not address this additional copy before the time step loop or on a failed step

### DIFF
--- a/src/arkode/arkode.c
+++ b/src/arkode/arkode.c
@@ -915,7 +915,7 @@ int ARKodeEvolve(void* arkode_mem, sunrealtype tout, N_Vector yout,
       if (ark_mem->PreStepFn)
       {
         retval = ark_mem->PreStepFn(ark_mem->tcur, ark_mem->ycur, ark_mem->nst,
-                                    ark_mem->nst_attempts, ark_mem->user_data);
+                                    attempts, ark_mem->user_data);
         if (retval != 0) { return (ARK_PRESTEPFN_FAIL); }
       }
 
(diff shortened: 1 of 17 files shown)


src/arkode/arkode.c
```diff
@@ -913,14 +911,13 @@ int ARKodeEvolve(void* arkode_mem, sunrealtype tout, N_Vector yout,
...
-      /* call the user-supplied step preprocessing function (if it exists) */
-      if (ark_mem->PreProcessStep != NULL && !skip_preprocess)
+      /* call the user-supplied pre-step function (if it exists) */
+      if (ark_mem->PreStepFn)
       {
-        retval = ark_mem->PreProcessStep(ark_mem->tcur, ark_mem->ycur,
-                                         ark_mem->ps_data);
-        if (retval != 0) { return (ARK_PREPROCESS_STEP_FAIL); }
+        retval = ark_mem->PreStepFn(ark_mem->tcur, ark_mem->ycur, ark_mem->nst,
+                                    ark_mem->nst_attempts, ark_mem->user_data);
```
drreynolds: If the purpose of this change is to notify the `PreStepFn` that it is re-attempting a step, then this should be the `int attempts` instead of the global `long int ark_mem->nst_attempts` counter.  This would also change the function signature in `include/arkode/arkode.h`
```suggestion
                                    attempts, ark_mem->user_data);
```

gardner48: Good catch, that's what I indeed to use.

drreynolds (approved):

PR #861, opened 2026-03-06
CMake: Enable C in SUNDIALSConfig if necessary

gardner48:
Check if C is enable before looking for MPI (e.g., with Fortran codes that don't enable C in CMake)


balos1 (dismissed):

Steven-Roberts (approved):
I tested this on the Fortran code I mentioned a few days ago, and this resolve the issue.

PR #837, opened 2026-01-28
CMake: Prefix post options with `SUNDIALS_BUILD_`

gardner48:
* Update SUNDIALS post options to use `SUNDIALS_ENABLE_` instead of `BUILD_`
* Fix `DEPENDS_ON` for `SUNDIALS_ENABLE_CALIPER` and `SUNDIALS_ENABLE_ADIAK`

PR #808, opened 2025-12-30
Maintenance: deprecate cvode monitor function(s)

balos1:

PR #867, opened 2026-03-08
CMake: prefix monitoring option with SUNDIALS_

gardner48:
`SUNDIALS_BUILD_WITH_MONITORING` -> `SUNDIALS_ENABLE_MONITORING`


gardner48:
This PR extracts some of the changes from #837 to consistently prefix CMake options with `SUNDIALS_`

balos1:
Ill note that option will be deprecated by #808.

gardner48:
> Ill note that option will be deprecated by #808.

Yes, but we should updated it for consistency and then remove it entirely when the feature is removed.

balos1 (dismissed):

PR #860, opened 2026-03-06
Bugfix: Disable logging output with empty string

balos1:
Fixes #844

PR #882, opened 2026-03-20
Bugfix: Default logger not printing errors to stderr and warnings to stdout

Steven-Roberts:
Resolves a bug introduced in #860 where the default logger had it's error and warning file pointers overwritten to NULL.


balos1 (dismissed):

src/sundials/sundials_logger.c
```diff
@@ -195,20 +197,32 @@ SUNErrCode SUNLogger_CreateFromEnv(SUNComm comm, SUNLogger* logger_out)
...
 
   do {
-    err = SUNLogger_SetErrorFilename(logger, error_fname_env);
-    if (err) { break; }
-    err = SUNLogger_SetWarningFilename(logger, warning_fname_env);
-    if (err) { break; }
-    err = SUNLogger_SetDebugFilename(logger, debug_fname_env);
-    if (err) { break; }
-    err = SUNLogger_SetInfoFilename(logger, info_fname_env);
+    if (!sunIsNullOrEmpty(error_fname_env))
```
gardner48: This treats `NULL` (the environment variable does not exist) and an empty string (the environment variable exists but is an empty string) the same way which is probably not what we want.

The docs currently say, "To disable output for one of the streams, then do not set the environment variable, or set it to an empty string." However, I think the behavior we want is to keep the default if the environment variable is not set and disable the logging stream if it is an empty string.

src/sundials/sundials_logger.c
```diff
@@ -218,162 +232,138 @@ SUNErrCode SUNLogger_CreateFromEnv(SUNComm comm, SUNLogger* logger_out)
...
+  if (!sunIsNullOrEmpty(error_filename))
   {
     /* Don't close the file here, that is managed by the underlying hashmap */
     logger->error_fp = NULL;
     return SUN_SUCCESS;
   }
 
-  if (error_filename && strcmp(error_filename, ""))
+  if (SUNHashMap_GetValue(logger->filenames, error_filename,
+                          (void**)&logger->error_fp))
```
gardner48: This (and similarly for the other cases) will evaluate to true when `SUNHashMap_GetValue` returns `SUNHASHMAP_ERROR`

src/sundials/sundials_logger.c
```diff
@@ -428,19 +418,12 @@ SUNErrCode SUNLogger_QueueMsg(SUNLogger logger, SUNLogLevel lvl,
       }
     }
   }
-#else
-  /* silence warnings when all logging is disabled */
-  ((void)logger);
-  ((void)lvl);
-  ((void)scope);
-  ((void)label);
-  ((void)msg_txt);
```
gardner48, line 437: I believe we handled the warnings this way (here and elsewhere) so we could still catch unused warnings when logging is enabled as `SUNDIALS_MAYBE_UNUSED` will suppresses the warning in all cases.

gardner48 (changes requested):
It would also be good to add a unit test for the case this fixes

balos1 (dismissed):

src/sundials/sundials_logger.c
```diff
@@ -222,153 +266,50 @@ SUNErrCode SUNLogger_SetErrorFilename(SUNLogger logger, const char* error_filena
...
+  return sunLoggerSetFilename(logger, info_filename, &logger->info_fp);
+#else
+  ((void)info_filename);
   return SUN_SUCCESS;
+#endif
 }
 
-SUNErrCode SUNLogger_SetDebugFilename(SUNLogger logger, const char* debug_filename)
+SUNErrCode SUNLogger_SetDebugFilename(
+  SUNLogger logger, SUNDIALS_MAYBE_UNUSED const char* debug_filename)
```
gardner48: I don't think `SUNDIALS_MAYBE_UNUSED` is needed now
```suggestion
  SUNLogger logger, const char* debug_filename)
```

gardner48 (changes requested):
Also need to correct the sentence in the logging docs ("To disable output for one of the streams, then do not set the environment variable, or set it to an empty string.") to say not setting the environment variable will result in the default being used and setting it to an empty string will disable the stream.

Steven-Roberts:
> Also need to correct the sentence in the logging docs ("To disable output for one of the streams, then do not set the environment variable, or set it to an empty string.") to say not setting the environment variable will result in the default being used and setting it to an empty string will disable the stream.

That was already done in #860

balos1:
> Also need to correct the sentence in the logging docs ("To disable output for one of the streams, then do not set the environment variable, or set it to an empty string.") to say not setting the environment variable will result in the default being used and setting it to an empty string will disable the stream.

It currently reads (on develop):
> To disable output for one of the streams, set the environment variable to an empty string. To leave the stream at its default output, do not set the environment variable.

gardner48:
Ah, I was looking at the docs for `main` not `develop`

gardner48 (approved):

PR #883, opened 2026-03-20
Tools: Make suntools an installable library

balos1:


suntools/README.md
drreynolds, line 1: I'm not knowledgeable about setting up Python libraries, but by its name, the `tools/pyproject.toml` file seems to call itself `tools`, but the contents of that file refer to itself as both `sundials-suntools` (project->name) and as `suntools` (tool.setuptools->packages).  This file then refers to it both as `tools` and `suntools`.

#. How consistent should these names be?
#. If this is only installed alongside the SUNDIALS source code then I imagine `tools` is sufficiently descriptive, but if this were available directly using `pip` then I'd think a more descriptive name would be helpful.

drreynolds (dismissed):
LGTM, aside from my question regarding naming.

suntools/README.md
balos1: Thanks for pointing out the inconsistency. I have updated both the naming and structure to make suntools more of a standalone module.

drreynolds (dismissed):

suntools/README.md
gardner48: Should we keep the `tools` directory? I think the original intent was we could eventually add other, non-python, tools here as well (like those in the [analysis-utilities](https://github.com/sundials-codes/analysis-utilities) repo).

Maybe this structure would work:
```
sundials/
+-- tools/
|   +-- suntools/
|       +-- pyproject.toml
|       +-- README.md
|       +-- examples/
|       `-- src/
```

balos1: I think those could go under `scripts` or be folded into a submodule of `suntools` if they are python.

drreynolds:
This seems to be failing the CI; once it passes I'll "approve"

suntools/README.md
gardner48: Do we foresee having non-python tools? In which case it might be nice to keep all tools grouped under a `tools` directory and all python-based tools go into `tools/suntools`

gardner48:
Looks like the tarscript in the Jenkins CI is failing because of the directory changes

suntools/README.md
balos1: I don't foresee anything that is as significant as suntools and it would be best that any significant tool is added as a suntools submodule.

drreynolds: I'll go ahead and mark this conversation as "resolved" since my original comment is satisfied.  However, if @gardner48 would like more follow-up, then please "unresolve" it.

drreynolds (dismissed):

gardner48 (approved):

PR #884, opened 2026-03-20
Tools: Add CLI interface for suntools

balos1:
This PR adds a CLI interface to suntools along with a command to parse logs and filter them. 

This is a part of a chain of PRs for suntools, starting with #883.


suntools/src/suntools/cli.py
```diff
@@ -0,0 +1,80 @@
...
+    )
+    parse_logs.set_defaults(func=_cmd_parse_logs)
+
+    return parser
+
+
+def main(argv: list[str] | None = None) -> int:
+    parser = build_parser()
+    args = parser.parse_args(argv)
+    return int(args.func(args))
```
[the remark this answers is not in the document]
balos1: OK, we should just require python 3.12 like sundials4py.

balos1:
@copilot apply changes based on the comments in [this thread](https://github.com/llnl/sundials/pull/884#pullrequestreview-4107449655).

drreynolds (commented):
This PR looks good to me.  My only request is that we provide some documentation so that users can better understand its potential and usage instructions.

suntools/src/suntools/logs.py
```diff
@@ -111,6 +112,214 @@ def _parse_logfile_line(line, line_number, all_lines):
...
+    return event, region_name, is_list
+
+
+def _split_filter_list(value: str) -> set[str]:
+    parts = [p.strip().lower() for p in value.split(",")]
+    return {p for p in parts if p}
+
+
+def _is_log_line(line: str) -> bool:
+    return line.startswith("[")
```
gardner48: Should this match the regex we use in `_parse_logfile_line` to identify logging output?

gardner48: Actually, I think we could remove this. See comment below.

suntools/src/suntools/logs.py
```diff
@@ -111,6 +112,214 @@ def _parse_logfile_line(line, line_number, all_lines):
...
+    lines: list[str], selected: set[str], invert: bool = False
+) -> Iterable[str]:
+    step_depth = 0
+    nonlinear_depth = 0
+    linear_depth = 0
+
+    i = 0
+    while i < len(lines):
+        line = lines[i]
+        if not _is_log_line(line):
```
gardner48: This can go under `if not entry:` after calling `_parse_logfile_line` below and `_is_log_line` could be removed

suntools/src/suntools/logs.py
```diff
@@ -111,6 +112,214 @@ def _parse_logfile_line(line, line_number, all_lines):
...
+            cats.add("integrator")
+
+        keep = bool(cats & selected)
+        if invert:
+            keep = not keep
+
+        # Only treat following non-log lines as a continuation block when the
+        # log line is an array/vector dump (payload values parsed as lists).
+        is_array_dump = any(isinstance(v, list) for v in entry.get("payload", {}).values())
+        continuation_end = _find_continuation_end(lines, i) if is_array_dump else i + 1
```
gardner48: Couldn't we use the length of the list created by `_parse_logfile_payload` instead and remove `_find_continuation_end`?

suntools/src/suntools/logs.py
```diff
@@ -111,6 +112,214 @@ def _parse_logfile_line(line, line_number, all_lines):
...
+        if region_name == "linear-solve":
+            if event == "begin":
+                linear_depth += 1
+            elif event == "end":
+                linear_depth = max(0, linear_depth - 1)
+
+        i = continuation_end
+
+
+def _get_history(log, key, step_status, time_range, step_range):
```
gardner48: This is a duplicate of `_get_history` at the end of this file.

test/unit_tests/suntools/test_cli_parse_logs.py
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env python3
+# ---------------------------------------------------------------
+# Programmer(s): SUNDIALS contributors
```
gardner48: I'm not sure a generic name is useful here
```suggestion
```

test/unit_tests/suntools/test_cli_parse_logs.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+class TestCliParseLogs(unittest.TestCase):
+    def _run(self, stdin_text, args):
+        env = os.environ.copy()
+        # Make sure we import the repo-local package.
+        repo_root = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", ".."))
+        env["PYTHONPATH"] = (
+            os.path.join(repo_root, "suntools", "src") + os.pathsep + env.get("PYTHONPATH", "")
+        )
```
gardner48: Should our other suntools test do this as well or is this only needed for the cli tests?

gardner48 (changes requested):

suntools/src/suntools/logs.py
gardner48: Should we add a python static analysis tool to the CI (e.g., pyright flags this issue)?

balos1: ruff has a [linter](https://docs.astral.sh/ruff/linter/) we can use now that we have switched to it for formatting.

balos1: I will setup linting in another PR.

test/unit_tests/suntools/test_cli_parse_logs.py
balos1: Yes, that way they can be executed directly and not just through ctest.

test/unit_tests/suntools/CMakeLists.txt
```diff
@@ -16,10 +16,11 @@
 # ---------------------------------------------------------------
 
 # Locate suntools module source directory.
-set(SUNTOOLS_DIR "${CMAKE_SOURCE_DIR}/suntools/src/suntools")
+set(SUNTOOLS_DIR "${CMAKE_SOURCE_DIR}/suntools/src")
```
gardner48: I think this change might be the source of the recent CI failures

test/unit_tests/suntools/CMakeLists.txt
```diff
@@ -18,8 +18,9 @@
 # Locate suntools module source directory.
 set(SUNTOOLS_DIR "${CMAKE_SOURCE_DIR}/suntools/src/suntools")
 
-if(NOT EXISTS "${SUNTOOLS_DIR}/logs.py")
-  message(WARNING "Skipped suntools log parser tests (logs.py not found)")
+if(NOT EXISTS "${SUNTOOLS_DIR}/suntools/logs.py")
```
gardner48: Fix path
```suggestion
if(NOT EXISTS "${SUNTOOLS_DIR}/logs.py")
```

gardner48 (changes requested):
One minor path fix

gardner48:
Looks like the issue is the CI container uses Python 3.8.10 but a newer Python is needed for the updated code

balos1:
> Looks like the issue is the CI container uses Python 3.8.10 but a newer Python is needed for the updated code

I think we can change the minimum back to 3.8. I had bumped it to match sundials4py but I dont think we are using anything from 3.12 in these scripts.

gardner48:
> I think we can change the minimum back to 3.8. I had bumped it to match sundials4py but I dont think we are using anything from 3.12 in these scripts.

I think there is some stuff like `set[str]` and `list[str] | None` that requires python 3.10

gardner48 (approved):

PR #886, opened 2026-03-22
Feature: Preallocate

drreynolds:
This PR adds a function ARKodeInit to enable stage-related data allocation before the first call to ARKodeEvolve (but after all other optional input routines have been called), to support users who measure memory usage before beginning a simulation.


doc/arkode/guide/source/Usage/User_callable.rst
```diff
@@ -5415,6 +5415,40 @@ Output all ARKODE solver parameters   :c:func:`ARKodeWriteParameters`
...
+allocation of stage-related internal data until the first call to
+:c:func:`ARKodeEvolve`.  However, in some cases the user may wish to
+preallocate this data earlier, for example to measure the memory footprint
+before beginning a calculation, or to check for allocation errors at an
+earlier time.  To request that that ARKODE preallocate all stage-related
+internal data before the first call to :c:func:`ARKodeEvolve`, the user
+may call the function :c:func:`ARKodeAllocateInternalData`.
+
+
+.. c:function:: int ARKodeAllocateInternalData(void* arkode_mem)
```
gardner48: I think it would be better to name this `ARKodeInit` to mirror the `Create` then `Init` call pattern of the other integrators with the difference here being that users to not have to call `ARKodeInit` (essentially we'll call it internally on the first step if necessary) but if they do, allocations will be performed at that time instead of in the first step.

I have some ideas for how we could clean up this function (and the initialization process in general) but all of that can happen behind the scenes at some point after this PR.
```suggestion
.. c:function:: int ARKodeInit(void* arkode_mem)
```

doc/arkode/guide/source/Usage/User_callable.rst
```diff
@@ -5415,6 +5415,40 @@ Output all ARKODE solver parameters   :c:func:`ARKodeWriteParameters`
...
+
+   Optionally allocates stage-related internal data for the current ARKODE time-stepper module.
```
gardner48: Maybe be more generic since we probably perform other allocation too
```suggestion
   Optionally allocates internal data for the current ARKODE time-stepper module.
```

src/arkode/arkode.c
```diff
@@ -1745,12 +1745,23 @@ int arkRwtSet(N_Vector y, N_Vector weight, void* data)
...
+  initialization, an error flag is returned. Otherwise, it
+  returns ARK_SUCCESS.
+
+  This routine should only be called by
+  (a) ARKodeReset (with the input init_type == RESET_INIT),
+  (b) an ARKODE timestepper module creation routine (with
+      init_type == FIRST_INIT), or
+  (c) an ARKODE timestepper module re-initialization routine
+      (with init_type == FIRST_INIT).
+  This should never by the user.
```
gardner48:
```suggestion
  This should never be called by the user.
```

src/arkode/arkode.c
```diff
@@ -1745,12 +1745,23 @@ int arkRwtSet(N_Vector y, N_Vector weight, void* data)
...
-  should be reinitialized (FIRST_INIT) or retained (RESET_INIT).
+  initialization, an error flag is returned. Otherwise, it
+  returns ARK_SUCCESS.
+
+  This routine should only be called by
+  (a) ARKodeReset (with the input init_type == RESET_INIT),
+  (b) an ARKODE timestepper module creation routine (with
+      init_type == FIRST_INIT), or
+  (c) an ARKODE timestepper module re-initialization routine
+      (with init_type == FIRST_INIT).
```
gardner48, line 1757: Add `ALLOC_INIT` case to this list (if it's needed, see comments below)

src/arkode/arkode_arkstep.c
```diff
@@ -1080,16 +1077,13 @@ int arkStep_Init(ARKodeMem ark_mem, SUNDIALS_MAYBE_UNUSED sunrealtype tout,
     {
       if (step_mem->z == NULL)
```
gardner48: This check can be removed since arkAllocVecArray will check if z == NULL (or updated to free existing data like in MRIStep)

src/arkode/arkode_arkstep.c
```diff
@@ -1169,7 +1163,7 @@ int arkStep_Init(ARKodeMem ark_mem, SUNDIALS_MAYBE_UNUSED sunrealtype tout,
   }
 
   /* Perform mass matrix solver initialization and setup (if applicable) */
-  if (step_mem->mass_type != MASS_IDENTITY)
+  if (step_mem->mass_type != MASS_IDENTITY && !step_mem->preallocated)
```
gardner48: Here (and similarly for the checks below), wouldn't these initialization calls still be needed if preallocating then calling `ReInit` later (e.g., to at least reinitialize counters)?

src/arkode/arkode_arkstep.c
```diff
@@ -1224,6 +1218,13 @@ int arkStep_Init(ARKodeMem ark_mem, SUNDIALS_MAYBE_UNUSED sunrealtype tout,
   /* Signal to shared arkode module that full RHS evaluations are required */
   ark_mem->call_fullrhs = SUNTRUE;
 
+  /* if init_type == ALLOC_INIT then store preallocated flag */
+  if (init_type == ALLOC_INIT) { step_mem->preallocated = SUNTRUE; }
+
+  /* if init_type == FIRST_INIT then reset preallocated flag (in case
+     of an eventual resize or reinit) */
+  if (init_type == FIRST_INIT) { step_mem->preallocated = SUNFALSE; }
```
gardner48: Related to the question above about preallocating then calling `ReInit`. If `preallocated` should be removed from those checks, then maybe this flag and `ALLOC_INIT` can be removed also.

Actually, this flag should probably go at the ARKODE level so the `step_init` call in `arkInitialSetup` can be skipped (and then the flag reset at the end of the first step) if it's already been called. This would fix the problem of calling the preallocation function only to have the stepper (e.g., MRIStep) free and reallocate the same memory in the first step.

doc/arkode/guide/source/Usage/User_callable.rst
```diff
@@ -5415,6 +5415,40 @@ Output all ARKODE solver parameters   :c:func:`ARKodeWriteParameters`
...
+
+   :retval ARK_SUCCESS: the function exited successfully.
+   :retval ARK_MEM_NULL:  ``arkode_mem`` was ``NULL``.
+   :retval ARK_MEM_FAIL:  a memory allocation failed.
+
+   .. warning::
+
+      This must be called **after** all other optional input routines have been called,
+      and **before** the first call to :c:func:`ARKodeEvolve`.  This routine should
+      be called at most once per ARKODE memory block.
```
gardner48, line 5448: What is the danger in calling this more than once? I think a user could call this function, then change some setting (e.g., switch to a method with more stages) and call it again to update the allocation size without an issue (well, at least in MRIStep where the `init` function frees data if it already exists).

src/arkode/arkode_erkstep.c
gardner48, line 388: If `ALLOC_INIT` is needed (see comment above)
```suggestion
  With initialization types ALLOC_INIT or FIRST_INIT this routine:
```

src/arkode/arkode_arkstep.c
```diff
@@ -920,6 +923,17 @@ int arkStep_GetGammas(ARKodeMem ark_mem, sunrealtype* gamma, sunrealtype* gamrat
...
+  For other initialization types, this routine:
+  - sets the relevant TakeStep routine based on the current
+    problem configuration
+  - checks for consistency between the system and mass matrix
+    linear solvers (if applicable)
+  - initializes and sets up the system and mass matrix linear
+    solvers (if applicable)
+  - initializes and sets up the nonlinear solver (if applicable)
+
   With initialization type FIRST_INIT this routine:
```
gardner48, line 923: Add `ALLOC_INIT` (if it's needed, see comments below in this file)
```suggestion
  With initialization type ALLOC_INIT or FIRST_INIT this routine additionally:
```

src/arkode/arkode_arkstep.c
```diff
@@ -1044,33 +1047,27 @@ int arkStep_Init(ARKodeMem ark_mem, SUNDIALS_MAYBE_UNUSED sunrealtype tout,
     {
       if (step_mem->Fe == NULL)
```
gardner48: This check can be removed since `arkAllocVecArray` will check if `Fe == NULL` (or updated to free existing data like in MRIStep)

src/arkode/arkode_arkstep.c
```diff
@@ -1044,33 +1047,27 @@ int arkStep_Init(ARKodeMem ark_mem, SUNDIALS_MAYBE_UNUSED sunrealtype tout,
...
           return (ARK_MEM_FAIL);
         }
       }
-      ark_mem->liw += step_mem->stages; /* pointers */
     }
 
     /*   Allocate Fi[0] ... Fi[stages-1] if needed */
     if (step_mem->implicit)
     {
       if (step_mem->Fi == NULL)
```
gardner48: This check can be removed since `arkAllocVecArray` will check if `Fi == NULL` (or updated to free existing data like in MRIStep)

src/arkode/arkode_erkstep.c
```diff
@@ -465,16 +465,13 @@ int erkStep_Init(ARKodeMem ark_mem, SUNDIALS_MAYBE_UNUSED sunrealtype tout,
   /*   Allocate F[0] ... F[stages-1] if needed */
   if (step_mem->F == NULL)
```
gardner48: This check can be removed since `arkAllocVecArray` will check if `F == NULL` (or updated to free existing data like in MRIStep)

src/arkode/arkode_forcingstep.c
```diff
@@ -90,11 +90,8 @@ static int forcingStep_Init(ARKodeMem ark_mem,
     return ARK_ILL_INPUT;
   }
 
-  /* immediately return if resize or reset */
-  if (init_type == RESIZE_INIT || init_type == RESET_INIT)
-  {
-    return ARK_SUCCESS;
-  }
+  /* immediately return if not called in FIRST_INIT mode */
+  if (init_type != FIRST_INIT) { return ARK_SUCCESS; }
```
gardner48: I suppose for methods that rely on other steppers the solution for now is for the user to call the preallocate function on each stepper separately. We should probably (in a later PR) add a SUNStepper function so this can just be called on the overall stepper and it calls the function for each stepper like we do with SUNStepper_Reset.

src/arkode/arkode_mristep.c
```diff
@@ -1263,7 +1275,7 @@ int mriStep_Init(ARKodeMem ark_mem, sunrealtype tout, int init_type)
   }
 
   /* Call linit (if it exists) */
-  if (step_mem->linit)
+  if (step_mem->linit && !step_mem->preallocated)
```
gardner48: Like in ARKStep, I think the `preallocated` check could be removed, so the `Init` functions here an below are called when preallocating and re-initializing later.

src/arkode/arkode_mristep.c
```diff
@@ -1286,76 +1298,85 @@ int mriStep_Init(ARKodeMem ark_mem, sunrealtype tout, int init_type)
...
-    if (ark_mem->hin == ZERO)
-    {
-      arkProcessError(ark_mem, ARK_ILL_INPUT, __LINE__, __func__,
-                      __FILE__, "Timestep adaptivity disabled, but missing user-defined fixed stepsize");
-      return (ARK_ILL_INPUT);
-    }
-  }
-  else
+  /*** Perform timestep adaptivity checks and initial setup (skip on ALLOC_INIT) ***/
+  if (init_type != ALLOC_INIT)
```
gardner48: I assume this guard is to avoid calling `mriStep_SlowRHS` in the alloc call and wait for the second time this is called in the firs step. Moving this to the `preallocated` flag to the ARKODE level would (in the current setup) require calling this at alloc time which is undesirable. Maybe `arkInitialSetup` should have a `arkComputeH0` function that by default is the usual method but MRI step could override that with its own function. That way there is a better separate between alloc actions and computations for the initial step.

src/arkode/arkode_mristep.c
```diff
@@ -2917,7 +2938,8 @@ int mriStep_TakeStepMERK(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPtr)
   /* for adaptive computations, reset the inner integrator to the beginning of this step */
   if (!ark_mem->fixedstep)
   {
-    retval = mriStepInnerStepper_Reset(step_mem->stepper, t0, ark_mem->ycur);
+    retval = mriStepInnerStepper_Reset(step_mem->stepper, ark_mem->tn,
```
gardner48: Match `tcur` with `ycur`
```suggestion
    retval = mriStepInnerStepper_Reset(step_mem->stepper, ark_mem->tcur,
```

src/arkode/arkode_sprkstep.c
```diff
@@ -414,6 +414,9 @@ int sprkStep_Init(ARKodeMem ark_mem, SUNDIALS_MAYBE_UNUSED sunrealtype tout,
         break;
       }
     }
+
+    /* Immediately return if called for allocation */
+    if (init_type == ALLOC_INIT) { return (ARK_SUCCESS); }
```
gardner48: If the `preallocated` flag is moved up to the ARKODE level so `init` is called once, this will need to be removed.

src/arkode/arkode_forcingstep.c
gardner48: Should LSRKStep and SplittingStep have a similar update?

gardner48 (commented):
One main question, should the `preallocated` flag be moved to the ARKODE level so the initialization function is conditionally called once instead of calling it again in the first step?

doc/arkode/guide/source/Usage/User_callable.rst
drreynolds: Multiple current steppers do not support changing the number of internal stages (e.g., ERKStep and ARKStep).  While both could be upgraded to support this feature, I see it as being only tangentially related to this PR, so I haven't implemented it here, and thus I added this warning to ensure that the data allocation is only called once (as would happen prior to this PR).

I agree that this might be a nice future feature for all steppers to support; at that time this sentence could be removed.

src/arkode/arkode_forcingstep.c
drreynolds: Is the question whether LSRKStep and SplittingStep should add the line
   `if (init_type != FIRST_INIT) { return ARK_SUCCESS; }`

If so, then since neither of those allocate any stage-related data, and all their other allocations are already guarded by things like `if (step_mem->cvals == NULL)`, then there is no harm in those `_Init` calls occurring twice, so I didn't add this line.  

Taking a step back, once I figure out how to follow your suggestion of moving the `preallocated` checks to the ARKODE level, I should be able to even remove this line from the existing locations where I'd added it.

gardner48: I didn't go back to check if those steppers had guarded allocations or not, so it sounds like this isn't needed in those cases and this could be a moot item if the `preallocated` flag can move to the ARKODE level

drreynolds:
I believe all PR comments have been resolved.

include/arkode/arkode.h
```diff
@@ -160,6 +160,8 @@ extern "C" {
 
 #define ARK_DEE_FAIL -59
 
+#define ARK_STEP_H0_FAIL -60
```
gardner48, line 163: Add this to the table to return codes in the docs

src/arkode/arkode_arkstep.c
```diff
@@ -1110,24 +1091,19 @@ int arkStep_Init(ARKodeMem ark_mem, SUNDIALS_MAYBE_UNUSED sunrealtype tout,
     }
 
     /* Allocate workspace for MRI forcing -- need to allocate here as the
-       number of stages may not bet set before this point and we assume
-       SetInnerForcing has been called before the first step i.e., methods
-       start with a fast integration */
-    if (step_mem->expforcing || step_mem->impforcing)
+       number of stages may not bet set before this point */
```
gardner48:
```suggestion
       number of stages may not be set before this point */
```

src/arkode/arkode_erkstep.c
```diff
@@ -493,24 +481,19 @@ int erkStep_Init(ARKodeMem ark_mem, SUNDIALS_MAYBE_UNUSED sunrealtype tout,
   }
 
   /* Allocate workspace for MRI forcing -- need to allocate here as the
-     number of stages may not bet set before this point and we assume
-     SetInnerForcing has been called before the first step i.e., methods
-     start with a fast integration */
-  if (step_mem->nforcing > 0)
+     number of stages may not bet set before this point */
```
gardner48:
```suggestion
     number of stages may not be set before this point */
```

src/arkode/arkode_mristep.c
```diff
@@ -1870,7 +1891,7 @@ int mriStep_TakeStepMRIGARK(ARKodeMem ark_mem, sunrealtype* dsmPtr, int* nflagPt
   SUNLogInfo(ARK_LOGGER, "begin-stages-list",
              "stage = 0, stage type = %d, tcur = " SUN_FORMAT_G, MRISTAGE_FIRST,
              ark_mem->tcur);
-  SUNLogExtraDebugVec(ARK_LOGGER, "slow stage", ark_mem->ycur, "z_0(:) =");
+  SUNLogExtraDebugVec(ARK_LOGGER, "slow stage", ark_mem->yn, "z_0(:) =");
```
gardner48: Bad merge?
```suggestion
  SUNLogExtraDebugVec(ARK_LOGGER, "slow stage", ark_mem->ycur, "z_0(:) =");
```

test/unit_tests/arkode/CXX_serial/ark_test_prealloc_arkstep.cpp
```diff
@@ -0,0 +1,150 @@
+/* -----------------------------------------------------------------------------
+ * Programmer(s): Daniel R. Reynolds @ UMBC
+ * based on test_logging_arkode_arkstep.cpp by David J. Gardner @ LLNL
+ * -----------------------------------------------------------------------------
+ * SUNDIALS Copyright Start
+ * Copyright (c) 2025, Lawrence Livermore National Security,
```
gardner48:
```suggestion
 * Copyright (c) 2025-2026, Lawrence Livermore National Security,
```

test/unit_tests/arkode/CXX_serial/ark_test_prealloc_erkstep.cpp
```diff
@@ -0,0 +1,135 @@
+/* -----------------------------------------------------------------------------
+ * Programmer(s): Daniel R. Reynolds @ UMBC
+ * based on test_logging_arkode_erkstep.cpp by David J. Gardner @ LLNL
+ * -----------------------------------------------------------------------------
+ * SUNDIALS Copyright Start
+ * Copyright (c) 2025, Lawrence Livermore National Security,
```
gardner48:
```suggestion
 * Copyright (c) 2025-2026, Lawrence Livermore National Security,
```

test/unit_tests/arkode/CXX_serial/ark_test_prealloc_forcingstep.cpp
```diff
@@ -0,0 +1,170 @@
+/* -----------------------------------------------------------------------------
+ * Programmer(s): Daniel R. Reynolds @ UMBC
+ * based on test_logging_arkode_forcingstep.cpp by David J. Gardner @ LLNL
+ * -----------------------------------------------------------------------------
+ * SUNDIALS Copyright Start
+ * Copyright (c) 2025, Lawrence Livermore National Security,
```
gardner48:
```suggestion
 * Copyright (c) 2025-2026, Lawrence Livermore National Security,
```

test/unit_tests/arkode/CXX_serial/ark_test_prealloc_lsrkstep.cpp
```diff
@@ -0,0 +1,134 @@
+/* -----------------------------------------------------------------------------
+ * Programmer(s): Daniel R. Reynolds @ UMBC
+ * based on test_logging_arkode_lsrkstep.cpp by David J. Gardner @ LLNL
+ * -----------------------------------------------------------------------------
+ * SUNDIALS Copyright Start
+ * Copyright (c) 2025, Lawrence Livermore National Security,
```
gardner48:
```suggestion
 * Copyright (c) 2025-2026, Lawrence Livermore National Security,
```

test/unit_tests/arkode/CXX_serial/ark_test_prealloc_mristep.cpp
```diff
@@ -0,0 +1,185 @@
+/* -----------------------------------------------------------------------------
+ * Programmer(s): Daniel R. Reynolds @ UMBC
+ * based on test_logging_arkode_mristep.cpp by David J. Gardner @ LLNL
+ * -----------------------------------------------------------------------------
+ * SUNDIALS Copyright Start
+ * Copyright (c) 2025, Lawrence Livermore National Security,
```
gardner48:
```suggestion
 * Copyright (c) 2025-2026, Lawrence Livermore National Security,
```

test/unit_tests/arkode/CXX_serial/ark_test_prealloc_splittingstep.cpp
```diff
@@ -0,0 +1,170 @@
+/* -----------------------------------------------------------------------------
+ * Programmer(s): Daniel R. Reynolds @ UMBC
+ * based on test_logging_arkode_splittingstep.cpp by David J. Gardner @ LLNL
+ * -----------------------------------------------------------------------------
+ * SUNDIALS Copyright Start
+ * Copyright (c) 2025, Lawrence Livermore National Security,
```
gardner48:
```suggestion
 * Copyright (c) 2025-2026, Lawrence Livermore National Security,
```

test/unit_tests/arkode/CXX_serial/ark_test_prealloc_sprkstep.cpp
```diff
@@ -0,0 +1,124 @@
+/* -----------------------------------------------------------------------------
+ * Programmer(s): Daniel R. Reynolds @ UMBC
+ * based on test_logging_arkode_sprkstep.cpp by David J. Gardner @ LLNL
+* -----------------------------------------------------------------------------
+ * SUNDIALS Copyright Start
+ * Copyright (c) 2025, Lawrence Livermore National Security,
```
gardner48, line 6:
```suggestion
 * Copyright (c) 2025-2026, Lawrence Livermore National Security,
```

gardner48 (changes requested):
A few trivial items, otherwise this looks good

gardner48 (approved):

PR #888, opened 2026-03-24
Bugfix: Fix quadrature initialization in IDAS

Steven-Roberts:
When IDAS takes its first time step, it calls the quadrature RHS' to fill in `ida_phiQ[1]` and `ida_phiQS[1]` vectors. When IDAS is reinitialized, it does not do this due to some logic errors. This caused a couple issues

- On the first step after reinitialization, the quadrature predictor is incorrect, so the quadrature error estimate tended to be a significant overestimate. Since the first time step is quite conservative, this usually didn't matter.
- When recomputing forward to fill in missing data between checkpoints, IDAS could produce very slightly different results than the original forward pass. This can cause a different number of time steps and eventually a segfault as found in a GridKit example (CC @pelesh)


pelesh:
Thank you, @Steven-Roberts, for tracing this issue down and identifying the root cause. This has been showing up in [GridKit](https://github.com/ORNL/GridKit) tests nondeterministically. See [here](https://github.com/ORNL/GridKit/issues/3), for example. Changes in simulation configuration would make these issues come and go.

pelesh (commented):
I ran preliminary tests and this fix seems to address the issue we were seeing in GridKit. I need to run more tests to confirm.

Steven-Roberts:
After a closer investigation, it seems the problem was the quadrature predictor was computed when `IDACalcIC` was called. At that point, it's possible `IDAQuadInit` had not even been called yet, so `phiQ` and `phiQS` would be left 0 initialized. Even if `IDAQuadInit` had been called, the quadrature predictor would use the inconsistent initial conditions.

drreynolds (approved):
LGTM (aside from the fact that the CI failures need to be addressed).

doc/shared/RecentChanges.rst
```diff
@@ -21,4 +21,9 @@ Fixed minor bug in reporting the maximum number of stages in
 Removed duplicate logging output that would cause the Python logging tools to fail
 with a repeated key error.
 
+Fixed a bug where IDAS would incorrectly compute the quadrature predictor when
+IDACalcIC was used. In some cases, the lead to an inconsistent solution in the
```
```suggestion
IDACalcIC was used. In some cases, this lead to an inconsistent solution in the
```

CHANGELOG.md
```diff
@@ -117,6 +117,11 @@ Fixed a bug where passing an empty string to
 corresponding logging stream ([Issue
 #844](https://github.com/llnl/sundials/issues/844)).
 
+Fixed a bug where IDAS would incorrectly compute the quadrature predictor when
+IDACalcIC was used. In some cases, the lead to an inconsistent solution in the
```
```suggestion
IDACalcIC was used. In some cases, this lead to an inconsistent solution in the
```

src/idas/idas.c
```diff
@@ -5184,6 +5154,70 @@ int IDAInitialSetup(IDAMem IDA_mem)
...
+ * IDAQuadSetup
+ *
+ * This routine is called by IDASolve once at the first step. It
+ * fills in phiQ[1] and phiQS[1] since they are not provided by
+ * the user. It is important this is NOT done in the IDAInitialSetup
+ * routine because that can be called just before finding consistent
+ * initial conditions. If quadrature is not set by that point, that
+ * would incorrectly use phiQ[1] = 0 and phiQS[1] = 0. Even if
+ * IDAQuadInit is called before IDACalcIC, phiQ[1] and phiQS[1] would
+ * be evaluated at the inconsistent initial condition.
```
gardner48: I think the wording here is a little unclear e.g., it's not clear if "that point" refers to the initial step or the IC computation. Maybe use something like the following:
```suggestion
 * This routine is called by IDASolve once at the first step. It
 * fills in phiQ[1] and phiQS[1] since they are not provided by
 * the user. It is important this is NOT done in IDAInitialSetup as
 * IDACalcIC will call IDAInitialSetup in which case inconsistent initial
 * conditions will be used to compute phiQ[1] and phiQS[1] (if
 * IDAQuadInit is called BEFORE IDACalcIC) or phiQ[1] and phiQS[1]
 * will not be initialized (if IDAQuadInit is called AFTER IDACalcIC).
```

gardner48 (changes requested):
Minor comment updates, otherwise this looks good

gardner48 (approved):

PR #889, opened 2026-03-25
Bugfix: Fix some python bindings to functions in ARKODE that should allow None

balos1:


bindings/sundials4py/arkode/arkode_arkstep_generated.hpp
```diff
@@ -10,8 +10,33 @@ m.def("ARKStepSetImplicit", ARKStepSetImplicit, nb::arg("arkode_mem"));
...
+                                            Bi_adapt_default_null,
+                                            Be_adapt_default_null);
+      return lambda_result;
+    };
+
+    return ARKStepSetTables_adapt_optional_arg_with_default_null(arkode_mem, q,
+                                                                 p, Bi, Be);
+  },
+  nb::arg("arkode_mem"), nb::arg("q"), nb::arg("p"),
+  nb::arg("Bi").none() = nb::none(), nb::arg("Be").none() = nb::none());
```
gardner48, line 39: Is it meaningful that these use `= nb::none()`? I don't recall seeing that elsewhere.

gardner48 (approved):
A minor question, otherwise this looks good

bindings/sundials4py/arkode/arkode_arkstep_generated.hpp
balos1: Its just an alternate syntax (`.none()` and `= nb::none()` do the same thing) albeit redundant here.

PR #892, opened 2026-03-27
Feature: add functions to SUNLogger API to allow user to provide their own file pointer

balos1:


bindings/sundials4py/sundials/sundials_logger_generated.hpp
```diff
@@ -83,15 +83,27 @@ m.def(
 m.def("SUNLogger_SetErrorFilename", SUNLogger_SetErrorFilename,
       nb::arg("logger"), nb::arg("error_filename"));
 
+m.def("SUNLogger_SetErrorFile", SUNLogger_SetErrorFile, nb::arg("logger"),
+      nb::arg("error_fp"));
```
gardner48: Does this need `.none()` (and similarly below)?

gardner48 (changes requested):
One python question and it would be good to add a unit tests for these functions as well.

gardner48:
@balos1 Could you add a unit test for these functions as well?

gardner48 (approved):

PR #896, opened 2026-04-02
Maintenance: Remove unreachable code (duplicated `return` statements)

vil02:
This change removes unreachable `return` statements. There are no other occurrences of such problem.

For such small change I decided not to update the change logs. If such update would be beneficial, please inform me.


gardner48 (approved):
LGTM

drreynolds (approved):

PR #902, opened 2026-04-03
Bugfix: Uninitialzed-istage-fix

drreynolds:
The recent group of PRs added a diagnostic variable to SplittingStep, that was not initialized during construction.  This PR fixes that.

### DIFF
--- a/src/arkode/arkode_splittingstep.c
+++ b/src/arkode/arkode_splittingstep.c
@@ -675,6 +675,7 @@ void* SplittingStepCreate(SUNStepper* steppers, int partitions, sunrealtype t0,
   step_mem->steppers          = NULL;
   step_mem->n_stepper_evolves = NULL;
   step_mem->coefficients      = NULL;
+  step_mem->istage            = 0;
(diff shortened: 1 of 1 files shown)


gardner48 (approved):

Steven-Roberts:
This is still giving a compiler warning for me, so don't merge quite yet

gardner48:
> This is still giving a compiler warning for me, so don't merge quite yet

The same initialization is also needed in `ForcingStepCreate` maybe that's the issue

drreynolds:
That PR never added a variable to ForcingStep.

Steven-Roberts:
On gcc 8.5.0 I get the warning
```
arkode_splittingstep.c:483:27: warning: 'step_mem' may be used uninitialized in this function [-Wmaybe-uninitialized]
     *istage     = step_mem->istage;
                   ~~~~~~~~^~~~~~~~
```
But with gcc 14.2.1, there's no warning.

So the change in the this PR is probably needed, but it seems old versions of GCC are giving false positives on the warning.

PR #905, opened 2026-04-08
Fix issues identified by cppcheck

### LINKED ISSUE #899: Usless assigments to argument
In these places:
https://github.com/llnl/sundials/blob/178eae817c7a5338e5f572d3a62d8cb6b4728fce/src/sundials/sundials_direct.c#L182
https://github.com/llnl/sundials/blob/178eae817c7a5338e5f572d3a62d8cb6b4728fce/src/sundials/sundials_direct.c#L260
https://github.com/llnl/sundials/blob/178eae817c7a5338e5f572d3a62d8cb6b4728fce/src/sundials/sundials_direct.c#L266
https://github.com/llnl/sundials/blob/178eae817c7a5338e5f572d3a62d8cb6b4728fce/src/sundials/sundials_nvector.c#L1080
`NULL` is assigned to a local copy of the pointer - this assignment has no effect for the caller (cf. [example at godbolt](https://godbolt.org/z/4Tda4Y5xM)).

Steven-Roberts:
cppcheck flagged a few memory leaks in the unlikely case of a malloc failing in a constructor, and codex did a decent job at fixing them. These fixes aren't so important, but they make the cppcheck report shorter so it will be easier to identify more important issues in the future.

### DIFF
--- a/src/cvodes/cvodea.c
+++ b/src/cvodes/cvodea.c
@@ -692,22 +692,23 @@ int CVodeCreateB(void* cvode_mem, int lmmB, int* which)
   }
   ca_mem = cv_mem->cv_adj_mem;
 
-  /* Create and set a new CVODES object for the backward problem */
-
-  cvodeB_mem = CVodeCreate(lmmB, cv_mem->cv_sunctx);
-  if (cvodeB_mem == NULL)
-  {
-    cvProcessError(cv_mem, CV_MEM_FAIL, __LINE__, __func__, __FILE__,
-                   MSGCV_MEM_FAIL);
-    return (CV_MEM_FAIL);
-  }
-
   /* Allocate space for new CVodeBMem object */
 
   new_cvB_mem = NULL;
   new_cvB_mem = (CVodeBMem)malloc(sizeof(struct CVodeBMemRec));
   if (new_cvB_mem == NULL)
   {
+    cvProcessError(cv_mem, CV_MEM_FAIL, __LINE__, __func__, __FILE__,
+                   MSGCV_MEM_FAIL);
+    return (CV_MEM_FAIL);
+  }
+
+  /* Create and set a new CVODES object for the backward problem */
+
+  cvodeB_mem = CVodeCreate(lmmB, cv_mem->cv_sunctx);
+  if (cvodeB_mem == NULL)
+  {
+    free(new_cvB_mem);
     cvProcessError(cv_mem, CV_MEM_FAIL, __LINE__, __func__, __FILE__,
                    MSGCV_MEM_FAIL);
     return (CV_MEM_FAIL);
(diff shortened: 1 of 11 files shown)


src/kinsol/kinsol.c
drreynolds, line 1: I don't quite understand the changes in this file -- is it that the KIN_PROFILER lives inside `kin_mem`, and so we cannot mark the function end after freeing?

drreynolds (commented):
Aside from my one question below, this LGTM

src/kinsol/kinsol.c
Steven-Roberts: Yes, exactly 
https://github.com/llnl/sundials/blob/0a1024b7fb902a5eae3142adf8c13561c6989543/src/kinsol/kinsol_impl.h#L59

balos1 (dismissed):

src/cvodes/cvodea.c
```diff
@@ -692,21 +692,21 @@ int CVodeCreateB(void* cvode_mem, int lmmB, int* which)
...
   }
 
-  /* Create and set a new CVODES object for the backward problem */
+  /* Allocate space for new CVodeBMem object */
 
-  cvodeB_mem = CVodeCreate(lmmB, cv_mem->cv_sunctx);
-  if (cvodeB_mem == NULL)
+  new_cvB_mem = NULL;
+  new_cvB_mem = (CVodeBMem)malloc(sizeof(struct CVodeBMemRec));
+  if (new_cvB_mem == NULL)
```
gardner48: Switching the ordering creates a different potential leak. If the `malloc` fails, we should call `CVodeFree` on `cvodeB_mem` below. I think we should keep the original ordering (create `new_cvB_mem` first since we'll attach other objects to it) and add `free(new_cvB_mem)` if `CVodeCreate` fails.

src/idas/idaa.c
```diff
@@ -682,18 +682,18 @@ int IDACreateB(void* ida_mem, int* which)
...
                     MSG_MEM_FAIL);
     return (IDA_MEM_FAIL);
   }
 
-  /* Allocate the IDAMem struct needed by this backward problem. */
-  ida_memB = IDACreate(IDA_mem->ida_sunctx);
-  if (ida_memB == NULL)
+  /* Allocate a new IDABMem struct. */
+  new_IDAB_mem = (IDABMem)malloc(sizeof(struct IDABMemRec));
+  if (new_IDAB_mem == NULL)
```
gardner48: Same comment as in CVODES

gardner48 (changes requested):

gardner48:
Also note in the change log / recent changes that potential memory leaks were fixed if a memory allocation fails.

gardner48 (approved):

PR #907, opened 2026-04-11
idas: wire NLSstg LSolveFn in IDASetNonlinearSolverSensStg to prevent NULL crash

OfficialDelta:
## Summary

`IDASetNonlinearSolverSensStg()` does not wire `LSolveFn` (or `LSetupFn`) on the replacement `NLSstg`. These callbacks are only set later by `idaNlsInitSensStg()`, which runs inside `IDAInitialSetup()`. If the staggered sensitivity NLS is replaced **after** `IDACalcIC()` has already run `IDAInitialSetup()` (setting `ida_SetupDone = TRUE`), the subsequent `IDASolve()` call skips `IDAInitialSetup()`, and Newton iterations on the new `NLSstg` dereference a NULL `LSolveFn`, causing a segmentation fault.

## Root Cause

1. `IDACalcIC()` calls `IDAInitialSetup()`, which calls `idaNlsInitSensStg()` — this correctly sets `LSolveFn = idaNlsLSolveSensStg` on `NLSstg` and marks `ida_SetupDone = TRUE`.
2. User calls `IDASetNonlinearSolverSensStg(ida_mem, NLS_new)` to replace `NLSstg` with a fresh `SUNNonlinSol_NewtonSens`. The new NLS has `LSolveFn = NULL`.
3. `IDASolve()` checks `ida_SetupDone` and **skips** `IDAInitialSetup()` entirely.
4. The Newton solver on `NLSstg` calls `LSolveFn(...)` → NULL pointer dereference → SIGSEGV.

The same crash pattern can occur even without explicit user NLS replacement, in any SUNDIALS configuration where `idaNlsInitSensStg()` has not yet run when the first Newton iteration fires on the staggered NLS.

## Fix

Wire `LSolveFn` and `LSetupFn` eagerly in `IDASetNonlinearSolverSensStg()`, immediately after setting up the NLS, if the linear solver is already attached. This mirrors the existing pattern in `idaNlsInitSensStg()` but ensures the callbacks are present regardless of whether `IDAInitialSetup()` runs again.

The fix is safe when the linear solver is not yet set (`ida_lsolve == NULL`): in that case the early wiring is skipped and `idaNlsInitSensStg()` handles it during the normal init path.

## Minimal Reproduction

A 10-species linear chain ODE with staggered FSA (Ns=10) demonstrates the crash:

```c
// After IDACalcIC (sets ida_SetupDone=TRUE):
SUNNonlinearSolver NLS_new = SUNNonlinSol_NewtonSens(Ns, delta, sunctx);
IDASetNonlinearSolverSensStg(ida_mem, NLS_new);  // LSolveFn = NULL on new NLS
IDASolve(ida_mem, t_final, &tret, yy, yp, IDA_NORMAL);  // SIGSEGV
```

With the fix applied, the same sequence completes without crash, and FSA gradients match finite-difference gradients with cosine similarity > 0.999.

## Testing

- Validated on SUNDIALS 7.6.0 and 7.7.0 conda builds
- 10-species system: crash reproduced without fix, FSA/FD cosine = 0.9994 with fix
- 764K-species production system: no SEGFAULT with fix applied

## Files Changed

- `src/idas/idas_nls_stg.c`: Added early `LSolveFn`/`LSetupFn` wiring in `IDASetNonlinearSolverSensStg()`


Steven-Roberts:
Thanks for the PR. The setup of `ida_SetupDone = TRUE` when `IDACalcIC` is called which skips `IDAInitialSetup` later is the culprit of #888 as well. I suspect there are more that these two situations where we can run into initialization issues, so we probably need some more testing of setters before and after `IDACalcIC` is called.

drreynolds (commented):
I think this change looks great.  However, we generally orient pull requests to the `develop` branch instead of `main` -- please direct this PR.

gardner48:
This fix works with the native nonlinear solver options because they don't perform any essential actions when their initialization functions are called and the initialize call is skipped if the nonlinear solver is attached after computing the initial conditions. However, this may not be true for user-supplied solvers and will change soon for the native solvers with update to reduce the overall memory footprint. Bumping this to the v8.0.0 release, so we can better address those cases.

PR #908, opened 2026-04-13
Maintenance: Switch to ruff

balos1:
Ruff is a much faster formatter than black: https://github.com/astral-sh/ruff.

### DIFF
--- a/examples/utilities/plot_data_2d.py
+++ b/examples/utilities/plot_data_2d.py
@@ -206,7 +206,6 @@ def main():
         plot_contour_ani(args, info, time, xvals, yvals, zdata)
 
     if args.plottype == "slice":
-
         # slice data
         if (args.yslice > -1) and (args.yslice < info["ny"]):
             svals = xvals
@@ -234,7 +233,6 @@ def main():
             plot_slice_time(args, info, time, svals, sdata, hlabel, suffix)
 
     if args.plottype == "point":
-
         # point data
         pdata = zdata[:, args.point[1], args.point[0], :]
         suffix = " at x = {:.4f}, y = {:.4f}".format(xvals[args.point[0]], yvals[args.point[1]])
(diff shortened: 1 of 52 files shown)


examples/utilities/plot_data_2d.py
```diff
@@ -528,7 +519,7 @@ def read_data(args, info, subdomains):
             print(
                 "WARNING: subdomain",
                 str(idx),
-                "has an incorrect number of" "output times (",
+                "has an incorrect number ofoutput times (",
```
gardner48:
```suggestion
                "has an incorrect number of output times (",
```

scripts/spack/packages/sundials/package.py
```diff
@@ -407,7 +407,7 @@ def filter_compilers(self):
 
         cxx_files = [
             "arkode/CXX_parallel/Makefile",
-            "arkode/CXX_serial/Makefile" "cvode/cuda/Makefile",
+            "arkode/CXX_serial/Makefilecvode/cuda/Makefile",
```
gardner48:
```suggestion
            "arkode/CXX_serial/Makefile",
            "cvode/cuda/Makefile",
```

gardner48 (changes requested):
I couple of typos highlight by the new formatting. Otherwise this looks good.

gardner48 (commented):
The changes look good but the python interface generation workflow is failing because [sundials4py-generate](https://github.com/sundials-codes/sundials4py-generate) needs to switch to ruff also

balos1:
> The changes look good but the python interface generation workflow is failing because [sundials4py-generate](https://github.com/sundials-codes/sundials4py-generate) needs to switch to ruff also

Switched in the PR https://github.com/sundials-codes/sundials4py-generate/pull/1. After it is merged, I will update the pinned commit here.

gardner48 (approved):

PR #854, opened 2026-03-05
Feature: SUNNonlinearSolver_Auto module that automatically switches between newton and fixed-point based on stiffness

balos1:
This implements the algorithm from https://doi.org/10.1007/BF01933714 and enables its use from CVODE/CVODES. Usage from ARKODE is a bit more complicated if we want to be true to what the paper does (use the max convergence rate estimate across all stages).  

The results wrt the benefits of using this module are mixed and we may still need to explore it more. I am opening the PR now so the implementation can be reviewed and we can decide on if it should be merged.

Answers PR: https://github.com/sundials-codes/answers/pull/76

PR #915, opened 2026-04-16
Skill: add skill to aid developers in addressing pr comments

balos1:
I started making this while handling PR #854. I think it would have been smoother if I had thought through the skill beforehand. Here is what I suggest:

1.After received PR comments, go through each one and leave a reply which an agent could use to decide if the comment should be addressed and how. If the comment you are replying to is a suggestion that can just be accepted, then just reply something like "LGTM" or "OK". If it is open ended, add some detail about how to address it. If you don't want the suggestion to be implemented, then say something along those lines. You dont have to reply to every comment, but it does help steer the agent.
2. Use the skill in your agent. In codex, you do something like
```
> $devs-address-pr-comments address <PR URL>
```
3. Wait for the agent
4. Review the markdown file in `.agents/scratch` for explanations on what the agent did
5. The agent will have made multiple commits, review the diff with a command `git diff origin/<branch>..HEAD`
6. Push


drreynolds (approved):
LGTM

gardner48 (approved):

PR #916, opened 2026-04-17
Feature: logging message handler

gardner48:
Add user-supplied functions to queue and flush log messages
Add an agent skill for creating python wrappers for functions that take callback functions


.agents/skills/python-wrappers/SKILL.md
balos1, line 1: This is great! There may be a reduced version of this that generates the same output with less tokens, but this is a pretty complex task. E.g., some of the example may be unnecessary, but I think lets leave them for now and see how agents handle it.

gardner48: I used this skill to add the wrapper for `SUNLoggerSetQueueAndFlushMsgFns` and it made all the needed modifications to the `SUNLogger` files and created/modified all the necessary Python files. The only thing I had to fix manually was adding the `else if (queue_fn && !flush_fn)` case in the wrapper.

CLAUDE.md
balos1, line 1: I don't think we want to delete this symlink. Claude looks for this file, not an AGENTS.md.

.gitignore
```diff
@@ -9,6 +9,12 @@
 /tarballs
 /thirdparty
 
+# Agent symlinks
+# links to .agents
+.claude
+# links to AGENTS.md
+CLAUDE.md
```
balos1: I don't think we want to ignore these symlinks and we should commit the .claude to .agents symlink.

CLAUDE.md
gardner48: I was thinking that's something Claude users could add locally so we don't have carry them around the repo. I had to add a `.claude` symlink to `.agents` to get it to pickup the skills.

.gitignore
gardner48: I don't think we need to clutter the top level of the repo with symlinks since this is something that can easily be done locally.

balos1: I think there are other ways we can cleanup the top level. E.g., we can move the .gitlab-ci.yml into .gitlab. We can move .readthedocs.yaml into doc/. If you dont have these Claude symlinks committed, then a workflow where you have an agent clone the repo to do something will be negatively effected, as the freshly cloned repo will not have these symlinks at first and a dev will have to jump in to add them. I.e., there is benefit to having these committed. We should commit them and find other ways to clean up. Hopefully, Anthropic gets on board with the standard directories soon enough.

src/sundials/sundials_logger.c
balos1, line 1: `SUNLogger_SetQueueAndFlushMsgFns` reads like a full override for the queue/flush behavior, but the implementation still applies `sunLoggerIsOutputRank()` before either callback is invoked. In practice that means ranks other than `output_rank` never reach the user callback at all, so a custom handler cannot aggregate or reroute messages from all ranks even though the callback API now exposes `rank`.

If the intent is to let users define their own routing policy, the output-rank filter needs to move into the default `sunQueueLogMessage` / `sunFlushLogMessage` implementations instead of wrapping the callback dispatch in `SUNLogger_QueueMsg` and `SUNLogger_Flush`. Otherwise the new API is much narrower than the signature and docs imply.

Agent: Codex (GPT-5)

test/unit_tests/sundials/test_sundials_logger.cpp
balos1, line 1: This test covers the happy path for installing a custom queue/flush pair, but it does not exercise the reset path. Since `SUNLogger_SetQueueAndFlushMsgFns(logger, NULL, NULL, NULL)` is part of the new public API surface, I would add a small follow-up assertion that resets the logger to defaults and verifies subsequent messages go back to the normal stderr/stdout destinations. That would protect the fallback path as well as the custom one.

Agent: Codex (GPT-5)

bindings/sundials4py/sundials/sundials_logger.cpp
balos1, line 1: This PR adds a new nanobind entry point, a new `logger->python` lifetime path, and two new wrapper trampolines, but there is no Python-side regression test with it. I would add a minimal `sundials4py` test that installs Python queue/flush callbacks, calls `SUNLogger_QueueMsg` / `SUNLogger_Flush`, and verifies the callbacks fire. A second assertion that passing `None` restores the default behavior would be even better.

Agent: Codex (GPT-5)

src/sundials/sundials_logger.c
gardner48: This is intentional, if a user wants to do message aggregation they can enable output from all ranks and their queue and flush functions can handle it.

balos1 (dismissed):

balos1 (approved):

PR #918, opened 2026-04-17
Maintenance: cleanup repository root files

balos1:
We will have to update the repository settings in the LC gitlab and in readthedocs to point them to these new locations once this merged. The rtd build job and gitlab jobs below will fail or not run until after that.


balos1:
We could also move the CONTRIBUTING.md if we renamed `doc/` to `docs/` (see https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors). I don't think we want to move it into .github as it is more broadly applicable.

gardner48:
> We could also move the CONTRIBUTING.md if we renamed `doc/` to `docs/` (see https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors). I don't think we want to move it into .github as it is more broadly applicable.

I think keeping `CONTRIBUTING.md` at the top-level is probably better so it's more visible

gardner48:
I reverted to docker directory and container workflow updates. These are superseded by #932.

gardner48 (approved):

PR #921, opened 2026-04-21
Bugfix: Log parsing duplicate key error

Steven-Roberts:
All of the integrators had logging keys duplicated within a step which caused an error here:
https://github.com/llnl/sundials/blob/0a1024b7fb902a5eae3142adf8c13561c6989543/tools/suntools/logs.py#L127-L129

This PR allows the duplicate key if the value is consistent and fixes one case in ARKODE where the `h_acc` key was used with two values.


tools/suntools/logs.py
```diff
@@ -124,9 +124,10 @@ def __repr__(self):
 
     def update(self, data):
         """Update the active dictionary"""
-        conflicts = self.stack[-1].keys() & data.keys()
+        top = self.stack[-1]
+        conflicts = [k for k in top.keys() & data.keys() if top[k] != data[k]]
```
gardner48, line 128: I'm not sure we should accept duplicate keys in any case. My concern is we accidentally reuse a key and its value changes in some cases but not in others so the parser will fail sometimes but not others.

doc/shared/RecentChanges.rst
```diff
@@ -15,4 +15,7 @@ Updated ``examples/cvode/petsc/cv_petsc_ex7.c`` to support PETSc 3.25.0.
 Fixed memory leaks in CVODES, IDAS, and KINSOL in the unlikely event of a failed
 ``malloc``.
 
+Fixed an error in the Python logging tools where repeated keys would cause an
+error.
```
```suggestion
Removed duplicate logging output that would cause the Python logging tools to fail with a repeated key error.
```

src/ida/ida.c
```diff
@@ -2514,7 +2514,7 @@ static int IDAStep(IDAMem IDA_mem)
   {
     SUNLogInfo(IDA_LOGGER, "begin-step-attempt",
                "step = %li, tn = " SUN_FORMAT_G ", h = " SUN_FORMAT_G
-               ", q = %d",
+               ", kk = %d",
```
gardner48: Is there an issue with leaving this as `q`? Using `q` makes the order key the same across the different integrators in the logging output.
```suggestion
               ", q = %d",
```

src/idas/idas.c
```diff
@@ -5926,7 +5926,7 @@ static int IDAStep(IDAMem IDA_mem)
   {
     SUNLogInfo(IDA_LOGGER, "begin-step-attempt",
                "step = %li, tn = " SUN_FORMAT_G ", h = " SUN_FORMAT_G
-               ", q = %d",
+               ", kk = %d",
```
gardner48: Is there an issue with leaving this as `q`? Using `q` makes the order key the same across the different integrators in the logging output.
```suggestion
               ", q = %d",
```

test/answers
gardner48, line 1: Open a PR in the answers repo with these updates

CHANGELOG.md
```diff
@@ -16,6 +16,9 @@ Updated `examples/cvode/petsc/cv_petsc_ex7.c` to support PETSc 3.25.0.
 Fixed memory leaks in CVODES, IDAS, and KINSOL in the unlikely event of a failed
 `malloc`.
 
+Fixed an error in the Python logging tools where repeated keys would cause an
+error.
```
```suggestion
Removed duplicate logging output that would cause the Python logging tools to fail with a repeated key error.
```

gardner48 (changes requested):

src/ida/ida.c
Steven-Roberts: I changed it to be consistent with the variable name and because elsewhere in the file, `kk` was used. I can revert to `q`, but perhaps in a separate PR it would be clearer to use `order` everywhere (possibly when we do https://github.com/llnl/sundials-private/issues/99)

gardner48: I was thinking the same thing and I we don't necessarily need to wait for updating the variable names before making the logging change.

Steven-Roberts: Are you suggesting I use `order` everywhere in this PR, then?

gardner48: No, I think we should do that separately from the fixes here.

gardner48 (changes requested):
The changes look good, but the updates in sundials-codes/answers#78 are missing the CVODE(S) changes and the switch back to `q` from `kk`

Steven-Roberts:
The Windows CI is failing because `mingw-w64-i686-gcc-fortran` seemingly disappeared from MINGW. I think it's related to https://github.com/msys2/MINGW-packages/pull/28751, but I'm not sure if this was intentional or has a simple workaround.

gardner48 (approved):

PR #922, opened 2026-04-22
Bugfix: SSP GetStageIndex

drreynolds:
During review for PR #819, we incorrectly revised the `max_stages` output for LSRKStep when using SSP methods.  Although the STS methods always evaluate the RHS at the end of the step by construction, the SSP methods do not.  Thus when calling `ARKodeGetStageIndex` with SSP methods returned "stage" values ranging from `0` to `max_stages-2` (instead of `max_stages-1`).  

This PR fixes that reporting error for SSP methods.


src/arkode/arkode_lsrkstep_io.c
```diff
@@ -786,8 +786,16 @@ int lsrkStep_GetStageIndex(ARKodeMem ark_mem, int* stage, int* max_stages)
...
+  if (step_mem->is_SSP)
+  {
+    *stage      = step_mem->istage;
+    *max_stages = step_mem->req_stages;
+  }
+  else
+  {
+    *stage      = step_mem->istage;
+    *max_stages = step_mem->req_stages + 1;
+  }
```
maggul:
```suggestion
  *stage      = step_mem->istage;
  *max_stages = step_mem->req_stages + (step_mem->is_SSP ? 0 : 1);
```

drreynolds:
@gardner48: I assembled an application-side fix for the Gkeyll interface to SUNDIALS, so this PR is ready to review.

gardner48 (changes requested):
The changes look good, but need to open a PR in the answers repo with the output updates

gardner48 (approved):

PR #938, opened 2026-05-22
Doc: Fix versions in added, changed, and deprecated notes

gardner48:
Consistently use the SUNDIALS version in added, changed, and deprecated notes. When the package version number differs from the SUNDIALS version number (ARKODE and IDAS) it is included as a parenthetical note after the SUNDIALS version number.


balos1:
Should we consider just syncing up the IDAS and ARKDOE versions with the rest of SUNDIALS? I see no real advantage of having them differ.

gardner48:
> Should we consider just syncing up the IDAS and ARKDOE versions with the rest of SUNDIALS? I see no real advantage of having them differ.

Yes, I think we should consider this and, if we do decide to sync the versions, make that change when we do a major release.

bangerth:
From a downstream perspective, that certainly would make sense. I don't think I would have expected sub-packages to have their own version numbering.

Thanks for addressing this so quickly!

gardner48:
The different versioning is a historical holdover from CVODE, IDA, and KINSOL predating SUNDIALS then IDAS and ARKODE getting added later. The versions have all been updated in lockstep with the SUNDIALS version for a while now, so moving to a single version number for everything would be reasonable.

drreynolds:
@gardner48: to clarify, is the plan to (a) merge this PR and then to subsequently bump the ARKODE and IDAS version numbers, or (b) abandon this PR and just bump the ARKODE and IDAS version numbers now?  I'm good with either, I just wanted to get caught up.

gardner48:
> @gardner48: to clarify, is the plan to (a) merge this PR and then to subsequently bump the ARKODE and IDAS version numbers, or (b) abandon this PR and just bump the ARKODE and IDAS version numbers now? I'm good with either, I just wanted to get caught up.

I'm thinking (a), we keep the version numbers as they are now but consider aligning the package versions with the sundials version in 8.0.0.

drreynolds (approved):

PR #948, opened 2026-05-30
Bugfix: Resize ERKStep before ARKodeEvolve

gardner48:
Fix a bug in ERKStep where calling `ARKodeResize` before `ARKodeEvolve` or `ARKodeInit` results in a segmentation fault.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,9 +43,12 @@ notes in the documentation to always use the SUNDIALS version number with the
 package version number as a parenthetical note when it differs from the SUNDIALS
 version number.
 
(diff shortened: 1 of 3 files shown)


drreynolds (commented):
This looks great; however, wouldn't the same "bug" would occur in both ARKStep and MRIStep, since we defer stage vector allocation until the first time step (unless `ARKodeInit` is called)?

gardner48:
> This looks great; however, wouldn't the same "bug" would occur in both ARKStep and MRIStep, since we defer stage vector allocation until the first time step (unless `ARKodeInit` is called)?

Both ARKStep and MRIStep already have this check so it's not a problem in those steppers

drreynolds (approved):

PR #950, opened 2026-06-02
Enable artifacts for failed Windows CI tests and fix Windows builds for runner image changes

Steven-Roberts:


drreynolds (dismissed):

balos1 (approved):

balos1:
I added commits that fix the windows builds. They started failing because of the windows-latest runners updating to Visual Studio 2026. For the VS build, we just needed to update the CMake generator that we use to be Visual Studio 2026. For the intel builds, I moved the runner image back to windows-2022. The intel oneapi compilers do not support Visual Studio 2026 until 2025.3.1 and the setup-fortran action we use to install the intel compilers only supports up to 2025.2 right now.

gardner48:
If you haven't already, we should make an issue or add a TODO item to update the intel builds when the setup-fortran action supports a newer version

PR #951, opened 2026-06-05
Bugfix: SUNDomEigEstimator_SetInitialGuess and Initialize order dependence 

maggul:
When `SUNDomEigEstimator_Initialize` is called after `SUNDomEigEstimator_SetInitialGuess`, initialize overwrites the user-set initial vector with the vector provided to the constructor. This PR fixes the bug by preserving the user-provided initial guess and making these two routines order-independent.

### DIFF
--- a/src/sundomeigest/arnoldi/sundomeigest_arnoldi.c
+++ b/src/sundomeigest/arnoldi/sundomeigest_arnoldi.c
@@ -134,40 +134,39 @@ SUNDomEigEstimator SUNDomEigEstimator_Arnoldi(N_Vector q, int kry_dim,
   DEE->content = content;
 
   /* Fill content */
-  content->ATimes         = NULL;
-  content->ATdata         = NULL;
-  content->V              = NULL;
-  content->q              = NULL;
-  content->rhs_linY       = NULL;
-  content->rhs_linT       = ZERO;
-  content->Fy             = NULL;
-  content->work           = NULL;
-  content->kry_dim        = kry_dim;
-  content->num_warmups    = DEE_NUM_OF_WARMUPS_ARNOLDI_DEFAULT;
-  content->num_iters      = 0;
-  content->num_ATimes     = 0;
-  content->init_guess_set = SUNFALSE;
-  content->warmup_to_tol  = SUNFALSE;
-  content->tol_warmup     = DEE_TOL_OF_WARMUPS_ARNOLDI_DEFAULT;
-  content->rhsfn          = NULL;
-  content->rhs_data       = NULL;
-  content->nfevals        = 0;
-  content->LAPACK_A       = NULL;
-  content->LAPACK_wr      = NULL;
-  content->LAPACK_wi      = NULL;
-  content->LAPACK_work    = NULL;
-  content->LAPACK_lwork   = 0;
-  content->LAPACK_arr     = NULL;
-  content->Hes            = NULL;
+  content->ATimes        = NULL;
+  content->ATdata        = NULL;
+  content->V             = NULL;
+  content->q             = NULL;
+  content->rhs_linY      = NULL;
+  content->rhs_linT      = ZERO;
+  content->Fy            = NULL;
+  content->work          = NULL;
+  content->kry_dim       = kry_dim;
+  content->num_warmups   = DEE_NUM_OF_WARMUPS_ARNOLDI_DEFAULT;
+  content->num_iters     = 0;
+  content->num_ATimes    = 0;
+  content->warmup_to_tol = SUNFALSE;
+  content->tol_warmup    = DEE_TOL_OF_WARMUPS_ARNOLDI_DEFAULT;
+  content->rhsfn         = NULL;
+  content->rhs_data      = NULL;
+  content->nfevals       = 0;
+  content->LAPACK_A      = NULL;
+  content->LAPACK_wr     = NULL;
+  content->LAPACK_wi     = NULL;
+  content->LAPACK_work   = NULL;
+  content->LAPACK_lwork  = 0;
+  content->LAPACK_arr    = NULL;
+  content->Hes           = NULL;
 
   /* Allocate content */
   content->q = N_VClone(q);
   SUNCheckLastErrNull();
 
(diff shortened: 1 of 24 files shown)


drreynolds (commented):
These look good to me, although the CHANGELOG and RecentChanges files may need a comment about the bugfix.

src/sundomeigest/arnoldi/sundomeigest_arnoldi.c
```diff
@@ -322,14 +323,21 @@ SUNErrCode SUNDomEigEstimator_Initialize_Arnoldi(SUNDomEigEstimator DEE)
...
+    normq = SUNRsqrt(normq);
 
-  N_VScale(ONE / normq, Arnoldi_CONTENT(DEE)->q, Arnoldi_CONTENT(DEE)->V[0]);
-  SUNCheckLastErr();
+    N_VScale(ONE / normq, Arnoldi_CONTENT(DEE)->q, Arnoldi_CONTENT(DEE)->V[0]);
+    SUNCheckLastErr();
+
+    /* Reset the initial guess flag to its default value */
+    Arnoldi_CONTENT(DEE)->init_guess_set = SUNFALSE;
+  }
```
gardner48: Instead of having to keep track of the `init_guess_set` flag, you could have the constructor perform this normalization and copy after allocating `V`.

src/sundomeigest/power/sundomeigest_power.c
```diff
@@ -284,14 +285,20 @@ SUNErrCode SUNDomEigEstimator_Initialize_Power(SUNDomEigEstimator DEE)
...
+    normq = SUNRsqrt(normq);
 
-  N_VScale(ONE / normq, PI_CONTENT(DEE)->q, PI_CONTENT(DEE)->V);
-  SUNCheckLastErr();
+    N_VScale(ONE / normq, PI_CONTENT(DEE)->q, PI_CONTENT(DEE)->V);
+    SUNCheckLastErr();
+
+    /* Reset the initial guess flag to its default value */
+    PI_CONTENT(DEE)->init_guess_set = SUNFALSE;
+  }
```
gardner48: Same here, this could be done in the constructor after allocating `V` so you don't need a flag to keep track of the state.

gardner48 (changes requested):

src/sundomeigest/power/sundomeigest_power.c
maggul: I could do that but prefered one less copy to reduce computation cost. I can switch to other option if you want.

gardner48: I prefer the alternative approach as keeping track of the flag only saves a copy in the case where the user provides an initial guess to the constructor then overwrites that guess by setting a new guess before calling estimate.

src/sundomeigest/arnoldi/sundomeigest_arnoldi.c
```diff
@@ -322,13 +322,13 @@ SUNErrCode SUNDomEigEstimator_Initialize_Arnoldi(SUNDomEigEstimator DEE)
...
-                                 Arnoldi_CONTENT(DEE)->q);
+  sunrealtype normV = N_VDotProd(Arnoldi_CONTENT(DEE)->V[0],
+                                 Arnoldi_CONTENT(DEE)->V[0]);
   SUNCheckLastErr();
 
-  normq = SUNRsqrt(normq);
+  normV = SUNRsqrt(normV);
 
-  N_VScale(ONE / normq, Arnoldi_CONTENT(DEE)->q, Arnoldi_CONTENT(DEE)->V[0]);
+  N_VScale(ONE / normV, Arnoldi_CONTENT(DEE)->V[0], Arnoldi_CONTENT(DEE)->V[0]);
```
gardner48: These computations should be moved to the constructor otherwise they duplicate the work done by a call to `SetInitialGuess` before calling `Initialize`

src/sundomeigest/power/sundomeigest_power.c
```diff
@@ -285,12 +285,12 @@ SUNErrCode SUNDomEigEstimator_Initialize_Power(SUNDomEigEstimator DEE)
...
-  sunrealtype normq = N_VDotProd(PI_CONTENT(DEE)->q, PI_CONTENT(DEE)->q);
+  sunrealtype normV = N_VDotProd(PI_CONTENT(DEE)->V, PI_CONTENT(DEE)->V);
   SUNCheckLastErr();
 
-  normq = SUNRsqrt(normq);
+  normV = SUNRsqrt(normV);
 
-  N_VScale(ONE / normq, PI_CONTENT(DEE)->q, PI_CONTENT(DEE)->V);
+  N_VScale(ONE / normV, PI_CONTENT(DEE)->V, PI_CONTENT(DEE)->V);
   SUNCheckLastErr();
```
gardner48: These computations should be moved to the constructor otherwise they duplicate the work done by a call to `SetInitialGuess` before calling `Initialize`

gardner48 (changes requested):

src/sundomeigest/power/sundomeigest_power.c
maggul: There is a follow-up PR coming soon that removes the vector `q` completely. Since we have already created a new work vector for the DQJ, I plan to reuse it everywhere `q` is currently needed; at that point, `q` serves as a work vector already. The follow-up PR also includes several other cleanups within the module.

If this is not urgent, I would prefer to make this change in the follow-up PR. Also, when we move this initialization to the constructor, the current power iteration initialization will no longer have any other role left. We should decide what to do with that as well. I believe all of these need much broader discussion and review process, and so should target the next release cycle.

gardner48: I don't think it matters if `q` is a temporary vector or not, this problem will exist in either case and computation will need to move to the constructor. We can close this PR but the fix is minor and I don't see a reason why it should wait for the next release. Moving this code to the constructor would result in the power iteration initialization function being just error checks which is fine and we could remove it in a future major release if desired.

gardner48 (approved):

PR #952, opened 2026-06-06
Feature: SUNDomEigEstimator Cleanup and remove extra vector memory allocation

maggul:
Removes the extra vector allocation from Arnoldi and improves the readability of the DEE estimator implementation.


doc/shared/sundomeigest/SUNDomEigEst_API.rst
```diff
@@ -247,6 +247,21 @@ instead of supplying a dummy routine.
       A :c:type:`SUNErrCode`.
 
 
+.. c:function:: SUNErrCode SUNDomEigEstimator_SetRhsAtLinearizationPoint(SUNDomEigEstimator DEE, N_Vector Fyt)
+
+   This *optional* function sets the value of the right-hand side function at the linearization point when using
+   :c:func:`SUNDomEigEstimator_SetRhs`.
```
drreynolds:
```suggestion
   :c:func:`SUNDomEigEstimator_SetRhs`.  When this is supplied, the *DEE* module 
   will use the value instead of re-evaluating the RHS at the linearization point.
```

drreynolds (commented):
I had one minor suggestion for the documentation.  Otherwise, I think that once you resolve the issues raised by Copilot, this should be good to merge.

src/arkode/arkode_lsrkstep_io.c
```diff
@@ -553,12 +553,12 @@ int LSRKStepSetDomEigEstimator(void* arkode_mem, SUNDomEigEstimator DEE)
...
 
-  /* Set the ATimes function for the DEE with A_data = arkode_mem */
-  retval = SUNDomEigEstimator_SetATimes(DEE, arkode_mem, lsrkStep_DQJtimes);
+  /* Set the RHS function for the DEE with user_data */
+  retval = SUNDomEigEstimator_SetRhs(DEE, ark_mem->user_data, step_mem->fe);
   if (retval != SUN_SUCCESS)
   {
     arkProcessError(ark_mem, ARK_DEE_FAIL, __LINE__, __func__, __FILE__,
-                    "SUNDomEigEstimator_SetATimes failed");
+                    "SUNDomEigEstimator_SetRhs failed");
```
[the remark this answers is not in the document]
maggul: @copilot, check the updates and let me know if your previously raised concerns persist.

maggul:
@copilot resolve the merge conflicts in this pull request

src/sundomeigest/arnoldi/sundomeigest_arnoldi.c
```diff
@@ -239,74 +257,94 @@ SUNErrCode SUNDomEigEstimator_SetRhsLinearizationPoint_Arnoldi(
...
+  SUNAssert(DEE, SUN_ERR_ARG_CORRUPT);
+  SUNAssert(Arnoldi_CONTENT(DEE), SUN_ERR_ARG_CORRUPT);
   SUNAssert(Arnoldi_CONTENT(DEE)->ATimes, SUN_ERR_ARG_CORRUPT);
   SUNAssert(Arnoldi_CONTENT(DEE)->V, SUN_ERR_ARG_CORRUPT);
-  SUNAssert(Arnoldi_CONTENT(DEE)->q, SUN_ERR_ARG_CORRUPT);
+
+  int* kry_dim     = &(Arnoldi_CONTENT(DEE)->kry_dim);
+  int* num_warmups = &(Arnoldi_CONTENT(DEE)->num_warmups);
+
+  if (*kry_dim < 2) { *kry_dim = DEE_KRYLOV_DIM_DEFAULT; }
```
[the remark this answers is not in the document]
maggul: is there any chance that `kry_dim` is changed after the contructor before or in the `SUNDomEigEstimator_Initialize_Arnoldi`? BTW, I agree that the following line is unncessary since we already check against 0 in the constructor and set the defult:
`if (*kry_dim < 2) { *kry_dim = DEE_KRYLOV_DIM_DEFAULT; }`

Do you instead suggest checking if `kry_dim < 3` in the constructor and set the default value otherwise?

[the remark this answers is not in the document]
maggul: @copilot, is there any chance that `kry_dim` is changed after the contructor before or in the `SUNDomEigEstimator_Initialize_Arnoldi`? BTW, I agree that the following line is unncessary since we already check against 0 in the constructor and set the defult:
`if (*kry_dim < 2) { *kry_dim = DEE_KRYLOV_DIM_DEFAULT; }`

Do you instead suggest checking if `kry_dim < 3` in the constructor and set the default value otherwise?

maggul:
PR passes all tests now and ready for your reviews.

doc/shared/sundomeigest/SUNDomEigEst_API.rst
```diff
@@ -230,6 +230,22 @@ instead of supplying a dummy routine.
...
+
+   This *optional* function provides a :c:type:`SUNPreRhsFn` function for performing
+   preprocessing of the right-hand side function, as well as a ``void*`` pointer to a data structure
+   used by this routine, to the dominant eigenvalue estimator.
+
+   **Arguments:**
+
+      * *DEE* -- a SUNDomEigEstimator object.
+      * *prerhs_fn_data* -- pointer to structure for ``prerhs_fn``.
+      * *prerhs_fn* -- function pointer to perform preprocessing of the right-hand side evaluations.  This is typically the same as the problem-defining function supplied to CVODE or ARKODE.
```
drreynolds: The `prerhs_fn` is called *before* the problem-defining function supplied to CVODE or ARKODE, so this comment is incorrect.  A more appropriate description of this function and its arguments would be:
```suggestion
   This *optional* function provides a :c:type:`SUNPreRhsFn` function for performing
   preprocessing of the right-hand side function, as well as a ``void*`` pointer to a data structure
   used by this routine, to the dominant eigenvalue estimator.  This *advanced* feature is similar to the ARKODE routine :c:func:`ARKPreRhsFn` -- we recommend that users read the warnings for that function before using this optional routine.

   **Arguments:**

      * *DEE* -- a SUNDomEigEstimator object.
      * *prerhs_fn_data* -- pointer to structure for ``prerhs_fn``.
      * *prerhs_fn* -- function pointer to perform preprocessing of the right-hand side evaluations.
```

doc/shared/sundomeigest/SUNDomEigEst_API.rst
```diff
@@ -247,6 +263,22 @@ instead of supplying a dummy routine.
       A :c:type:`SUNErrCode`.
 
 
+.. c:function:: SUNErrCode SUNDomEigEstimator_SetRhsAtLinearizationPoint(SUNDomEigEstimator DEE, N_Vector Fyt)
+
+   This *optional* function sets the value of the right-hand side function at the linearization point when using
+   :c:func:`SUNDomEigEstimator_SetRhs`.  When this is supplied, the *DEE* module 
+   will use the value instead of re-evaluating the RHS at the linearization point.
```
drreynolds:
```suggestion
   will use the provided vector instead of re-evaluating the RHS at the linearization point.
```

doc/shared/sundomeigest/SUNDomEigEst_API.rst
```diff
@@ -475,6 +507,13 @@ provide either a :c:type:`SUNATimesFn` or a :c:type:`SUNRhsFn`, as described bel
    to do its job. The time parameter :math:`t` and the vector :math:`y` should be left 
    unchanged.
 
+.. c:type:: int (*SUNPreRhsFn)(sunrealtype t, N_Vector y, void* prerhs_fn_data)
+
+   Used to perform any preprocessing of the right-hand side function before
+   performing a discrete Jacobian-vector product using quotient approximations of the Jacobian.
+   The parameter *prerhs_fn_data* is a pointer to any information about RHS which the function needs in order 
+   to do its job. The time parameter :math:`t` and the vector :math:`y` should be left 
+   unchanged.
```
drreynolds:
```suggestion
   Used to perform any preprocessing of the right-hand side function before
   performing a Jacobian-vector product using difference quotient approximations.
   The parameter *prerhs_fn_data* is a pointer to any problem-defining information that this function needs in order 
   to do its job. Both the time parameter :math:`t` and the vector :math:`y` inputs should be left 
   unchanged.
```

src/arkode/arkode_lsrkstep.c
```diff
@@ -2890,6 +2890,40 @@ int lsrkStep_ComputeNewDomEig(ARKodeMem ark_mem, ARKodeLSRKStepMem step_mem)
...
+
+    /* Synchronize the preprocessing functions */
+    retval = SUNDomEigEstimator_SetPreRhsFn(step_mem->DEE, ark_mem->user_data,
+                                            ark_mem->PreRhsFn);
+    if (retval != SUN_SUCCESS)
+    {
+      arkProcessError(ark_mem, ARK_DEE_FAIL, __LINE__, __func__, __FILE__,
+                      "SUNDomEigEstimator_SetPreRhsFn failed");
+      return ARK_DEE_FAIL;
+    }
```
drreynolds, line 2926: This should only be called if LSRKStep was provided a `PreRhsFn` from the user:
```suggestion
    /* Synchronize the preprocessing functions */
    if (ark_mem->PreRhsFn)
    {
      retval = SUNDomEigEstimator_SetPreRhsFn(step_mem->DEE, ark_mem->user_data,
                                              ark_mem->PreRhsFn);
      if (retval != SUN_SUCCESS)
      {
        arkProcessError(ark_mem, ARK_DEE_FAIL, __LINE__, __func__, __FILE__,
                        "SUNDomEigEstimator_SetPreRhsFn failed");
        return ARK_DEE_FAIL;
      }
    }
```

drreynolds (commented):
I have a few recommended corrections, but otherwise this looks great.

src/arkode/arkode_lsrkstep.c
maggul: @drreynolds, DEE has its own `prerhs_fn` pointer and we set this pointer by `SUNDomEigEstimator_SetPreRhsFn`. When the user turns of `PreRhsFn` for ARKODE, this should also reflect in the DEE memory. Therefore, I particularly removed this check to make sure if `PreRhsFn` is set to `NULL` at some point, this should effectively reflect in the DEE memory as well. This approach effectively sets `X_CONTENT(DEE)->prerhs_fn = NULL` if `ark_mem->PreRhsFn = NULL`.

drreynolds: I get that you want to support users turning on/off their `PreRhsFn`, but that goal seems to require you to call `SUNDomEigEstimator_SetPreRhsFn` at every internal time step.  I struggle to envision a user who actually wants to toggle their `PreRhsFn` on/off, so I worry that the current implementation will add a small (but frequently repeated) tax for the 99% of other users who do not toggle.

I could be mistaken, but doesn't `if (step_mem->DEE != NULL)` only evaluate to `True` with user-provided DEEs?  In that case, I think it would be cleaner and more efficient to just tell users to call `SUNDomEigEstimator_SetPreRhsFn` on their `DEE` module directly, instead of  having LSRKStep pass this on at each step.  In the extremely rare case where a user wants to provide a `PreRhsFn` and use a custom SUNDomEigEstimator, I don't think it is too much to ask them to call both `ARKodeSetPreRhsFn` and `SUNDomEigEstimator_SetPreRhsFn`.

maggul: I prefer to keep the current automatic synchronization instead of asking users to call both setters themselves. Users can do this, but it may be easy to forget, and normally the `DEE` preprocessing function should match `ARKODE`’s preprocessing function. This synchronization approach only sets two pointers in the `SUNDomEigEstimator_SetPreRhsFn` function, so I expect its cost to be very small. It also handles both enabling and disabling PreRhsFn consistently.

```
SUNErrCode SUNDomEigEstimator_SetPreRhsFn_Arnoldi(SUNDomEigEstimator DEE,
                                                  void* prerhs_fn_data,
                                                  SUNPreRhsFn prerhs_fn)
{
  SUNFunctionBegin(DEE->sunctx);

  SUNAssert(DEE, SUN_ERR_ARG_CORRUPT);
  SUNAssert(Arnoldi_CONTENT(DEE), SUN_ERR_ARG_CORRUPT);

  /* set function pointers to integrator-supplied Preprocess RHS routine
     and data, and return with success */
  Arnoldi_CONTENT(DEE)->prerhs_fn      = prerhs_fn;
  Arnoldi_CONTENT(DEE)->prerhs_fn_data = prerhs_fn_data;

  return SUN_SUCCESS;
}
```

Would you be OK with keeping this approach? Otherwise, I can remove this set call and add a documentation note explaining that users should call `SUNDomEigEstimator_SetPreRhsFn` themselves when they do it for ARKODE.

drreynolds: I am not the only reviewer on this PR (just the first).  I suggest that we leave this conversation as "unresolved" until others can provide their feedback.

gardner48: I think this can be handled in a different way with two updates:
1. Have LSRKStep provide its own implementation of `ARKodeSetPreRhsFn` that calls `SUNDomEigEstimator_SetPreRhsFn` if an estimator is attached. This way a user should only ever need to call `ARKodeSetPreRhsFn` in the time integration setting.
2. At initialization, call `SUNDomEigEstimator_SetPreRhsFn` if an estimator is attached. This covers the case of calling `ARKodeSetPreRhsFn` before attaching an estimator.

This will remove the additional overhead of calling `SUNDomEigEstimator_SetPreRhsFn` every time an estimate is needed, but still allows a user to change the function whenever they want. The only limitation is a user should *never* call `SUNDomEigEstimator_SetPreRhsFn` unless using it as a standalone module.

src/arkode/arkode_lsrkstep.c
```diff
@@ -2890,6 +2890,40 @@ int lsrkStep_ComputeNewDomEig(ARKodeMem ark_mem, ARKodeLSRKStepMem step_mem)
 
   if (step_mem->DEE != NULL)
   {
+    /* Pass the current time and solution to the dominant eigenvalue estimator */
```
gardner48, line 2895: The cost is likely small, but I want to note by moving to using the `ATimes` function supplied by the estimator, we are adding the overhead of calling `SetRhsLinearizationPoint` and `SetRhsAtLinearizationPoint` each estimate.

src/arkode/arkode_lsrkstep.c
```diff
@@ -2890,6 +2890,40 @@ int lsrkStep_ComputeNewDomEig(ARKodeMem ark_mem, ARKodeLSRKStepMem step_mem)
...
+                                                         ark_mem->yn);
+    if (retval != SUN_SUCCESS)
+    {
+      arkProcessError(ark_mem, ARK_DEE_FAIL, __LINE__, __func__, __FILE__,
+                      "SUNDomEigEstimator_SetRhsLinearizationPoint failed");
+      return ARK_DEE_FAIL;
+    }
+
+    /* Pass the current RHS vector fn to the dominant eigenvalue estimator */
+    if (ark_mem->fn_is_current == SUNTRUE)
```
gardner48: Should we evaluate the RHS function if it is not already current? That way we can save it and re-use within the step rather than the estimator evaluating it.

src/arkode/arkode_lsrkstep.c
```diff
@@ -2910,6 +2944,14 @@ int lsrkStep_ComputeNewDomEig(ARKodeMem ark_mem, ARKodeLSRKStepMem step_mem)
     }
     step_mem->num_dee_iters += num_iters;
 
+    retval = SUNDomEigEstimator_GetNumRhsEvals(step_mem->DEE, &(step_mem->nfeDQ));
```
gardner48, line 2948: Is this a cumulative count or just the number of evaluations from the most recent estimate? It should be the count from the last estimate to be consistent with `GetNumIters`. In which case, a this use use a local value and `nfeDQ` should be updated below (like above with `GetNumIters`).

src/arkode/arkode_lsrkstep.c
```diff
@@ -3230,89 +3272,6 @@ int lsrkStep_legendre_P_complex(int s, sunrealtype zR, sunrealtype zI,
...
-                          "F_n(:) =");
-      SUNLogInfo(ARK_LOGGER, "DomEig JvTimes",
-                 "status = failed rhs eval, retval = %i", retval);
-      return (ARK_RHSFUNC_FAIL);
-    }
-    ark_mem->fn_is_current = SUNTRUE;
-  }
-
-  /* Initialize perturbation to 1/||v|| */
-  sig = ONE / N_VWrmsNorm(v, ark_mem->ewt);
```
gardner48, line 3283: The estimators use a different value for the perturbation that does not take into account the error weights and that switch is probably why the cleanup in this RP changed the answer files when I would have expected them to be unaffected. 

I suggest we add a function to pass the weights to the estimator so it can compute the perturbation as before and only use the unweighted version if weights are not provided (e.g., standalone usage).

src/arkode/arkode_lsrkstep_io.c
```diff
@@ -553,12 +553,12 @@ int LSRKStepSetDomEigEstimator(void* arkode_mem, SUNDomEigEstimator DEE)
   /* Attach the DEE pointer to the step memory */
   step_mem->DEE = DEE;
 
-  /* Set the ATimes function for the DEE with A_data = arkode_mem */
-  retval = SUNDomEigEstimator_SetATimes(DEE, arkode_mem, lsrkStep_DQJtimes);
+  /* Set the RHS function for the DEE with user_data */
+  retval = SUNDomEigEstimator_SetRhs(DEE, ark_mem->user_data, step_mem->fe);
```
gardner48, line 557: This should be done at initialization since the RHS function can be changed when calling `LSRKStepReInit*`.

src/sundials/sundials_domeigestimator.c
```diff
@@ -64,22 +64,24 @@ SUNDomEigEstimator SUNDomEigEstimator_NewEmpty(SUNContext sunctx)
...
+  ops->setreltol                  = NULL;
+  ops->setinitialguess            = NULL;
+  ops->initialize                 = NULL;
+  ops->estimate                   = NULL;
+  ops->getres                     = NULL;
+  ops->getnumiters                = NULL;
+  ops->getnumrhsevals             = NULL;
+  ops->getnumatimescalls          = NULL;
+  ops->write                      = NULL;
+  ops->destroy                    = NULL;
```
gardner48, line 84: In the future avoid reordering lists like this so it's easier to check what's been added or removed.

src/sundomeigest/arnoldi/sundomeigest_arnoldi.c
```diff
@@ -239,74 +257,93 @@ SUNErrCode SUNDomEigEstimator_SetRhsLinearizationPoint_Arnoldi(
   N_VScale(ONE, y, Arnoldi_CONTENT(DEE)->rhs_linY);
   SUNCheckLastErr();
 
+  /* Update the cached RHS flag */
+  Arnoldi_CONTENT(DEE)->Fy_is_current = SUNFALSE;
+
```
gardner48, line 262: If `SetRhsAtLinearizationPoint` is called before `SetRhsLinearizationPoint`, this will incorrectly update the flag. 
```suggestion
```
In the next major release we should update `SetRhsLinearizationPoint` to also take the RHS vector as input (can optionally be `NULL`) and remove `SetRhsAtLinearizationPoint` to avoid the call order dependence.

src/sundomeigest/arnoldi/sundomeigest_arnoldi.c
```diff
@@ -239,74 +257,93 @@ SUNErrCode SUNDomEigEstimator_SetRhsLinearizationPoint_Arnoldi(
...
-  }
-  if (Arnoldi_CONTENT(DEE)->num_warmups < 0)
+  if (Arnoldi_CONTENT(DEE)->Fy == NULL)
   {
-    Arnoldi_CONTENT(DEE)->num_warmups = DEE_NUM_OF_WARMUPS_ARNOLDI_DEFAULT;
+    Arnoldi_CONTENT(DEE)->Fy = N_VClone(Fyt);
+    SUNCheckLastErr();
   }
 
+  N_VScale(ONE, Fyt, Arnoldi_CONTENT(DEE)->Fy);
```
gardner48, line 281: Is this clone and copy necessary? It looks like `Fy` is read-only, so in the integrator case we could just use the input vector since we know it won't be updated or destroyed. In the less common standalone use case, the user will need to be aware of this.

src/sundomeigest/arnoldi/sundomeigest_arnoldi.c
```diff
@@ -239,74 +257,93 @@ SUNErrCode SUNDomEigEstimator_SetRhsLinearizationPoint_Arnoldi(
...
+  SUNFunctionBegin(DEE->sunctx);
+
+  SUNAssert(DEE, SUN_ERR_ARG_CORRUPT);
+  SUNAssert(Arnoldi_CONTENT(DEE), SUN_ERR_ARG_CORRUPT);
   SUNAssert(Arnoldi_CONTENT(DEE)->ATimes, SUN_ERR_ARG_CORRUPT);
   SUNAssert(Arnoldi_CONTENT(DEE)->V, SUN_ERR_ARG_CORRUPT);
-  SUNAssert(Arnoldi_CONTENT(DEE)->q, SUN_ERR_ARG_CORRUPT);
+
+  int* kry_dim     = &(Arnoldi_CONTENT(DEE)->kry_dim);
+  int* num_warmups = &(Arnoldi_CONTENT(DEE)->num_warmups);
```
gardner48: These values should not be updated below, so they don't need to be pointers and you can remove all the dereferences below
```suggestion
  const int kry_dim     = Arnoldi_CONTENT(DEE)->kry_dim;
  const int num_warmups = Arnoldi_CONTENT(DEE)->num_warmups;
```

src/sundomeigest/arnoldi/sundomeigest_arnoldi.c
```diff
@@ -239,74 +257,93 @@ SUNErrCode SUNDomEigEstimator_SetRhsLinearizationPoint_Arnoldi(
...
+
+  if (*num_warmups < 0) { *num_warmups = DEE_NUM_OF_WARMUPS_ARNOLDI_DEFAULT; }
```
gardner48: This is unnecessary, the value is initialized to the default and the set function guards against invalid values
```suggestion
```

src/sundomeigest/arnoldi/sundomeigest_arnoldi.c
```diff
@@ -239,74 +257,93 @@ SUNErrCode SUNDomEigEstimator_SetRhsLinearizationPoint_Arnoldi(
...
+  xgeev_f77(&jobvl, &jobvr, &N, A, &lda, wr, wi, NULL, &ldvl, NULL, &ldvr,
+            &work, &lwork, &info);
 
   /* The workspace size is returned as the first entry of the work array */
   Arnoldi_CONTENT(DEE)->LAPACK_lwork = (sunindextype)work;
 
-  Arnoldi_CONTENT(DEE)->LAPACK_work = (sunrealtype*)malloc(
-    Arnoldi_CONTENT(DEE)->LAPACK_lwork * sizeof(sunrealtype));
+  Arnoldi_CONTENT(DEE)->LAPACK_work =
+    (sunrealtype*)malloc(((sunindextype)work) * sizeof(sunrealtype));
```
gardner48: This was slightly clearer before since `work` is unpacked into `LAPACK_lwork` above.
```suggestion
  Arnoldi_CONTENT(DEE)->LAPACK_work = (sunrealtype*)malloc(
    Arnoldi_CONTENT(DEE)->LAPACK_lwork * sizeof(sunrealtype));
```

src/sundomeigest/arnoldi/sundomeigest_arnoldi.c
```diff
@@ -407,41 +444,53 @@ SUNErrCode SUNDomEigEstimator_Estimate_Arnoldi(SUNDomEigEstimator DEE,
...
   SUNAssert(Arnoldi_CONTENT(DEE)->V, SUN_ERR_ARG_CORRUPT);
-  SUNAssert(Arnoldi_CONTENT(DEE)->q, SUN_ERR_ARG_CORRUPT);
   SUNAssert(Arnoldi_CONTENT(DEE)->Hes, SUN_ERR_ARG_CORRUPT);
 
   int retval;
-  sunindextype n = Arnoldi_CONTENT(DEE)->kry_dim;
-  sunrealtype normq;
-  Arnoldi_CONTENT(DEE)->num_ATimes = 0;
-  Arnoldi_CONTENT(DEE)->num_iters  = 0;
+  int* num_warmups = &(Arnoldi_CONTENT(DEE)->num_warmups);
```
gardner48: Like above, this should not be updated here, so it does not need to be a pointer and the dereferences below can be removed.
```suggestion
  const int num_warmups = Arnoldi_CONTENT(DEE)->num_warmups;
```

src/sundomeigest/power/sundomeigest_power.c
```diff
@@ -234,6 +260,33 @@ SUNErrCode SUNDomEigEstimator_SetRhsLinearizationPoint_Power(
   N_VScale(ONE, y, PI_CONTENT(DEE)->rhs_linY);
   SUNCheckLastErr();
 
+  /* Update the cached RHS flag */
+  PI_CONTENT(DEE)->Fy_is_current = SUNFALSE;
```
gardner48, line 264: Same update as in Arnoldi
```suggestion
```

src/sundomeigest/power/sundomeigest_power.c
```diff
@@ -234,6 +260,33 @@ SUNErrCode SUNDomEigEstimator_SetRhsLinearizationPoint_Power(
...
+  SUNAssert(PI_CONTENT(DEE), SUN_ERR_ARG_CORRUPT);
+  SUNAssert(Fyt, SUN_ERR_ARG_CORRUPT);
+
+  if (PI_CONTENT(DEE)->Fy == NULL)
+  {
+    PI_CONTENT(DEE)->Fy = N_VClone(Fyt);
+    SUNCheckLastErr();
+  }
+
+  N_VScale(ONE, Fyt, PI_CONTENT(DEE)->Fy);
```
gardner48, line 284: Like in Arnoldi, I think we could avoid the clone and copy

src/sundomeigest/power/sundomeigest_power.c
```diff
@@ -377,83 +437,92 @@ SUNErrCode SUNDomEigEstimator_Estimate_Power(SUNDomEigEstimator DEE,
...
-  if (PI_CONTENT(DEE)->is_complex && (PI_CONTENT(DEE)->q_prev == NULL))
-  {
-    /* allocate q_prev vector */
-    PI_CONTENT(DEE)->q_prev = N_VClone(PI_CONTENT(DEE)->q);
-    SUNCheckLastErr();
-  }
+
+  sunbooleantype is_complex = PI_CONTENT(DEE)->is_complex;
+
+  if (is_complex) { SUNAssert(PI_CONTENT(DEE)->v_prev, SUN_ERR_ARG_CORRUPT); }
```
gardner48, line 444: The is already checked else where. 
```suggestion
```

doc/shared/sundomeigest/SUNDomEigEst_API.rst
gardner48, line 1: Note the newly added functions in the change log and recent changes files

doc/shared/sundomeigest/SUNDomEigEst_API.rst
```diff
@@ -230,6 +230,24 @@ instead of supplying a dummy routine.
...
+
+   **Arguments:**
+
+      * *DEE* -- a SUNDomEigEstimator object.
+      * *prerhs_fn_data* -- pointer to structure for ``prerhs_fn``.
+      * *prerhs_fn* -- function pointer to perform preprocessing of the right-hand side evaluations.
+
+   **Return value:**
+
+      A :c:type:`SUNErrCode`.
```
gardner48, line 249:
```suggestion
      A :c:type:`SUNErrCode`.

   .. versionadded:: x.y.z
```

doc/shared/sundomeigest/SUNDomEigEst_API.rst
```diff
@@ -247,6 +265,22 @@ instead of supplying a dummy routine.
...
+   will use the provided vector instead of re-evaluating the RHS at the linearization point.
+
+   **Arguments:**
+
+      * *DEE* -- a SUNDomEigEstimator object.
+      * *Fyt* -- the value of the right-hand side function at the linearization point.
+
+   **Return value:**
+
+      A :c:type:`SUNErrCode`.
```
gardner48, line 283:
```suggestion
      A :c:type:`SUNErrCode`.

   .. versionadded:: x.y.z
```

doc/shared/sundomeigest/SUNDomEigEst_API.rst
```diff
@@ -475,6 +509,13 @@ provide either a :c:type:`SUNATimesFn` or a :c:type:`SUNRhsFn`, as described bel
    to do its job. The time parameter :math:`t` and the vector :math:`y` should be left 
    unchanged.
 
+.. c:type:: int (*SUNPreRhsFn)(sunrealtype t, N_Vector y, void* prerhs_fn_data)
+
+   Used to perform any preprocessing of the right-hand side function before
+   performing a Jacobian-vector product using difference quotient approximations.
+   The parameter *prerhs_fn_data* is a pointer to any problem-defining information
+   that this function needs in order to do its job. Both the time parameter
+   :math:`t` and the vector :math:`y` inputs should be left unchanged.
```
gardner48, line 522:
```suggestion
   :math:`t` and the vector :math:`y` inputs should be left unchanged.
   
   .. versionadded:: x.y.z
```

gardner48 (changes requested):

src/arkode/arkode_lsrkstep.c
gardner48: Looks like this is cumulative across estimate calls, so the estimate functions need to be updated to zero this counter at the start of each estimation.

src/arkode/arkode_lsrkstep.c
maggul: You are right, the behavior changed with this PR. I can pass the weights to the estimator for sure, before doing that, as much as I remember, this WRMS-norm choice was not a strong preference. What if `sig` is always computed as in DEE, would that cause any problem? 

Also, we have an external DEE example where user can create a DEE, do their own estimation and return it with `dom_eig_fn` instead. This seems to be an approach that comes with more control over the estimator, and already adopted by gkeyll. In such cases, reverting this implementation, external DEE approach and DEE attached to the lsrk will produce slightly different estimates, and so user will observe stat differences, as you observed with the answer files. This is a possible downside that came to my mind immediately after seeing this comment.

What do you think? Should we keep it like this, or revert to `WRMS` norm by passing `ewt`? @drreynolds @gardner48

gardner48: We should include the weighting as this can be important for some problems. The version currently in the estimator is similar to what KINSOL does, but does not have the weighting that KINSOL uses.

src/sundomeigest/arnoldi/sundomeigest_arnoldi.c
maggul: This was intensional done to ensure that if the linerization point is updated, possibly changed, it will enforce a new RHS call at the new linerization point. What if the linearization point changes, but the estimator thinks that RHS is up to date, it is more serious issue than calling extra RHS. Maybe we can add a note on the doc to ensure users know about what the order of calls should be?

gardner48: That would be fine. Use the `.. danger::` directive when adding this to the docs so the call order dependence is clear.

src/arkode/arkode_lsrkstep.c
maggul: Moved the first rhs evaluation up in the step so that it will be current for DEE regardless.

src/sundomeigest/arnoldi/sundomeigest_arnoldi.c
gardner48: Some discussion started [here](https://github.com/llnl/sundials-private/issues/155) about not introducing `SetRhsAtLinearizationPoint` in 7.9.0 only to remove it in 8.0.0. If we go this route, I think we should then keep the integrator provided `ATimes` functions to avoid the additional RHS call this function prevents. Then in 8.0.0 we can move to using the estimator provided `ATimes` implementation. What do others think?

maggul: I can split this PR into two parts: one for `7.9` and another for `8.0`. This way, we would not need to deprecate anything immediately after introducing it, and we could also avoid extra RHS evaluations caused by untimely updates in who will compute `ATimes`.

gardner48: Sounds good to me

src/sundomeigest/power/sundomeigest_power.c
maggul: @gardner48, users can change the `is_complex` flag directly without calling our set function, even though we do not expect that usage. For example, in `ark_brusselator_lsrk_externaldomeigest`, adding the following line immediately before the estimate call changes the flag:

`((SUNDomEigEstimatorContent_Power)(DEE->content))->is_complex = SUNFALSE;
`

Therefore, I think we should keep this check so that we return the appropriate error code.

We can also add cloning like above here for a `false`-to-`true` change, so it executes as intended.

gardner48: This is fine for now. In the next major release, we should make the content structure internal, so we don't need unnecessary guards against this kind of behavior.

src/arkode/arkode_lsrkstep.c
maggul: It is not cummulative, every estimate call updates it to zero

gardner48: Where is `nfevals` reset to zero? I don't see it in the power or Arnoldi implementations.

maggul: I am confused with `GetNumIters`. You are right! It seems like a bug.

Issue #955, opened 2026-06-10
[BUG] arkLsMassSolve ignores some return values from user callbacks.


bangerth:
One of the folks working on the deal.II SUNDIALS interfaces, @vovannikov, has spent what I think must have been a quality afternoon tracing why one of our tests fails. The details are in https://github.com/dealii/dealii/pull/19838 but come down to code that I (mis-)wrote: If a user callback fails in user-provided linear solvers, I either returned +1 or -1, depending on whether it's a recoverable or non-recoverable failure (and zero if everything is fine). I *think* that's what the documentation of SUNDIALS at some point said I should do, but it's possible that that was about the nonlinear solver callbacks in KINSOL and not about the SUNDIALS linear solvers. Regardless, for the linear solvers, I need to return `SUNLS_PSOLVE_FAIL_REC` or `SUNLS_PSOLVE_FAIL_UNREC`, with values +805 and -808, respectively.

In any case, my mistake fell through the cracks because the place where these return codes are checked, namely https://github.com/llnl/sundials/blob/0a1024b7fb902a5eae3142adf8c13561c6989543/src/arkode/arkode_ls.c#L3940-L3972 , has a `switch` statements with a lot of cases, but no `default:` clause that would have caught my invalid values. The code then simply falls through to the `return 0` that indicates success.

@vovannikov kindly fixed the bugs on our side, but it would be useful if on the SUNDIALS side, the `switch` statement would gain a `default:` case that checks that only specific values can be returned by user callbacks, and errors out otherwise.

(In deal.II, pretty much every `switch` statement, regardless of where it is, has a `default: DEAL_II_ASSERT_UNREACHABLE();` case that aborts the whole program if we end up with an unexpected value.)

drreynolds:
This should be fixed by PR #956.  Once that is approved/merged, we'll close this issue.

balos1:
Fixed in v7.8.0

PR #956, opened 2026-06-10
Bugfix: Catch unknown linear solver return value

drreynolds:
Fixed a bug where an unrecognized error return flag from a user-provided
SUNLinearSolver module would register as a successful linear solve.

This addresses issue #955


bangerth:
Nice, thank you for fixing it in all of the other places as well. I wonder what other mistakes of mine that's going to uncover :-)

gardner48 (approved):

PR #961, opened 2026-06-16
Bugfix: cmake FindPETSC paths and static builds

gardner48:
* Remove MPI, Kokkos, and KokkosKernels imported targets from FindPETSC
* Use PETSc provided link flags for all dependencies
* Fix linking with static PETSc builds


balos1 (approved):

balos1:
@BarrySmith this should fix the issue you were having in #935

PR #962, opened 2026-06-16
Bugfix/sundials4py sunstepper

drreynolds:
Fixed bug in the sundials4py wrappers for user-provided SUNStepper evolve and one-step functions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,9 @@ embedded implicit-explicit MRI-GARK coupling tables.
 
 ### Bug Fixes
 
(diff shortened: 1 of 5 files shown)


balos1 (dismissed):

balos1:
@drreynolds  Looks like the one step function and evolve functions defined in https://github.com/llnl/sundials/blob/0a1024b7fb902a5eae3142adf8c13561c6989543/bindings/sundials4py/test/fixtures.py#L46 need to be updated for the new signature.

balos1 (approved):

PR #974, opened 2026-07-01
sundials4py: add cuda support

balos1:


balos1:
This PR reminds me that I think we should add `N_VSetDeviceArrayPointer` to the N_Vector API.

bindings/sundials4py/test/test_nvector.py
drreynolds, line 1: It's great that this tests so many routines in the CUDA N_Vector, but it looks like it's missing tests for vectors with managed memory, and for vectors that use the SUNDIALS memory helper.  I'm not sure how easy it is to use these features from Python, but since they're included in `bindings/sundials4py/nvector/nvector_cuda.cpp` then I recommend that they be tested here as well.

examples/python/arkode/ark_heat1D.py
drreynolds, line 1: It would be nice if the non-CUDA features of this file and `ark_heat1D_cuda.py` matched, so that it would be easier for readers to learn the CUDA-specific differences.  I think that the CUDA version seems nicer, so I recommend that the non-CUDA items from that file be propagated here.

balos1:
@copilot apply the diff from the format_check

balos1:
Needs #980.

examples/python/arkode/ark_heat1D.py
```diff
@@ -54,174 +54,146 @@
...
+        JV = sun.N_VGetArrayPointer(Jvvec)
+
         JV[:] = 0.0
         c1 = k / dx / dx
         c2 = -2.0 * k / dx / dx
-        # Vectorized tridiagonal product
         JV[1:-1] = c1 * V[:-2] + c2 * V[1:-1] + c1 * V[2:]
         JV[0] = 0.0
         JV[-1] = 0.0
         return 0
```
drreynolds, line 112: Can this be moved to sit just before `def main():`, so that the structure better matches the cupy, jax, and torch versions?

examples/python/arkode/ark_heat1D_jax.py
```diff
@@ -0,0 +1,247 @@
...
+        c1 = self.k / self.dx / self.dx
+        c2 = -2.0 * self.k / self.dx / self.dx
+        result = self.jnp.zeros_like(v)
+        result = result.at[1:-1].set(c1 * v[:-2] + c2 * v[1:-1] + c1 * v[2:])
+        result = result.block_until_ready()
+        if self.array_device == "cuda":
+            sun.N_VSetDeviceArrayPointer_Cuda(result, Jvvec)
+        else:
+            sun.N_VGetNumpyArray(Jvvec)[:] = np.asarray(result)
+        return 0
```
drreynolds, line 193: In the other versions of this problem (native Python, cupy, and torch), stationary boundary conditions are applied in `f` and `jtv`, but it looks like they are missing from this version.

drreynolds (commented):
Overall this looks great.  I had two comments; once these are resolved (and the PR passes CI testing), then I think this is ready to merge.

bindings/sundials4py/nvector/generate.yaml
```diff
@@ -104,3 +104,22 @@ modules:
...
+    # pointer-address handling in Python.
+    - "^N_VNewEmpty_Cuda$"
+    - "^N_VNew_Cuda$"
+    - "^N_VNewManaged_Cuda$"
+    - "^N_VNewWithMemHelp_Cuda$"
+    - "^N_VMake_Cuda$"
+    - "^N_VMakeManaged_Cuda$"
+    - "^N_VSetHostArrayPointer_Cuda$"
+    - "^N_VSetDeviceArrayPointer_Cuda$"
+    - "^N_VGetHostArrayPointer_Cuda$"
```
gardner48, line 125: Should this include `GetDeviceArrayPointer_Cuda`?
```suggestion
    - "^N_VGetHostArrayPointer_Cuda$"
    - "^N_VGetDeviceArrayPointer_Cuda$"
```

bindings/sundials4py/nvector/nvector_cuda.cpp
```diff
@@ -0,0 +1,543 @@
...
+      device_array = std::move(it->second.device_array);
+      owners.erase(it);
+    }
+  }
+
+  destroy(v);
+
+  if (device_array.ptr())
+  {
+    nb::gil_scoped_acquire gil;
```
gardner48: What does this line do?

bindings/sundials4py/nvector/nvector_cuda.cpp
```diff
@@ -0,0 +1,543 @@
...
+  if (it == owners.end()) { return; }
+
+  if (v && v->ops &&
+      v->ops->nvdestroy == destroy_cuda_nvector_with_python_device_array)
+  {
+    v->ops->nvdestroy = it->second.destroy;
+  }
+  owners.erase(it);
+}
+
```
gardner48, line 155: Why do the GPU wrappers need the above memory management utilities but the serial vector does not?

bindings/sundials4py/nvector/nvector_cuda.cpp
```diff
@@ -0,0 +1,543 @@
...
+  catch (const nb::cast_error&)
+  {
+    throw sundials4py::error_returned(std::string(name) +
+                                      " must be a string or None");
+  }
+}
+
+ArrayDevice parse_device(nb::object device, N_Vector v)
+{
+  auto value = optional_string(device, "device");
```
gardner48: Could this have a more descriptive name e.g., `device_type` or `memory_type`? Is the `name` input needed? Looks like it's only for the error message and I'm not sure it makes the message more clear.

bindings/sundials4py/nvector/nvector_cuda.cpp
```diff
@@ -0,0 +1,543 @@
...
+void require_cuda_nvector(N_Vector v)
+{
+  if (!is_cuda_nvector(v))
+  {
+    throw sundials4py::error_returned(
+      "CUDA array access requires a CUDA N_Vector");
+  }
+}
+
+void apply_copy_from(N_Vector v, ArrayDevice device, CopyFrom copy_from)
```
gardner48: Does this need both `device` and `copy_from` as inputs i.e., could we have `None -> return`, `Cpu -> CopyToDevice`, and `Device -> CopyFromDevice`?

bindings/sundials4py/nvector/nvector_cuda.cpp
```diff
@@ -0,0 +1,543 @@
...
+
+nb::object get_cupy_array(N_Vector v, nb::object device, nb::object copy_from)
+{
+  auto target = parse_device(device, v);
+  if (target == ArrayDevice::Cpu)
+  {
+    throw sundials4py::error_returned(
+      "N_VGetCupyArray does not support device='cpu'");
+  }
+  apply_copy_from(v, target, parse_copy_from(copy_from));
```
gardner48: I assume because we combine accessing and copying between memory spaces in this function is why we need both `ArrayDevice` and `CopyFrom` instead of a single device or memory space type. Is that right? Do we document somewhere that some of the accessors have this additional option that's not on the C side?

bindings/sundials4py/nvector/nvector_cuda.cpp
```diff
@@ -0,0 +1,543 @@
...
+      N_VSetHostArrayPointer_Cuda(h_vdata_1d.data(), v);
+    },
+    nb::arg("h_vdata_1d"), nb::arg("v"));
+
+  m.def(
+    "N_VSetDeviceArrayPointer_Cuda",
+    [](std::uintptr_t d_vdata, N_Vector v)
+    {
+      release_python_device_array(v);
+      N_VSetDeviceArrayPointer_Cuda(reinterpret_cast<sunrealtype*>(d_vdata), v);
```
gardner48: Should this call `retain_python_device_array` or is that not needed for raw pointers?

bindings/sundials4py/sundials/sundials_nvector.cpp
```diff
@@ -22,13 +22,131 @@
...
+
 #include <sundials/sundials_nvector.hpp>
+#include "sundials/sundials_config.h"
 
 namespace nb = nanobind;
 using namespace sundials::experimental;
 
 namespace sundials4py {
 
+namespace {
```
gardner48, line 41: Can this be removed? I think everything in this anonymous namespace is in `nvector_cuda.cpp`, so could be shared

bindings/sundials4py/sundials/sundials_nvector.cpp
```diff
@@ -48,20 +166,36 @@ void bind_nvector(nb::module_& m)
     },
     nb::rv_policy::reference);
 
+#ifndef SUNDIALS_NVECTOR_CUDA
```
gardner48, line 265: Same here, I think everything in this `#ifndef SUNDIALS_NVECTOR_CUDA` block is in `nvector_cuda.cpp`

bindings/sundials4py/sundials/sundials_nvector.cpp
gardner48: Looking at the tests now, I see these are needed to use the different backends with the serial vector. Should we move this shared code to a common file so it's not duplicated here and in `nvector_cuda.cpp`?

bindings/sundials4py/sundials/sundials_nvector.cpp
gardner48: Noted this above after looking at the tests, these are needed to use the different backends with the serial vector. Should we move this shared code to a common file so it's not duplicated here and in `nvector_cuda.cpp`?

bindings/sundials4py/test/test_nvector.py
```diff
@@ -78,6 +85,287 @@ def test_make_nvector(vector_type, sunctx):
...
+
+
+@pytest.mark.skipif(
+    "N_VNewWithMemHelp_Cuda" not in globals(), reason="CUDA bindings are not enabled"
+)
+@pytest.mark.parametrize("use_managed_mem", [False, True])
+def test_create_nvector_cuda_with_memhelp(use_managed_mem, sunctx):
+    mem_helper = SUNMemoryHelper_Cuda(sunctx)
+    if mem_helper is None:
+        pytest.skip("CUDA memory helper creation failed")
```
gardner48: Should this fail the test instead of skipping?

bindings/sundials4py/test/test_sunmemoryhelper.py
```diff
@@ -25,3 +25,13 @@
...
     assert mem_helper is not None
+
+
+@pytest.mark.skipif(
+    "SUNMemoryHelper_Cuda" not in globals(), reason="CUDA bindings are not enabled"
+)
+def test_create_memory_helper_cuda(sunctx):
+    mem_helper = SUNMemoryHelper_Cuda(sunctx)  # noqa: F405
+    if mem_helper is None:
+        pytest.skip("CUDA memory helper creation failed")
```
gardner48: Should this fail rather than skip?

gardner48 (changes requested):
Are any updates to `pyproject.toml` needed? Should also update the install instructions to include enabling cuda, jax, and pytorch support.

bindings/sundials4py/nvector/nvector_cuda.cpp
balos1: Yes. The goal is to provide the same semantics as torch, cupy, jax etc.

bindings/sundials4py/nvector/nvector_cuda.cpp
balos1: Its not the GPU that forces this, its JAX. JAX arrays are immutable so users cannot do
```python
y = N_VGetArrayPointer(yvec)
y[:] = calculation(...)
```
Instead, they have to do 
```python
y = N_VGetJaxArray(yvec)
result = self.jnp.zeros_like(y)
result.at[:].set(calculation(...))
N_VSetJaxArray(result)
```
SetJaxArray is a helper for SetArrayPointer/SetDeviceArrayPointer. Since we are now setting a pointer, there has to be some owner of it and the pointer previously owned by the N_Vector has to be freed. Hence the extra memory management.

Alternatively, they can use the experimental JAX Array Refs and let SUNDIALS work on it directly. However, this is experimental and not the typical JAX workflow.

bindings/sundials4py/nvector/nvector_cuda.cpp
balos1: "device" is what its called in torch/jax etc.

bindings/sundials4py/nvector/nvector_cuda.cpp
balos1: Grabs the interpreter lock

bindings/sundials4py/nvector/nvector_cuda.cpp
gardner48: Ok, that makes sense. Maybe in the future we should add function like `TakeArrayPointer` and `GiveArrayPointer` to as functions that transfer ownership as counter parts to `Get` and `Set`.

bindings/sundials4py/nvector/nvector_cuda.cpp
gardner48: Sorry, I wasn't clear about what I was referring. By "this" I meant the function name `optional_string`.

bindings/sundials4py/sundials/sundials_nvector.cpp
balos1: I have now deduplicated as much I could without requiring backend checks at runtime.

bindings/sundials4py/nvector/nvector_cuda.cpp
```diff
@@ -0,0 +1,518 @@
+#include "sundials4py.hpp"
```
gardner48, line 23: Add copyright block

bindings/sundials4py/nvector/nvector_cuda.cpp
```diff
@@ -0,0 +1,518 @@
...
+
+  if (device == ArrayDevice::Cuda &&
+      (status != cudaSuccess || (attrs.type != cudaMemoryTypeDevice &&
+                                 attrs.type != cudaMemoryTypeManaged)))
+  {
+    cudaGetLastError();
+    throw sundials4py::error_returned(
+      "Device array pointer is not CUDA-accessible");
+  }
+
```
gardner48: Does there need to be a case to handle `device == ArrayDevice::Cpu`?

bindings/sundials4py/nvector/nvector_cuda.cpp
```diff
@@ -0,0 +1,518 @@
...
+  if (status != cudaSuccess) { cudaGetLastError(); }
```
gardner48: Should this go right after `cudaPointerGetAttributes` so we don't need to check it again in each block above?

bindings/sundials4py/nvector/nvector_cuda.cpp
```diff
@@ -0,0 +1,518 @@
...
+void copy_to_cuda_array_pointer(N_Vector v, const sunrealtype* ptr,
+                                ArrayDevice source_device)
+{
+  require_cuda_nvector(v);
+
+  auto length = N_VGetLength(v);
+  if (length == 0) { return; }
+  if (!ptr)
+  {
+    throw sundials4py::error_returned("JAX array pointer must not be null");
```
gardner48: Is this specifically for the JAX backend?
```suggestion
    throw sundials4py::error_returned("Array pointer must not be null");
```

bindings/sundials4py/nvector/nvector_cuda.cpp
```diff
@@ -0,0 +1,518 @@
...
+  if (source_device == ArrayDevice::Cuda)
+  {
+    cudaPointerAttributes attrs;
+    auto status = cudaPointerGetAttributes(&attrs, ptr);
+    if (status != cudaSuccess || (attrs.type != cudaMemoryTypeDevice &&
+                                  attrs.type != cudaMemoryTypeManaged))
+    {
+      cudaGetLastError();
+      throw sundials4py::error_returned(
+        "JAX array pointer is not CUDA-accessible");
```
gardner48: Is this specifically for the JAX backend?
```suggestion
        "Array pointer is not CUDA-accessible");
```

bindings/sundials4py/nvector/nvector_cuda.cpp
```diff
@@ -0,0 +1,518 @@
...
+  require_cuda_nvector(v);
+
+  auto length = N_VGetLength(v);
+  if (length == 0) { return; }
+  if (!ptr)
+  {
+    throw sundials4py::error_returned("JAX array pointer must not be null");
+  }
+
+  SUNMemoryType source_type = SUNMEMTYPE_HOST;
```
gardner48: Should this match `source_device`?

bindings/sundials4py/nvector/nvector_cuda.cpp
```diff
@@ -0,0 +1,518 @@
...
+  {
+    throw sundials4py::error_returned(
+      "CUDA N_Vector does not have device storage");
+  }
+
+  auto source = SUNMemoryHelper_Wrap(content->mem_helper,
+                                     const_cast<sunrealtype*>(ptr), source_type);
+  if (!source)
+  {
+    throw sundials4py::error_returned("Failed to wrap the JAX array pointer");
```
gardner48: Is this specifically for the JAX backend?
```suggestion
    throw sundials4py::error_returned("Failed to wrap the array pointer");
```

bindings/sundials4py/nvector/nvector_cuda.cpp
```diff
@@ -0,0 +1,518 @@
...
+  }
+
+  auto dealloc_status = SUNMemoryHelper_Dealloc(content->mem_helper, source,
+                                                stream);
+  if (cuda_status != cudaSuccess || copy_status != SUN_SUCCESS ||
+      dealloc_status != SUN_SUCCESS)
+  {
+    cudaGetLastError();
+    throw sundials4py::error_returned(
+      "Failed to copy JAX array data into the CUDA N_Vector");
```
gardner48: Is this specifically for the JAX backend?
```suggestion
      "Failed to copy array data into the CUDA N_Vector");
```

bindings/sundials4py/nvector/nvector_cuda.cpp
```diff
@@ -0,0 +1,518 @@
...
+  cuda_status   = cudaGetDevice(&device_id);
+  if (cuda_status != cudaSuccess)
+  {
+    cudaGetLastError();
+    return 0;
+  }
+  return device_id;
+}
+
+void sync_to_host(N_Vector v)
```
gardner48: Maybe `copy_to_host` to avoid confusion with `sync_device` below means a call to `cudaStreamSynchronize`.

Should there be a `copy_to_device` also?

bindings/sundials4py/sundials/sundials_nvector.cpp
```diff
@@ -22,57 +22,288 @@
...
+
+  std::string platform;
+  if (copy)
+  {
+    auto devices = nb::cast<nb::list>(
+      nb::module_::import_("builtins").attr("list")(data.attr("devices")()));
+    if (devices.size() != 1)
+    {
+      throw sundials4py::error_returned(
+        "N_VSetJaxArray requires a single-device jax.Array");
```
gardner48, line 95: Should we add multi-device support as something to support in the future?

bindings/sundials4py/sundials/sundials_nvector.cpp
```diff
@@ -22,57 +22,288 @@
...
+  {
+    throw sundials4py::error_returned(
+      "N_VSetJaxArray only supports JAX CPU and CUDA devices");
+  }
+
+  auto shape = nb::cast<nb::tuple>(data.attr("shape"));
+  if (shape.size() != 1)
+  {
+    throw sundials4py::error_returned(
+      "N_VSetJaxArray requires a one-dimensional array");
```
gardner48, line 128: Should multi-dimensional array support be another future feature to consider?

bindings/sundials4py/sundials/sundials_nvector.cpp
```diff
@@ -22,57 +22,288 @@
...
-        [](sundials4py::Array1d arr, N_Vector v)
-        {
-          if (arr.shape(0) != N_VGetLength(v))
-          {
-            throw sundials4py::error_returned(
-              "Array shape does not match vector length");
-          }
-          N_VSetArrayPointer(arr.data(), v);
-        });
+  m.def("N_VSetJaxArray", &set_jax_array, nb::arg("array"), nb::arg("v"),
```
gardner48, line 284: Should there be an equivalent function for the other backends?

doc/shared/Python/Usage.rst
gardner48, line 1: Add CUDA support in the python interface to the change log and update to Python guide to note which backends are supported and how to enable them.

doc/shared/Python/Usage.rst
```diff
@@ -251,6 +251,14 @@ This allows you to use numpy operations for vector and matrix data, and to pass
 numpy arrays to and from SUNDIALS routines efficiently and without unnecessary
 copies.
 
+For CPU ``N_Vector`` objects, ``N_VGetNumpyArray`` and
+``N_VGetTorchTensor`` create lightweight zero-copy views. ``N_VGetJaxArray``
+also exposes the vector data without copying, but importing an external buffer
+into the JAX runtime has appreciably higher fixed overhead. In performance
+sensitive CPU callbacks or Python loops, obtain a JAX view once and reuse it
+rather than creating a new view for every operation. JAX is generally most
+effective when its JIT compilation or GPU execution amortizes this setup cost.
```
gardner48, line 304: Add documentation on the automatic memory movement from the device to host and how users move data from the host to the device.

doc/shared/Python/Usage.rst
```diff
@@ -222,7 +222,7 @@ Arrays
 ``N_Vector`` objects in sundials4py are compatible with numpy's
 ``ndarray``. Each ``N_Vector`` can work on a numpy arrays without copies, and
 you can access and modify the underlying data directly using
-``N_VGetArrayPointer``, which returns a numpy ``ndarray`` view of the data.
+``N_VGetNumpyArray``, which returns a numpy ``ndarray`` view of the data.
```
gardner48, line 257: Note the move to use `N_VGetNumpyArray` in the change log

doc/shared/sundials_vars.py
```diff
@@ -191,3 +191,12 @@
     ("c:func", "SUNSparseMatrix_ToCSC"),
     ("c:func", "SUNSparseMatrix_ToCSR"),
 ]
+
+# Python bindings that do not wrap a corresponding SUNDIALS C API function.
+python_only_functions = {
+    "N_VGetCupyArray",
+    "N_VGetJaxArray",
+    "N_VGetNumpyArray",
+    "N_VGetTorchTensor",
```
gardner48, line 200: Can users `Set` these also?

gardner48 (changes requested):

bindings/sundials4py/sundials/sundials_nvector.cpp
balos1: No. We only need this for JAX due to the immutable arrays.

doc/shared/sundials_vars.py
balos1: They should not so we do not provide Set functions for them. Users should work with the returned arrays directly except when its JAX.

bindings/sundials4py/nvector/nvector_array_helpers.hpp
```diff
@@ -0,0 +1,266 @@
...
+inline std::shared_ptr<std::remove_pointer_t<N_Vector>> wrap_nvector(N_Vector v)
+{
+  return sundials::experimental::our_make_shared<
+    std::remove_pointer_t<N_Vector>, sundials::experimental::N_VectorDeleter>(v);
+}
+
+enum class ArrayDevice
+{
+  Cpu,
+  Cuda
```
Steven-Roberts: The dev docs say to use upper case for enum values
```suggestion
  CPU,
  CUDA
```

bindings/sundials4py/sundials/sundials_nvector.cpp
balos1: A user can just reshape it to be 1d.

bindings/sundials4py/sundials/sundials_nvector.cpp
balos1: Yes. I have a follow up pr already in progress.

gardner48 (dismissed):

gardner48 (dismissed):

gardner48 (approved):

PR #978, opened 2026-07-23
Bugfix: don't check convergence rate when m=1 in auto solver switching 

balos1:


drreynolds (commented):
This looks good to me, so once it is passing the CI I'll approve.

drreynolds (commented):
Is there a reason this PR is still marked as "draft"?  

Aside from the CI failures, I think this looks good to go.

src/sunnonlinsol/auto/sunnonlinsol_auto.c
```diff
@@ -286,13 +286,22 @@ int SUNNonlinSolConvTest_Auto(SUNNonlinearSolver sub_nls, N_Vector y,
...
+    SUNErrCode iter_retval;
     SUNErrCode crate_retval;
 
     /* If the integrator-provided convergence test passed, exit with success and
        don't consider switching, since fixed-point is still converging fine. */
     if (retval == SUN_SUCCESS) { return SUN_SUCCESS; }
 
+    /* The convergence rate estimate is not meaningful on the first nonlinear
+       iteration. Integrators initialize crate to 1 and only update it after an
+       iteration pair is available. */
```
gardner48: Minor comment clarification. The convergence rate is initialized to one before every solve if the nonlinear solver does not have a linear solver setup function. Otherwise it can be the prior convergence rate estimate unless a linear solver setup if performed. For our current use case, so this is just in case we attach something like Picard instead of fixed-point. 
```suggestion
    /* The convergence rate estimate is not meaningful on the first nonlinear
       iteration as integrators only update it after an iteration pair is available. */
```

gardner48 (changes requested):
Very minor comment update. Otherwise looks like all that's need is to update the output files and open a PR to the answers repo.

gardner48 (approved):

PR #986, opened 2026-08-07
ARKODE Order Checking Improvements

Steven-Roberts:
This replaces https://github.com/llnl/sundials/pull/640 and removes the breaking changes so this is suitable for a minor release.


src/arkode/arkode_butcher.c
```diff
@@ -430,686 +305,456 @@ sunbooleantype ARKodeButcherTable_IsStifflyAccurate(ARKodeButcherTable B)
   int i;
   for (i = 0; i < B->stages; i++)
   {
-    if (SUNRabs(B->b[i] - B->A[B->stages - 1][i]) > 100 * SUN_UNIT_ROUNDOFF)
+    if (SUNRabs(B->b[i] - B->A[B->stages - 1][i]) > SUN_UNIT_ROUNDOFF)
```
drreynolds: I'm a bit concerned that if a user supplies a Butcher table for us to use, and if this is filled in using "complex" calculations, it's possible that what they intended as a stiffly accurate table could fail this test if our tolerance is exactly `SUN_UNIT_ROUNDOFF`.

src/arkode/arkode_butcher.c
```diff
@@ -430,686 +305,456 @@ sunbooleantype ARKodeButcherTable_IsStifflyAccurate(ARKodeButcherTable B)
...
-  /*    order 2 condition */
-  if ((*q) == 1)
+
+  /* Copy the tree from the current buffer to the list */
+  memcpy(&gen->list[gen->length], gen->current, tree_length * sizeof(*gen->list));
+  gen->length = new_length;
+  return ARK_SUCCESS;
+}
+
+/* A utility function to write a tree in text for with t representing a leaf and
```
drreynolds:
```suggestion
/* A utility function to write a tree in text, with t representing a leaf and
```

src/arkode/arkode_butcher.c
```diff
@@ -1118,3164 +763,56 @@ int ARKodeButcherTable_CheckOrder(ARKodeButcherTable B, int* q, int* p,
...
-  {
-    if ((q > Q + R + 1) || (q > 2 * Q + 2)) { break; }
-    q++;
-  }
-  free(tmp);
-  return (q);
+  ARKodeButcherTable tables[] = {B1, B2};
+  int retval                  = check_tables(tables, SUNTRUE, q, p, outfile);
+  // TODO(SBR): remove this for SUNDIALS 8 to make this consistent with non-ARK
+  // version of this function
```
drreynolds: Can you clarify what "this" refers to, the entire `ARKodeButcherTable_CheckARKOrder` function, or the manipulation of `retval` below?

drreynolds (commented):
Overall this looks great.  It dramatically simplifies what I'd assembled previously and extends checking to higher order methods.  I only have two minor comments/requests, plus one grammar suggestion.

src/arkode/arkode_butcher.c
Steven-Roberts: I can revert this. Ideally, no calculations would be done since that introduces floating point errors, but we can't really enforce that.

drreynolds: Exactly.  I realize that we worked to enforce this for our own tables, but we have less control over user behavior.  I also question whether tightening this tolerance actually accomplishes anything, since the change only catches situations where  `eps < |b_i-A_{j,i}| < 100*eps`.  I find it unlikely that this would occur for any user unless they thought the method was stiffly accurate but they mistakenly computed the coefficients using slightly different formulas.

That said, if you feel strongly about the tighter tolerance, then we could explain the stiff accuracy test in the documentation for `ARKStepSetTables` and `ERKStepSetTable`, in that it allows us to leverage the structure to improve performance.  If we provided the formula and tolerance for this test, then it would be the user's responsibility to ensure their stiffly accurate input tables actually satisfy this tighter tolerance.

Steven-Roberts: I just reverted to the old check.

drreynolds (dismissed):

test/answers
gardner48, line 1: Open a PR in the answers repo to update the output files

src/arkode/arkode_butcher.c
```diff
@@ -438,678 +313,448 @@ sunbooleantype ARKodeButcherTable_IsStifflyAccurate(ARKodeButcherTable B)
...
+  return retval;
+}
 
-  /*    check failure modes first */
-  if (((*q) < B->q) && ((*q) < 6)) { return (-1); }
-  if (d)
+static int check_tables(ARKodeButcherTable* tables, sunbooleantype ark, int* q,
+                        int* p, FILE* outfile)
+{
+  if (!is_valid_table(tables[0]) || (ark && !is_valid_table(tables[1])))
```
gardner48, line 723: This assume two tables are input but below `ARKodeButcherTable_CheckOrder`  makes the call `return check_tables(&B, SUNFALSE, q, p, outfile);`

gardner48 (changes requested):

src/arkode/arkode_butcher.c
Steven-Roberts: This only assumes two tables are given when `ark` is true. If the concern is accessing `tables[1]`, that should never happen due to short circuiting.

gardner48: Ah, I missed the `ark &&` before `!is_valid_table(tables[1])`. Looks good.

gardner48 (approved):

PR #991, opened 2026-08-16
Agent skill: add code-review skill for github copilot that can check for compliance with sundials style guide

balos1:


.agents/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,86 @@
+---
+name: code-review
```
balos1, line 2: This code-review skill doesnt follow the new `sundials-` prefix convention because it is a special name recognized by GitHub Copilot.

Steven-Roberts (commented):
This looks reasonable to me. The "Source Review Checklist" and "Documentation Review Checklist" summarize conventions already in the developer docs, so it we update the docs, we'll need to remember to update this. I wonder if a review would differ much if we removed those sections. Copilot is already instructed to review our docs.

gardner48 (dismissed):

gardner48 (approved):

PR #1011, opened 2026-09-04
reorganize cvode examples

balos1:
I think it is a good time to reoganize/retitle our examples directory to make it easier for users to find examples. As a first step, I have reorganized the cvode examples with this proposed structure:

```
examples/cvode/
  ├── README.md
  ├── CMakeLists.txt
  ├── c/
  │   ├── README.md
  │   ├── CMakeLists.txt
  │   └── <example>/
  │       ├── CMakeLists.txt
  │       ├── <example>.c
  │       ├── <example>.out
  │       ├── <input-data>
  │       └── <supporting-script>.py
  ├── cpp/
  │   ├── README.md
  │   ├── CMakeLists.txt
  │   └── <example>/
  │       ├── CMakeLists.txt
  │       ├── <example>.cpp
  │       ├── <example>.hpp
  │       ├── <example>.out
  │       ├── <input-data>
  │       └── <supporting-script>.py
  ├── cuda/
  ├── hip/
  ├── sycl/
  ├── fortran/
```

 Examples are first grouped by implementation language. Cuda, hip are treated as languages since this is what CMake does. I thought it made since then to bring sycl up to the level with them. Each executable then has its own directory containing its source, reference .out files, inputs, scripts, and leaf CMakeLists.txt.

Execution models and optional dependencies - such as MPI, OpenMP, PETSc, hypre, RAJA, and Kokkos - are handled by CMake feature guards instead of additional directory levels. This avoids ambiguous categorization when an example uses multiple technologies. Installed examples mirror the same hierarchy.


balos1:
@copilot  fix the format check

drreynolds (commented):
Overall, I like the revised structure, but I'm a bit confused about the subfolders in each `package/example`.  At first, it looked to me like the pattern is `package/example/language-tpl`, but then I saw instances where the final sorting was not based on third party library, but instead dimensionalty (e.g., `cvode/cvBrusselator/cpp-kokkos` versus `cvode/cvBrusselator/cpp-kokkos-2d`) or vector interface (e.g., `cvode/cvDiurnal/c-mpi` versus `cvode/cvDiurnal/c-mpi-mpimanyvector`).  Can you better explain the logic behind the final sorting structure, i.e., if it is `package/example/language-X` then what is `X`?

If `X` were TPL alone, then it makes perfect sense at the CMake level.  On the other hand, I'd think of 1D versus 2D as living at a higher level of sorting, e.g., `cvode/cvBrusselator1D/cpp-kokkos` versus `cvode/cvBrusselator2D/cpp-kokkos`, since they're effectively solving different problems with similar names.

balos1:
@drreynolds I suppose the logic was `example/variant` where `variant` was either a TPL or otherwise notable distinction such as the case where there are versions that use different vectors. In the 1D/2D case, I agree, they can be thought of as distinct examples/problems.

gardner48:
> In the 1D/2D case, I agree, they can be thought of as distinct examples/problems.

In the `cvode/cvBrusselator/cpp-kokkos-2d` case this is the same example problem, but it is using a 2D Kokkos view rather than a 1D view, so I think this example variant should stay where it is but maybe with a revised name e.g., `cvode/cvBrusselator/cpp-kokkos-2d-view`

examples/cvode/README.md
```diff
@@ -0,0 +1,87 @@
...
+variants named for their implementation language and optional TPLs. For
+example, `cvAdvDiff/c-mpi-hypre` is the MPI/HYPRE C variant of the
+advection-diffusion example. Each variant contains its source files, reference
+output, input data, supporting scripts, and `CMakeLists.txt`.
+
+```text
+examples/cvode/
+├── README.md
+├── CMakeLists.txt
+├── cvAdvDiff/
```
Steven-Roberts: Unless this can be updated automatically, I would remove it or provide a truncated version.

Steven-Roberts (commented):
A couple questions to consider:

- Should we remove `cv_`, `ark_`, etc prefixes on example file names since the directory structure already classifies by package?
- Should we rename examples that don't use `snake_case.c` file names, e.g., `cvAdvDiff_bnd.c`?

drreynolds:
> A couple questions to consider:
> 
> * Should we remove `cv_`, `ark_`, etc prefixes on example file names since the directory structure already classifies by package?
> * Should we rename examples that don't use `snake_case.c` file names, e.g., `cvAdvDiff_bnd.c`?

On the former: I think this is "no", at least with our current CI infrastructure, that collects all `.out` files for a given build configuration into a single folder.

On the latter: this sounds fine to me.

